Issue #183, opened 2014-12-31
custom post type not passed through to search results


tott:
When searching for a custom post type the post type does not appear in search results.

I used the following query

```
$query = new WP_Query( array(
    'ep_match_all'   => true,
        'post_type' => 'profile',
        'post_status' => 'publish',
        'sites' => 'all',
        'posts_per_page' => $limit,
        'offset' => $offset,
) );
```

As in filter_the_posts() the get_post() function get utilized to fill in missing parameters the default parameters of WP_Post get inherited through the new WP_Post call.

I addressed this in
https://github.com/tott/ElasticPress/commit/fd2965095a0e1762890c799d7514e158cd474f3f

PR #184, opened 2014-12-31
When post_type is not being set in post object, it gets overwritten by g...

tott:
This addresses https://github.com/10up/ElasticPress/issues/183

### DIFF
--- a/classes/class-ep-wp-query-integration.php
+++ b/classes/class-ep-wp-query-integration.php
@@ -85,7 +85,7 @@ class EP_WP_Query_Integration {
 			restore_current_blog();
 
 			switch_to_blog( $post->site_id );
-			
+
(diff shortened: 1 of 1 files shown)


tlovett1:
Can you reopen this against `develop`?

tott:
Will do, also thought of a better solution and will change this accordingly.

AaronHolbrook:
Hi @tott - do you mind reopening this against develop so we can get this into the next version?

Thanks!
Aaron

tott:
Absolutely no problem. As i noticed that my PR just addressed one particular issue i'm still working on a more general solution for mapping those attributes. Will reopen when this is in place

tlovett1:
Somehow this already made it into develop anyway.

PR #189, opened 2015-01-14
Develop

### LINKED ISSUE #136: Output header indicating if ES is being used for results
For search results pages, it would be helpful to have a header in the response that indicates if elastic search was used for the results on the page or not. In the case of multiple queries on the same page using ES, I think the main query would be the important one to know. Something like ```x-elastic-search true``` would do for me - I often find myself wondering if the results for the search are WordPress or Elastic Search, and I think this would be an easy way to determine this.

ivankruchkoff:
bug in wp cli when calling network params on single-site installs.
resolved #136

### DIFF
--- a/bin/wp-cli.php
+++ b/bin/wp-cli.php
@@ -34,7 +34,7 @@ class ElasticPress_CLI_Command extends WP_CLI_Command {
 	public function put_mapping( $args, $assoc_args ) {
 		$this->_connect_check();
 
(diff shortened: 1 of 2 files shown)


tlovett1:
The network wide fix looks good. I have a comment on #136. Also, PR needs to be refactored with proper tabbing.

tlovett1:
Looks good. Going forward let's try to separate things like this into multiple PRs.

PR #192, opened 2015-01-20
Issue-190

### LINKED ISSUE #190: Provide filters for post statuses in sync manager class
Right now the function ```action_sync_on_transition``` of the ```EP_Sync_Manager``` class contains a couple of hard coded post status (publish). One of the problem I've been facing with creating the plugin that would provide search in the admin is that we need to index all potential post status. 

@AaronHolbrook I wanted your input before creating a pull request. I think some filters to let the elasticpress admin plugin override some post status would be interesting but completely overriding the function could also be another solution to this problem.

jonathanbardo:
This should fix #190.

Here's what I've done:
- Provided a filter for post statuses which default to publish
- Change the sync mecanism to make sure it takes all the post statuses into account
- Change Elasticsearch mapping of post_status to "not_analyzed" so we are able to filter the search query (require re-indexing for people to start using this in queries)

By implementing those changes, we are able to implement those filters to make the admin search work:

```php
public static function filter_ep_indexable_post_status( post_statuses ) {
  $post_statuses[] = 'draft';
  $post_statuses[] = 'pending';
  //...
  return $post_statuses;
}
add_filter( 'ep_indexable_post_status', 'filter_ep_indexable_post_status' , 10, 1 );

public static function filter_ep_formatted_args( $formatted_args ) {
  if ( ! is_admin() ) {
      $formatted_args['filter']['and'][] = array(
      'term' => array( 'post_status' => array( 'publish' ) ),
    );
  }

  return $formatted_args;
}
add_filter( 'ep_formatted_args',  'filter_ep_formatted_args' , 10, 1 );
```

@tlovett1 or @AaronHolbrook any feedback to make this better is greatly appreciated!

### DIFF
--- a/classes/class-ep-api.php
+++ b/classes/class-ep-api.php
@@ -447,7 +447,7 @@ class EP_API {
 						),
 						'post_status' => array(
 							'type' => 'string',
-							'index' => 'no'
+							'index' => 'not_analyzed'
 						),
 						'post_name' => array(
 							'type' => 'multi_field',
@@ -583,7 +583,7 @@ class EP_API {
 			'post_title'        => get_the_title( $post_id ),
 			'post_excerpt'      => $post->post_excerpt,
 			'post_content'      => apply_filters( 'the_content', $post->post_content ),
-			'post_status'       => 'publish',
+			'post_status'       => $post->post_status,
 			'post_name'         => $post->post_name,
 			'post_modified'     => $post_modified,
 			'post_modified_gmt' => $post_modified_gmt,
--- a/classes/class-ep-sync-manager.php
+++ b/classes/class-ep-sync-manager.php
@@ -52,7 +52,9 @@ class EP_Sync_Manager {
 			return;
 		}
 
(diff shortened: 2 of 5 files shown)


classes/class-ep-sync-manager.php
```diff
@@ -61,7 +63,7 @@ public function action_sync_on_transition( $new_status, $old_status, $post ) {
 		}
 
 		// Our post was published, but is no longer, so let's remove it from the Elasticsearch index
```
AaronHolbrook: Should update the comment here to be more accurate with what is happening (no longer just checking 'published' status)

classes/class-ep-api.php
```diff
@@ -447,7 +447,7 @@ public function put_mapping() {
 						),
 						'post_status' => array(
 							'type' => 'string',
-							'index' => 'no'
+							'index' => 'not_analyzed'
```
AaronHolbrook: good catch @jonathanbardo

AaronHolbrook:
This looks great @jonathanbardo - glad to have your contributions!

props @jonathanbardo

PR #195, opened 2015-01-28
Abide by the 'exclude_from_search' flag when running search queries.

ryanboswell:
WordPress core search enforces limiting the post types searched to just those without the 'exclude_from_search' flag turned on when it would otherwise be 'any', but previously ElasticPress didn't enforce this limit at all, and just opened up to any post type if one wasn't specified already.

This PR does the following:
1. If we're doing a search for 'any' post, default to post types not excluded from search.
2. For all other query types (not search), continue to allow querying across all indexed post types.

### DIFF
--- a/classes/class-ep-wp-query-integration.php
+++ b/classes/class-ep-wp-query-integration.php
@@ -191,7 +191,30 @@ class EP_WP_Query_Integration {
 
 		$query_vars = $query->query_vars;
 		if ( 'any' == $query_vars['post_type'] ) {
-			unset( $query_vars['post_type'] );
+			
+			if( $query->is_search() ) {
+				
+				/*
+				 * This is a search query
+				 * To follow WordPress conventions,
+				 * make sure we only search 'searchable' post types
+				 */
+				$searchable_post_types = get_post_types( array( 'exclude_from_search' => false ) );
+				
+				// Conform the post types array to an acceptable format for ES
+				$post_types = array();
+				foreach( $searchable_post_types as $type ) {
+					$post_types[] = $type;
+				}
+				$query_vars['post_type'] = $post_types;
+			} else {
+				
+				/*
+				 * This is not a search query
+				 * so unset the post_type query var
+				 */
+				unset( $query_vars['post_type'] );
+			}
 		}
 
 		$scope = 'current';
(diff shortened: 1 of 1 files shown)


AaronHolbrook:
Hey @ryanboswell thanks for the submission here, this looks like a good start, but I think there's a couple things we'll want to add.

Unfortunately with the way Github is set up, I'm unable to simply commit some changes onto your branch, and the changes aren't huge, so I'll just post up what I have and let you commit them (if that's alright?)

Here's my thinking for that entire conditional block:

```
if ( 'any' === $query_vars['post_type'] ) {

			if ( $query->is_search() ) {

				/*
				 * This is a search query
				 * To follow WordPress conventions,
				 * make sure we only search 'searchable' post types
				 */
				$searchable_post_types = get_post_types( array( 'exclude_from_search' => false ) );

				// If we have no searchable post types, there's no point going any further
				if ( empty( $searchable_post_types ) ) {
					return;
				}

				// Conform the post types array to an acceptable format for ES
				$post_types = array();
				foreach( $searchable_post_types as $type ) {
					$post_types[] = $type;
				}

				// These are now the only post types we will search
				$query_vars['post_type'] = $post_types;
			} else {

				/*
				 * This is not a search query
				 * so unset the post_type query var
				 */
				unset( $query_vars['post_type'] );
			}
		}
```

Most important was the realization that hey, if we no longer find any searchable post types, we can actually just return right then and there - that's pretty cool, no need wasting cycles requesting info from Elasticsearch. Second was (and this is our fault, not yours) - the first part should be a strict equality check that uses three `===` not just two `==` - might as well fix it while we're in here though.

If you could add those changes and push that up to this PR I'd be happy to merge them in.

Thanks again for your contribution - this was a nice find!

Cheers,
Aaron

ryanboswell:
@AaronHolbrook, happy to put those additions in.

I actually did think about the case where there were no results since I remember seeing in the core query that it forces an empty MySQL result, and I meant to add a check but I forgot, so thanks for catching that.

AaronHolbrook:
Looks great @ryanboswell thanks for the changes.

props to @ryanboswell for a good suggestion and contribution here @tlovett1

tlovett1:
Thanks @ryanboswell !

GhostToast:
While the search is working based on a post_type's `exclude_from_search` flag, the `get_indexable_post_types()` defaults to `public` flag. I realize this can be circumvented with the filter, but seems the default behaviors of mapping and searching would utilize the same check.

PR #206, opened 2015-02-12
Support date in orderby - closes #203

### LINKED ISSUE #203: Enable orderby date
Are there plans to implement this? Seems odd that it's not available seeing it's the default wp_query orderby argument.

tlovett1:
PR adds support for `date` in orderby and adds a unit test.

### DIFF
--- a/tests/test-single-site.php
+++ b/tests/test-single-site.php
@@ -716,9 +716,9 @@ class EPTestSingleSite extends EP_Test_Base {
 
 		$this->assertEquals( 3, $query->post_count );
 		$this->assertEquals( 3, $query->found_posts );
-		$this->assertEquals( 'ordertes 333', $query->posts[0]->post_title );
+		$this->assertEquals( 'ordertest 222', $query->posts[0]->post_title );
 		$this->assertEquals( 'ordertest 111', $query->posts[1]->post_title );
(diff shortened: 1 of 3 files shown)


tests/test-single-site.php
```diff
@@ -691,6 +691,33 @@ public function testSearchPostTitleOrderbyQuery() {
...
 
 	/**
+	 * Test post_date orderby query
+	 *
+	 * @since 1.4
+	 */
+	public function testSearchPostDateOrderbyQuery() {
+		ep_create_and_sync_post( array( 'post_title' => 'ordertes 333' ) );
+		ep_create_and_sync_post( array( 'post_title' => 'ordertest 111' ) );
+		ep_create_and_sync_post( array( 'post_title' => 'ordertest 222' ) );
```
AaronHolbrook: I'm wondering that if it's not failing because in your test you're indexing them so immediately one after another that there is no discernible timestamp difference. @tlovett1

tlovett1: That's a good idea. It passed locally for me. I'll add a sleep call.

AaronHolbrook:
Ok - reran the travis build, there was one test that timed out.

Finally passed! merging

tlovett1:
Yay!

PR #215, opened 2015-02-24
Remove duplicates filters and allow sync post termination

tlovett1:
Remove duplicate `ep_post_sync_args` filter and add `ep_prepare_post_args` filter. If ep_post_sync_args is empty, don't sync

### DIFF
--- a/classes/class-ep-api.php
+++ b/classes/class-ep-api.php
@@ -608,6 +608,9 @@ class EP_API {
 			//'site_id'         => get_current_blog_id(),
 		);
 
+		/**
+		 * This filter is named poorly but has to stay to keep backwards compat
+		 */
 		$post_args = apply_filters( 'ep_post_sync_args', $post_args, $post_id );
 
 		return $post_args;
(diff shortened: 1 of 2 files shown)


tlovett1:
@AaronHolbrook #reviewmerge

classes/class-ep-api.php
```diff
@@ -608,7 +608,7 @@ public function prepare_post( $post_id ) {
 			//'site_id'         => get_current_blog_id(),
 		);
 
-		$post_args = apply_filters( 'ep_post_sync_args', $post_args, $post_id );
+		$post_args = apply_filters( 'ep_prepare_post_args', $post_args, $post_id );
```
AaronHolbrook: @tlovett1 changing this here could result in buggy behavior while performing a bulk index (this function is called from two places, sync_post and queue_post)

AaronHolbrook: called here: https://github.com/10up/ElasticPress/blob/bugfix/duplicate-filter/classes/class-ep-sync-manager.php#L108

and here: https://github.com/10up/ElasticPress/blob/bugfix/duplicate-filter/bin/wp-cli.php#L342

tlovett1:
@AaronHolbrook I am merging this. The build error was due to a Travis problem.

PR #217, opened 2015-02-26
Allow for filtered query integration on admin ajax requests

tlovett1:
By default no integration will happen on admin ajax. This PR add a filter for this.

### DIFF
--- a/classes/class-ep-wp-query-integration.php
+++ b/classes/class-ep-wp-query-integration.php
@@ -32,7 +32,7 @@ class EP_WP_Query_Integration {
 		$admin_integration = apply_filters( 'ep_admin_wp_query_integration', false );
 
 		if ( defined( 'DOING_AJAX' ) && DOING_AJAX ) {
-			if ( apply_filters( 'ep_ajax_wp_query_integration', false ) ) {
+			if ( ! apply_filters( 'ep_ajax_wp_query_integration', false ) ) {
 				return;
 			} else {
 				$admin_integration = true;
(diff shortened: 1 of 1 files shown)


classes/class-ep-wp-query-integration.php
```diff
@@ -23,8 +23,23 @@ public function __construct() { }
...
+
+		if ( defined( 'DOING_AJAX' ) && DOING_AJAX ) {
+			if ( apply_filters( 'ep_ajax_wp_query_integration', false ) ) {
+				return;
+			} else {
+				$admin_integration = true;
+			}
+		}
+
+		if ( is_admin() && ! $admin_integration ) {
```
AaronHolbrook: I think this approach actually creates a hole.

Example use case: someone wants to enable integration on admin ajax requests, but NOT on the general admin integration.

I believe this encapsulates the logic more fully:
```
if (
  ( is_admin() && ! apply_filters( 'ep_admin_wp_query_integration', false ) ) ||
  ( defined( 'DOING_AJAX' ) && DOING_AJAX && ! apply_filters( 'ep_admin_ajax_wp_query_integration', false ) ) )
{
  return;
}
```

tlovett1:
@AaronHolbrook I don't think that will work since when DOING_AJAX is defined is_admin() will always return true.

AaronHolbrook:
It can be EITHER

is_admin && admin_integration
OR
DOING_AJAX && ajax_integration

So even if `is_admin` is true during DOING_AJAX, it doesn't matter

AaronHolbrook:
Looks good now

PR #220, opened 2015-03-05
Refactor wp-cli stats for multisite

Jaace:
Add functionality to stats method for displaying info for all sites on a multisite network.


tlovett1:
Nice work!

PR #221, opened 2015-03-11
Feature/date query

joeyblake:
Lets review what I have so far. I noticed some strangeness with multiple range filters when I was writing the tests so thats something that needs to be looked at for sure.

### DIFF
--- a/classes/class-ep-api.php
+++ b/classes/class-ep-api.php
@@ -336,49 +336,6 @@ class EP_API {
 					"date_detection" => false,
 					"dynamic_templates" => array(
 						array(
-							"template_date_terms" => array(
-								"path_match" => "date_terms.*",
-								"mapping" => array(
-									"type" => "object",
-									"path" => "full",
-									'fields' => array(
-										"year" => array(
-											"type" => "string",
-											'index' => 'not_analyzed'
-										),
-										"monthnum" => array(
-											"type" => "string",
-											'index' => 'not_analyzed'
-										),
-										"m" => array(
-											"type" => "string",
-											'index' => 'not_analyzed'
-										),
-										"w" => array(
-											"type" => "string",
-											'index' => 'not_analyzed'
-										),
-										"day" => array(
-											"type" => "string",
-											'index' => 'not_analyzed'
-										),
-										"hour" => array(
-											"type" => "string",
-											'index' => 'not_analyzed'
-										),
-										"minute" => array(
-											"type" => "string",
-											'index' => 'not_analyzed'
-										),
-										"second" => array(
-											"type" => "string",
-											'index' => 'not_analyzed'
-										)
-									)
-								)
-							)
-						),
-						array(
 							"template_meta" => array(
 								"path_match" => "post_meta.*",
 								"mapping" => array(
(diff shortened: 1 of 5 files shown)


classes/class-ep-api.php
```diff
@@ -511,6 +511,51 @@ public function put_mapping() {
...
+								"year" => array(
+									"type" => "integer",
+								),
+								"month" => array(
+									"type" => "integer",
+								),
+								"monthnum" => array(
+									"type" => "integer",
+								),
+								"m" => array(
```
tlovett1: What is the difference between `m` and `monthnum`?

classes/class-ep-api.php
```diff
@@ -612,10 +658,29 @@ public function prepare_post( $post_id ) {
 		 * This filter is named poorly but has to stay to keep backwards compat
 		 */
 		$post_args = apply_filters( 'ep_post_sync_args', $post_args, $post_id );
-
 		return $post_args;
 	}
 
+	private function prepare_date_terms( $post_date_gmt ) {
```
tlovett1: Docblock please :)

classes/class-ep-api.php
```diff
@@ -612,10 +658,29 @@ public function prepare_post( $post_id ) {
...
+		$timestamp = strtotime( $post_date_gmt );
+		$date_terms = array(
+			'year' => (int) date( "Y", $timestamp),
+			'month' => (int) date( "m", $timestamp),
+			'monthnum' => (int) date( "m", $timestamp),
+			'w' => (int) date( "W", $timestamp),
```
tlovett1: Why store this twice?

classes/class-ep-api.php
```diff
@@ -835,6 +900,33 @@ public function format_args( $args ) {
...
+		if ( $date_filter = EP_WP_Date_Query::simple_es_date_filter( $args ) ) {
+			$filter['and'][] = $date_filter;
+			$use_filters = true;
+		}
+
+		/**
+		 * 'date_query' arg support.
+		 * 
+		 */
+		if ( ! empty( $args['date_query'] ) ) {
```
tlovett1: Kind of a weird scenario but if someone had the `date_query` param as well as the simple date stuff, what would happen?

classes/class-ep-wp-date-query.php
```diff
@@ -0,0 +1,234 @@
...
+
+class EP_WP_Date_Query extends WP_Date_Query {
+
+	public function get_es_filter() {
+		$filter = $this->get_es_filter_for_clauses();
+
+		return $filter;
+	}
+
+	protected function get_es_filter_for_clauses() {
```
tlovett1: I'm confused on this. Why are we returning a piece of a SQL query?

classes/class-ep-wp-date-query.php
```diff
@@ -0,0 +1,234 @@
...
+					$filter_count = count( $clause_filter );
+					if ( ! $filter_count ) {
+						$filter_chunks['filters'][] = '';
+					} else {
+						$filter_chunks['filters'][] = $clause_filter;
+					}
+
+					// This is a subquery, so we recurse.
+				} else {
+					$clause_filter              = $this->get_es_filter_for_query( $clause, $depth + 1 );
```
tlovett1: We definitely need a unit test for this one.

classes/class-ep-wp-date-query.php
```diff
@@ -0,0 +1,234 @@
...
+					$filter_chunks['filters'][] = $clause_filter;
+				}
+			}
+		}
+
+		if ( empty( $relation ) ) {
```
tlovett1: I assume it's a todo to implement other relations?

classes/class-ep-wp-date-query.php
```diff
@@ -0,0 +1,234 @@
...
+
+		return $filter_array;
+	}
+
+	protected function get_es_filter_for_clause( $query, $parent_query ) {
+
+		// The sub-parts of a $where part.
+		$filter_parts = array();
+
+		$column = ( ! empty( $query['column'] ) ) ? esc_sql( $query['column'] ) : $this->column;
```
tlovett1: Don't understand this? Why `esc_sql`?

classes/class-ep-api.php
joeyblake: m is yearmonth YYYYMM, monthnum is the month number, month and monthnum are the same though

classes/class-ep-api.php
joeyblake: We don't need to, but we'll need a translation so query params get mapped to the correct place. I'll add that.

joeyblake:
@tlovett1 updated. The only question left was what happens if you mix a simple query with a date query. The answer is they are just combined in the filter, so the simple term gets lumped in with date query filter.

tlovett1:
I'm going to merge into develop to do testing a cleanup.

PR #223, opened 2015-03-16
Move mappings array to separate file.

mikaelmattsson:


mikaelmattsson:
This is the wrong pull request. Had hoped that it would dissapear when I deleted the repository.

mikaelmattsson:
Here is the correct pull request: https://github.com/10up/ElasticPress/pull/224

PR #230, opened 2015-03-30
adds single site post "fields => ids", and "fields => id=>parent" response

### LINKED ISSUE #227: If using fields => ids in WP_Query, ElasticPress nullifies the request
Just stumbled across this one.

**The Problem**
----------------------

Given a `WP_Query` arg array of the following:

```
$args = array(
  'post_type' => 'post',
  'fields' => 'ids',
  's' => 'findme',
);
```

ElasticPress runs and may / may not find posts. However, in WP's own `query.php` on L:3433: https://github.com/WordPress/WordPress/blob/bce851dcf25439a74c3f7ce42bb784c84ca8f3cc/wp-includes/query.php#L3433

`get_posts` actually just takes the request SQL and determines the number of found posts and returns immediately.

This deviates from what ElasticPress is expecting, which intentionally causes the SQL query to be nullified (`1 = 0`) because it wants to refill it later in `the_posts` filter down on http://github.com/WordPress/WordPress/blob/bce851dcf25439a74c3f7ce42bb784c84ca8f3cc/wp-includes/query.php#L3631 
where it refills the query object with all the posts it found.

**Ways forward**
-----------------------

I think during our filter of `posts_request`, we check to see if the user is requesting one of these early return fields `ids` or `id=>parent` and return a valid SQL query that will supply the IDs and site IDs of the posts returned. The problem with this that I see is that even though we could supply the IDs for cross site, the only field return is IDs, which doesn't provide a way for us to indicate which blog the post ID came from.

joeyblake:
and a multisite results workaround. Multisite results need to be handled this way because of the nature of WP_Query content sanitization. There are no filters available to allow altering IDs or id=>parent query responses. So I stick those in an ep_posts property on the query object.

Otherwise, single site works exactly as non-ep searches.

Fix #227

### DIFF
--- a/classes/class-ep-wp-query-integration.php
+++ b/classes/class-ep-wp-query-integration.php
@@ -309,7 +309,7 @@ class EP_WP_Query_Integration {
 			if ( is_multisite() ) {
 				$sql_query = ( 'id=>parent' == $query->query_vars['fields'] ) ? "SELECT ID, post_parent FROM $wpdb->posts WHERE 1=0" : "SELECT ID FROM $wpdb->posts WHERE 1=0";
 				$this->elasticsearch_fields = $query->query_vars['fields'];
-				unset( $query->query_vars['fields'] );
+				$query->query_vars['fields'] = '';
 			} else {
 				return $query->request;
 			}
--- a/classes/class-ep-wp-query-integration.php
+++ b/classes/class-ep-wp-query-integration.php
@@ -168,10 +168,47 @@ class EP_WP_Query_Integration {
 		if ( ! ep_elasticpress_enabled( $query ) || apply_filters( 'ep_skip_query_integration', false, $query ) || ! isset( $this->posts_by_query[spl_object_hash( $query )] ) ) {
 			return $posts;
 		}
-
-		$new_posts = $this->posts_by_query[spl_object_hash( $query )];
-
+		if ( ! empty( $this->elasticsearch_fields ) ) {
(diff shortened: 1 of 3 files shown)


tlovett1:
Can you add a unit test please?

tlovett1:
I'm testing this query:

```php
$args = array(
	's' => 'post',
	'sites' => 'all',
	'posts_per_page' => 100,
	'fields' => 'ids',
);

$query = new WP_Query( $args );
```

It is returning nothing when there should be 8 posts. I am also seeing these warnings:
```
Notice: Undefined index: fields in /srv/www/wordpress-trunk/wp-includes/query.php on line 3427
Notice: Undefined index: fields in /srv/www/wordpress-trunk/wp-includes/query.php on line 3436
```

Also, I don't understand the multisite logic here. Why are we only setting `elasticsearch_fields` in multisite?

joeyblake:
@tlovett1 The warning is a bug, I found it on friday in the middle of writing the tests. I' can have the patch ready in an hour with my tests.

As for the elasticsearch_fields, they are only used in multisite because a single site instance handles the ids and id=>parent in the default wp_query way. Multisite elasticpress needs to work differently so that the results can actually contain multi-site responses.

tlovett1:
I just tested `fields => 'ids'` on a single site install and it did not work. Have you tried this?

tlovett1:
Details in #227

PR #231, opened 2015-03-31
adds index_exists function, and hooks for multisite blog archival/deletion to delete blog index

joeyblake:


joeyblake:
Yes!

tlovett1:
Nice work.

Issue #82, opened 2014-08-29
Hook into a taxonomy term update and post meta update


AaronHolbrook:
Currently if a taxonomy term is updated, ElasticPress does not know to re-index all posts that have that taxonomy term associated with it - this could be a major problem.

tlovett1:
Same thing with post meta.

ChrisWiegman:
@dkotter After reviewing the two attached issues and their related PRs I'm going ahead and closing this issue as it appears to be fixed. Some testing seems as if it is working as well.

PR #234, opened 2015-04-06
added reindex function once term updated

EduardMaghakyan:
Hook into edit_term action to re-index posts once term is edited


tlovett1:
Which issue is this for?

EduardMaghakyan:
this was for #82

Issue #232, opened 2015-03-31
Emails about errors, even when there were not any errors


cmmarslender:
Just reindexed a multisite, and got a bunch of emails (one per site) about indexing errors, even when a site didn't have any actual errors. 

See screenshots:
https://cloudup.com/cuXQ4sX9udk

tlovett1:
@cmmarslender it must have been trying to index posts that don't exist:
https://github.com/10up/ElasticPress/blob/develop/bin/wp-cli.php#L433

How is the database setup? Is there multiple DB servers? Perhaps, they are out of sync.

cmmarslender:
Does have multiple DB servers, but I dont think they were out of sync - Maybe a possibility.

I was reindexing a multisite, and after double checking the emails, the very first email I got (which I ignored, assuming it was empty like the rest....) actually had posts. Maybe there is an issue with doing this network wide, where ```$this->failed_posts``` is not cleared for each site or something like this? Haven't look at the code is too much detail here, so maybe clearing ```$failed_posts``` is happening somewhere, just not seeing it at first glance.

tlovett1:
@EduardMaghakyan can you explain how this fixes the bug?

EduardMaghakyan:
@tlovett1 Well as far as I've understood send_bulk_errors is called for each site and if $this->failed_posts is not empty the email is sent. So if $this->failed_posts is not empty one site email will be sent for that site and for all sites which are indexed afterwards, even if there aren't any failed posts.

tlovett1:
Awesome thanks

PR #236, opened 2015-04-07
cleared failed_posts array after errors email is sent

EduardMaghakyan:
cleared failed_posts array after email is sent, to not send it only for one site for #232


tlovett1:
Thanks!

Issue #235, opened 2015-04-07
filter_posts_request() - object instead of ID passed into get_posts


qubika:
Installed elasticpress and work seamlessly, but noticed that author details weren't being returned in the search result.

After a bit of digging, found that the filter_posts_request() function in class-ep-wp-query-integration.php is passing the $post object into get_post rather than $post->ID so the object isn't being enriched.

Changing this seemed to fix things.

PR #239, opened 2015-04-09
use post id in get_post to enrich the post object and make it possible for multisite

EduardMaghakyan:
fix for #235 
Just passing post ID as argument to get_post will override some data in the initial post object and will not work on multisite, but it's better to use as it will return all possible fields. Then merging it with the initial object populates all fields and keeps the initial values.
I've added additional function to make this method work for multisite as well.

### DIFF
--- a/classes/class-ep-wp-query-integration.php
+++ b/classes/class-ep-wp-query-integration.php
@@ -297,7 +297,7 @@ class EP_WP_Query_Integration {
 
 		// if multisite is enabled get post object for each site
 		if( is_multisite() ) {
-			$new_posts = $this->fetch_wp_post_object( $new_posts );
+			$new_posts = $this->fill_post_objects( $new_posts );
 		}
 
 
@@ -317,9 +317,11 @@ class EP_WP_Query_Integration {
 	 * for WP_Post object
 	 *
 	 * @param  array $post_list list of found post objects
+	 *
+	 * @since  1.4
 	 * @return array            list of WP_Post objects
 	 */
-	public function fetch_wp_post_object( $post_list ) {
+	public function fill_post_objects( $post_list ) {
 		$post_objs = array();
 		$posts_ordered = array();
 		$grouped = array();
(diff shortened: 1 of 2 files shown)


tlovett1:
@EduardMaghakyan anyway you could push a test that demonstrates this bug?

classes/class-ep-wp-query-integration.php
```diff
@@ -298,6 +311,58 @@ public function filter_posts_request( $request, $query ) {
...
+		$post_objs = array();
+		$posts_ordered = array();
+		$grouped = array();
+		$order = array();
+
+		// group by site_id to decrease number of switch_to_blogs
+		foreach ( $post_list as $post_data ) {
+			$grouped[$post_data->site_id][] = $post_data;
+			// use this to keep the initial order of posts
+			$order[] = md5( $post_data->ID . ' ' . $post_data->post_date_gmt . ' ' . $post_data->site_id );
```
tlovett1: Why `md5` and `post_date_gmt`? Can't we just store like this:
```php
$order[] = $post_data->ID . '-' . $post_data->site_id;
```

classes/class-ep-wp-query-integration.php
```diff
@@ -298,6 +311,58 @@ public function filter_posts_request( $request, $query ) {
 	}
 
 	/**
+	 * Get post object data in multisite network.
+	 * Using post ID is better as get_post function pupulates all possible fields
+	 * So passing post ID and then merging with inital post object pupulates all expected fields
+	 * for WP_Post object
+	 *
+	 * @param  array $post_list list of found post objects
```
tlovett1: Please add @since 1.4

classes/class-ep-wp-query-integration.php
```diff
@@ -298,6 +311,58 @@ public function filter_posts_request( $request, $query ) {
...
+	 * @return array            list of WP_Post objects
+	 */
+	public function fetch_wp_post_object( $post_list ) {
```
tlovett1: Rename to `fill_post_objects`

EduardMaghakyan:
@tlovett1 I've added unit test to check the author ID, this should show the bug

classes/class-ep-wp-query-integration.php
EduardMaghakyan: Sure we can, I've removed that part md5 and post_date_gmt

classes/class-ep-wp-query-integration.php
EduardMaghakyan: Done

classes/class-ep-wp-query-integration.php
EduardMaghakyan: Added

PR #247, opened 2015-04-17
Fix wrong author ID in post object

jonathanbardo:
When using ElasticPress in the admin section of WP and doing a search, the user ID from get_post is set to zero even when Elastic Search has the ID.

This was tested and resolve the issue I am having.


tlovett1:
@jonathanbardo does #239 fix this?

PR #248, opened 2015-04-19
Bug/fix date query unit test

EduardMaghakyan:
Passing the time to before date to be inclusive fixed the failing tests, because if the before date doesn't have time it converts to 00:00:00 and that date won't be included, even if "inclusive" is set to true


tlovett1:
@EduardMaghakyan this is still failing.

EduardMaghakyan:
yes the test which fails is new one. I've added that for the bug with date_query which I've found. I can remove that for now and create another pull request for that, let me know what's better?

tlovett1:
I'll just merge.

PR #260, opened 2015-05-19
Adding filter to set default sortorder

HKandulla:
Hi,
I need use the $post->post_modified as the default sortorder and I havn't found a way to do this. It would be great if you add this feature to your plugin.

Thanks for the great work, Hannes


tlovett1:
Great addition!

PR #264, opened 2015-05-29
Change sync hook

jonathanbardo:
Hi! 

Right now a post would sync using the `transition_post_status` action which is a little early for other plugin (I'm looking at you ACF!). This would therefore not save all updated information on post metas.

Using the `wp_insert_post` action with a low priority would let all other plugin hook before we actually sync with the ElasticSearch index.

Any feedback greatly accepted!
Cheers!

### DIFF
--- a/classes/class-ep-sync-manager.php
+++ b/classes/class-ep-sync-manager.php
@@ -15,7 +15,7 @@ class EP_Sync_Manager {
 	 * @since 0.1.2
 	 */
 	public function setup() {
-		add_action( 'transition_post_status', array( $this, 'action_sync_on_transition' ), 10, 3 );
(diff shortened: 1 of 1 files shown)


psorensen:
Wow! I just ran into this issue 20 mins before you posted this patch. Pulled it and now ACF fields are being published. Thanks!

jonathanbardo:
@psorensen Glad it works for you!

tlovett1:
@jonathanbardo any idea why this test is failing?

jonathanbardo:
@tlovett1 I believe the build has errors on only one specific instance. Any way we could start the tests again in travis on your side?

tlovett1:
This looks solid.

PR #265, opened 2015-06-01
Fix fields not loaded from elasticsearch query

Stayallive:
The `post_excerpt` and `post_author` fields were not set in the query integration causing `the_author()` and related functions to fail and `the_excerpt()` to handle the excerpt like there is none.

### DIFF
--- a/classes/class-ep-wp-query-integration.php
+++ b/classes/class-ep-wp-query-integration.php
@@ -274,7 +274,9 @@ class EP_WP_Query_Integration {
 			$post->post_status = $post_array['post_status'];
 			$post->post_title = $post_array['post_title'];
 			$post->post_parent = $post_array['post_parent'];
+			$post->post_excerpt = $post_array['post_excerpt'];
 			$post->post_content = $post_array['post_content'];
+			$post->post_author = $post_array['post_author']['id'];
 			$post->post_date = $post_array['post_date'];
 			$post->post_date_gmt = $post_array['post_date_gmt'];
 			$post->post_modified = $post_array['post_modified'];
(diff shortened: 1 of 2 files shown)


Stayallive:
Hmm, I see the (very slow :P) tests are failing on this line...

https://github.com/10up/ElasticPress/blob/develop/tests/test-multisite.php#L1713

Should it be changed to:
```$this->assertEquals( $site['blog_id'], $post->post_author );```

I'm not totally sure why that field is chosen to test... something :confused:

**Comment by tlovett1** [comment]
Hmmm that is very strange. That test doesn't make sense at all.

**Comment by tlovett1** [comment]
@Stayallive the intention of this test was to verify each instance variable in the post object. Somehow it was forgotten. It would be great if you could fill it out. `site_id` is added to the post object here https://github.com/10up/ElasticPress/blob/develop/classes/class-ep-wp-query-integration.php#L266

**Comment by Stayallive** [comment]
@tlovett1, no problem, will do this tomorrow :)

**Comment by Stayallive** [comment]
@tlovett1, I looked at it again and it looks to me the test tests if the site id is set correctly when loading the post back in when in a multisite environment. Since I fixed the post_author field this also broke :)

I hope my patch to use another field fixed it, tests are a tad bit slow :zzz:... so we'll see!

**Comment by EduardMaghakyan** [comment]
@tlovett1 @Stayallive just a comment regarding test so this function 
```php
ep_create_and_sync_post( array( 'post_title' => 'findme', 'post_author' => $site['blog_id'] ) );
```
https://github.com/10up/ElasticPress/blob/develop/tests/test-multisite.php#L1689

Creates a post with author ID set to same as site ID, site_id is available in $post object as well as post_author, so I guess I'm missing something, but this test should just check whether the post author and site_id will match for that specific site, if both are populated correctly then they should match.

tlovett1:
Why are we setting things to the `site_id` anyway?

Stayallive:
For multisite purposes? This is just to confirm that a query over multiple sites also works, with the correct id's loaded back. Since the `post_author` now is actually loaded from the `author.id` field (which is not set in this test), the test failed. Should be fixed now and doing the same thing as before!?

PR #266, opened 2015-06-01
Add --offset paramater to cli index command

Stayallive:
This allows to resume a index action when failed or stopped before finishing, without restarting the index operation.

PR #269, opened 2015-06-01
Add --offset argument to cli index command

Stayallive:
This allows to resume a index action when failed or stopped before finishing, without restarting the index operation.

Redo of #266 :)


tlovett1:
Thank you!!!

Issue #263, opened 2015-05-27
Supporting 'like' query in post meta


tuanmh:
It seems that currently EP does not support 'LIKE' query against post meta data. This is actually a need as sometimes we don't want to match exact by using 'term' but using 'match'. This could be really handy. I can do a PR if desired.

tlovett1:
Have you looked at the `search_fields` parameter?

tuanmh:
Yes. I'm using it for one of my post_meta fields. However, in our app, we have a search query which queries against the search fields and another query using in filter. For example:
- ``Introduction``: main query on fields ``post title, post content, post_meta.subtitle``
- ``elasticpress``: a filter, something like ``elasticpress`` on other fields, e.g. ``post_meta.ABC``

So it's basically the same as other operators but it requires a 'match' not 'term'.

tuanmh:
Something like this commit (05a8b0b) would be really appreciated. 

```php
						case 'like':
							if ( isset( $single_meta_query['value'] ) ) {
								$terms_obj = array(
									'query' => array(
										"match" => array(
											'post_meta.' . $single_meta_query['key'] => $single_meta_query['value'],
										)
									),
								);
							}
							break;
```

tlovett1:
@tuanmh can you open a PR?

tuanmh:
@tlovett1 I will add the test ASAP.

tuanmh:
Closing via c142064 and 99d334c

PR #270, opened 2015-06-01
Enhancement for #263 Support like query in metadata, in addition with main query

tuanmh:
- Main query can actually do a like query to meta data. However,
sometimes we need to have both main query and meta data query. They
are, of course, different, hence, it would be great if we can use meta
data query as a filter for main query and this meta data query supports
LIKE rather than normal ‘term’ filter.

### DIFF
--- a/classes/class-ep-api.php
+++ b/classes/class-ep-api.php
@@ -877,6 +877,17 @@ class EP_API {
 							}
 
 							break;
+						case 'like':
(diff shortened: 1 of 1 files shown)


tuanmh:
related #263

tlovett1:
@tuanmh can you add a test for this please?

tuanmh:
will do!

Issue #258, opened 2015-05-17
MySQL db is still be used


crebacz:
Hello--

I'm working on elasticpress in my local environment. When I perform a search, the wordpress database looks like it is still being hit...and a lot.  Based on Query Monitor, I'm getting over 900 queries to the database. Same search request with elasticpress deactivated is 53 queries of the database.  From the docs, it sounded like WP_Query is only run if elasticsearch fails.  However, I know I'm getting results from elasticsearch because I'm using terms (slightly misspelled) that return no results with default wordpress search.

Here is an example of the type of query that is being run (basically it is search each individual post return for each taxonomy--even those I didn't specify). 776 of the queries are wp_get_object_terms() calls.  

```
SELECT t.*, tt.* FROM wp_terms AS t INNER JOIN wp_term_taxonomy AS tt ON tt.term_id = t.term_id INNER JOIN wp_term_relationships AS tr ON
tr.term_taxonomy_id = tt.term_taxonomy_id WHERE tt.taxonomy IN ('hardware') AND tr.object_id IN (10369) ORDER BY t.name ASC;

```

The query that I'm running in search.php is: 

```
$paged = (get_query_var('paged')) ? get_query_var('paged') : 1;
            $search = new WP_Query(
                    array(
                    's' => get_search_query(),
                    'search_fields' => array(
                            'post_title',
                            'post_content',
                            'post_excerpt',
                        ),
                    'post_type' => array('post','page','equipment', 'policy'),
                    'posts_per_page' => 20,
                    'paged' => $paged,
                    'tax_query' => array(
                        array(
                            'taxonomy'  => 'category',
                            'terms'     => array('network-policy'),
                            'field'     => 'slug'
                            )
                        )
                    )
                );
  ```

I did also try to run a similar query from a pre_get_posts action.  This also resulted in over 900 queries of the wordpress database.  

```
            $search_fields = array( 'post_title', 'post_content','post_excerpt', );
            $post_types = array('page','post','equipment');
            $numPosts = 20;
            $tax_query = array(
                array(
                    'taxonomy'  => 'category',
                    'terms'     => array('network-policy'),
                    'field'     => 'slug'
                    )
                );

            $query->set('search_fields', $search_fields);
            $query->set('paged', $paged);
            $query->set('posts_per_page', $numPosts);
            $query->set('tax_query', $tax_query);
```

[...]

PR #273, opened 2015-06-03
Mysql db queries

EduardMaghakyan:
Remove get_post function calls and populate ```post``` object from EP index to meet #258 enhancement request


tlovett1:
This looks great.

PR #278, opened 2015-06-03
Feature/fallback hosts

ChrisWiegman:
This gives the option to have one or more fallback servers should the primary ES host fail. Fallback servers can be specified in a global array $eb_backup_hosts. Upon failure of the primary host the backup hosts will be checked sequentially until a good host is found or an error will be returned. Good hosts are cached for an hour before the process is checked again.

### DIFF
--- a/classes/class-ep-api.php
+++ b/classes/class-ep-api.php
@@ -36,13 +36,17 @@ class EP_API {
 	 */
 	public function index_post( $post ) {
 
-		$index_url = ep_get_index_url();
+		$index = trailingslashit( ep_get_index_name() );
 
-		$url = $index_url . '/post/' . $post['post_id'];
+		$path = $index . 'post/' . $post['post_id'];
 
-		$request_args = array( 'body' => json_encode( $post ), 'method' => 'PUT', 'timeout' => 15, 'headers' => $this->format_request_headers() );
+		$request_args = array(
+			'body'    => json_encode( $post ),
+			'method'  => 'PUT',
+			'timeout' => 15,
+		);
 
-		$request = wp_remote_request( $url, apply_filters( 'ep_index_post_request_args', $request_args, $post ) );
+		$request = ep_remote_request( $path, apply_filters( 'ep_index_post_request_args', $request_args, $post ) );
 
 		if ( ! is_wp_error( $request ) ) {
 			$response_body = wp_remote_retrieve_body( $request );
@@ -71,9 +75,10 @@ class EP_API {
 	 * @return bool
 	 */
 	public function refresh_index() {
-		$request_args = array( 'method' => 'POST', 'headers' => $this->format_request_headers() );
 
-		$request = wp_remote_request( ep_get_index_url() . '/_refresh', apply_filters( 'ep_refresh_index_request_args', $request_args ) );
+		$request_args = array( 'method' => 'POST' );
+
+		$request = ep_remote_request( '_refresh', apply_filters( 'ep_refresh_index_request_args', $request_args ) );
 
 		if ( ! is_wp_error( $request ) ) {
 			if ( isset( $request['response']['code'] ) && 200 === $request['response']['code'] ) {
@@ -105,15 +110,20 @@ class EP_API {
 			foreach ( $scope as $site_id ) {
 				$index[] = ep_get_index_name( $site_id );
 			}
+
+			$index = implode( ',', $index );
+		} else {
+			$index = ep_get_index_name();
 		}
 
-		$index_url = ep_get_index_url( $index );
+		$path = $index . '/post/_search';
 
-		$url = $index_url . '/post/_search';
+		$request_args = array(
+			'body'    => json_encode( $args ),
+			'method'  => 'POST',
+		);
 
-		$request_args = array( 'body' => json_encode( $args ), 'method' => 'POST', 'headers' => $this->format_request_headers() );
-
-		$request = wp_remote_request( $url, apply_filters( 'ep_search_request_args', $request_args, $args, $scope ) );
+		$request = ep_remote_request( $path, apply_filters( 'ep_search_request_args', $request_args, $args, $scope ) );
 
 		if ( ! is_wp_error( $request ) ) {
 
@@ -182,13 +192,14 @@ class EP_API {
 	 * @return bool
 	 */
 	public function delete_post( $post_id  ) {
-		$index_url = ep_get_index_url();
 
-		$url = $index_url . '/post/' . $post_id;
+		$index = trailingslashit( ep_get_index_name() );
 
-		$request_args = array( 'method' => 'DELETE', 'timeout' => 15, 'headers' => $this->format_request_headers() );
+		$path = $index . '/post/' . $post_id;
 
-		$request = wp_remote_request( $url, apply_filters( 'ep_delete_post_request_args', $request_args, $post_id ) );
+		$request_args = array( 'method' => 'DELETE', 'timeout' => 15 );
+
+		$request = ep_remote_request( $path, apply_filters( 'ep_delete_post_request_args', $request_args, $post_id ) );
 
 		if ( ! is_wp_error( $request ) ) {
 			$response_body = wp_remote_retrieve_body( $request );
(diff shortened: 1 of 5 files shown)


tlovett1:
@ChrisWiegman can we add tests?

ChrisWiegman:
Good point. I'm usually told not to add tests with Voltron work. Let me see about doing something there.

ChrisWiegman:
Added some tests to verify the ep_get_host function. Tests ran successfully. Note this is a rather strange scenario as we're defining the primary EP_HOST in bootstrap by necessity however this would normally make testing the function difficult. To overcome I've introduced the use of an extra constant that forces the bypass of the defined host in favor of the backup array. This should be CRed before further push.

tlovett1:
@ChrisWiegman seeing two things that might be issues
1. `ep_get_hosts` only seems to be used for some of the API methods. For example, it doesn't seem to be used in search?
1. I think we can be a bit more clever with the number of requests sent to the ES server. Instead of checking if a host is available in `ep_get_hosts`, why not just check the response code of `wp_remote_request`? If it 404's, try the next host. This is what I meant by a `wp_remote_request` wrapper.

ChrisWiegman:
Gotcha. Still getting familiar so it is highly possible I missed some. I'll replace the wp_http calls instead which will help. I also realized this morning that it wouldn't automatically fail over for up to an hour when I looked again this morning. Let me clean this up and I'll resubmit.

classes/class-ep-api.php
```diff
@@ -1236,6 +1241,44 @@ public function elasticsearch_alive() {
...
+	 * @since 1.5
+	 *
+	 * @param string $url  Site URL to retrieve.
+	 * @param array  $args Optional. Request arguments. Default empty array.
+	 *
+	 * @return WP_Error|array The response or WP_Error on failure.
+	 */
+	public function remote_request( $url, $args = array() ) {
+
+		$host = ep_get_host();
```
tlovett1: Where does this get used?

classes/class-ep-api.php
```diff
@@ -1236,6 +1241,44 @@ public function elasticsearch_alive() {
...
+	 * (if present) should the primary EP host fail.
+	 *
+	 * @since 1.5
+	 *
+	 * @param string $url  Site URL to retrieve.
+	 * @param array  $args Optional. Request arguments. Default empty array.
+	 *
+	 * @return WP_Error|array The response or WP_Error on failure.
+	 */
+	public function remote_request( $url, $args = array() ) {
```
tlovett1: Seems like it makes sense to use path instead of url. Instead of passing the full index URL to each remote_request call, calculate the full URL in here.

classes/class-ep-api.php
ChrisWiegman: In an attempt to maintain consistency with existing code there is an accessor function, ep_remote_request, that makes use of it.

classes/class-ep-api.php
ChrisWiegman: I could do that, the existing code seemed to have a couple different methods for calculating the $url, when first written I actually was going to extract it however in order to make it more compatible with what is there I elected to take this approach.

classes/class-ep-api.php
```diff
@@ -387,6 +409,10 @@ public function prepare_post( $post_id ) {
 		$post_date_gmt = $post->post_date_gmt;
 		$post_modified = $post->post_modified;
 		$post_modified_gmt = $post->post_modified_gmt;
+		$comment_count = absint( $post->comment_count );
```
tlovett1: This should be in a separate PR.

classes/class-ep-api.php
```diff
@@ -543,11 +573,23 @@ public function prepare_meta( $post ) {
 	 */
 	public function delete_index( $index_name = null ) {
 
-		$index_url = ep_get_index_url( $index_name );
+		if ( null === $index_name ) {
+
+			$index = ep_get_index_name();
+
+		} elseif ( is_array( $index_name ) ) {
```
tlovett1: Array support should be in a separate PR.

classes/class-ep-api.php
```diff
@@ -568,11 +610,24 @@ public function delete_index( $index_name = null ) {
...
 	 */
 	public function index_exists( $index_name = null ) {
-		$index_url = ep_get_index_url( $index_name );
 
-		$request_args = array( 'headers' => $this->format_request_headers() );
+		if ( null === $index_name ) {
+
+			$index = ep_get_index_name();
+
+		} elseif ( is_array( $index_name ) ) {
```
tlovett1: Array support should be in a separate PR

classes/class-ep-api.php
```diff
@@ -636,13 +691,17 @@ public function format_args( $args ) {
...
-			$formatted_args['sort'] = array(
+			$default_sort = array(
 				array(
 					'_score' => array(
 						'order' => $order,
 					),
 				),
 			);
+
+			$default_sort = apply_filters( 'ep_set_default_sort', $default_sort, $order );
```
tlovett1: Wondering if maybe you should rebase this PR?

classes/class-ep-api.php
```diff
@@ -877,6 +936,17 @@ public function format_args( $args ) {
 							}
 
 							break;
+						case 'like':
```
tlovett1: Wondering if maybe you should rebase this PR?

classes/class-ep-api.php
```diff
@@ -1236,6 +1316,61 @@ public function elasticsearch_alive() {
...
+		//If we have a failure we'll try it again with a backup host
+		if ( false === $request || is_wp_error( $request ) || ( isset( $request['response']['code'] ) && 200 !== $request['response']['code'] ) ) {
+
+			$host = ep_get_host( true, $use_backups );
+
+			if ( is_wp_error( $host ) ) {
+				return $host;
+			}
+
+			return wp_remote_request( esc_url( trailingslashit( $host ) . $path ), $args );
```
tlovett1: Seems like this only supports one backup host. I was thinking more like a loop of all backup hosts with a filter to limit the amount of attempts.

classes/class-ep-config.php
```diff
@@ -19,6 +19,67 @@ public static function factory() {
...
+	public function get_ep_host( $force = false, $use_only_backups = false ) {
+
+		global $ep_backup_host;
+
+		// Delete the transient if we want to force a new good host lookup
+		if ( true === $force ) {
+			delete_transient( 'ep_last_good_host' );
+		}
+
+		$last_good_host = get_transient( 'ep_last_good_host' );
```
tlovett1: Remember this is sometimes intended to run site-wide. Might make more sense to use site transients.

classes/class-ep-api.php
ChrisWiegman: The odd thing about this one is it is present before my changes (see https://github.com/10up/ElasticPress/blob/aafc3999487598347335b2dc38b9979b8c7eff74/classes/class-ep-api.php#L394) so as to why it keeps showing as a newline in this changeset is interesting. I'll see if I can get that one out in the rebase.

classes/class-ep-api.php
ChrisWiegman: Fixed

classes/class-ep-config.php
ChrisWiegman: Excellent point

classes/class-ep-api.php
ChrisWiegman: Rebasing complete. (GH really needs a better way to check these off before a push. Push coming shortly).

classes/class-ep-api.php
ChrisWiegman: Rebasing complete

classes/class-ep-api.php
ChrisWiegman: Removed for now. Will do a new PR after this is merged

classes/class-ep-api.php
ChrisWiegman: It actually does the loop, just not here. The first time it calls ep_get_host it will use the transient (if available). Then, the host is verified and if it fails it goes back to ep_get_host with the force param set to true. It is here that it will loop through all the available hosts until a success is found. This just makes it slightly more efficient in the case of a cached host.

tlovett1:
This looks solid. I'm merging. @allan23 mind spending a little time QAing this?

PR #287, opened 2015-06-12
Improve error reporting for bad responses from ES server

### LINKED ISSUE #286: On index failure (request entity too large), no notification
Even though the `_bulk` index request failed, it was not reported during the index operation.

This should be addressed for better debugging and troubleshooting.

AaronHolbrook:
Fixes #286 

@tlovett1 #reviewmerge

### DIFF
--- a/classes/class-ep-api.php
+++ b/classes/class-ep-api.php
@@ -1103,7 +1103,17 @@ class EP_API {
 
 		$request = wp_remote_request( $url, apply_filters( 'ep_bulk_index_posts_request_args', $request_args, $body ) );
 
-		return is_wp_error( $request ) ? $request : json_decode( wp_remote_retrieve_body( $request ), true );
+		if ( is_wp_error( $request ) ) {
+			return $request;
+		}
+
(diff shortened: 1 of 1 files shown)


classes/class-ep-api.php
```diff
@@ -1103,7 +1103,17 @@ public function bulk_index_posts( $body ) {
...
 
-		return is_wp_error( $request ) ? $request : json_decode( wp_remote_retrieve_body( $request ), true );
+		if ( is_wp_error( $request ) ) {
+			return $request;
+		}
+
+		$response = wp_remote_retrieve_response_code( $request );
+
+		if ( 200 !== $response ) {
+			WP_CLI::error( wp_remote_retrieve_response_message( $request ) );
```
tlovett1: I'm concerned that a WP_CLI method is being called outside of wp-cli.php. I think this should be restructured a bit to return an error that is then passed to WP_CLI::error in wp-cli.php

AaronHolbrook: Good point - will do so

Issue #249, opened 2015-04-22
Add security X- header to requests


zamoose:
It'd be great to have an `X-` header sent along with every* EP request so that, in addition to locking access down via firewall we can secure access to individual Elasticsearch instances with a few Nginx rules.

I'm envisioning setting something like 

* `X-ElasticPress-API-key`
* with a payload of some md5 hash or similar that is a pre-shared key between the ES and EP sides
* set it in `wp-config.php` via a `define( 'EP_KEY', [gnarly key] );` 

*(At the very least, protecting indexing/PUT/POST requests via this methodology would get us a long way towards this)

This idea is based on conversations with @TheLastCicada and @tott.

zamoose:
Additional thoughts:

* EP should ignore this header if one is unspecified in the config. It should not be mandatory.
* Nginx can check for the presence of the header and check it against its shared value and reject the traffic if an incorrect key is supplied or no header is sent when one is expected.
* We'd need some convention for the Nginx side of things to send back so that EP can coherently handle error conditions. If it's just a simple HTTP 403, that might suffice, but more granularity would likely be good.

zamoose:
...Or we could just use my favorite error code, [HTTP 418](http://en.wikipedia.org/wiki/Hyper_Text_Coffee_Pot_Control_Protocol#Commands_and_replies).

tlovett1:
Are you defining "EP Request" as EP making a request to Elasticsearch?

zamoose:
Yup, a request from EP to the Elasticsearch server/cluster. Ideally, the "potentially destructive" requests can be covered, but if we want to also maintain some form of access control so Joe Sixpack from the Internet can't run JSON queries against even the read-only indexes, then all requests should probably be covered.

tlovett1:
So here is the logic I see:
If and only if `EP_API_KEY` is defined, included it in ALL requests to ES as a `X-ElasticPress-API-Key` header.

Does this work for you?

zamoose:
That would be brilliant.

zamoose:
Circling back around to this one:
I've just been using this in a test setup -- it works like a charm.

PR #290, opened 2015-06-12
WP_CLI GET Calls missing headers

ChrisWiegman:
In #249 the option for a security header (API key) was added however it does not function on the stats and status calls when called by WP-CLI as these calls don't use the header. This fixes the issue and adds a callable function for getting the headers in future classes.


tlovett1:
Nice!

Issue #272, opened 2015-06-02
--network-wide commands stop after 100 sites


foofy:
The index, put-mapping, and delete-index commands, when used with the --network-wide flag, stop working after 100 sites. This is because EP_API::get_sites() uses wp_get_sites() which only returns 100 sites by default:

https://github.com/10up/ElasticPress/blob/develop/classes/class-ep-api.php#L1063

AaronHolbrook:
Nice catch @foofy

tlovett1:
I think we just need to document this. Thoughts?

AaronHolbrook:
I think this should be documented and a filter added for the arguments array that is used to call `wp_get_sites()`.

https://github.com/10up/ElasticPress/blob/develop/classes/class-ep-api.php#L1088

Right now you could filter it yourself and return it, but it's redundant and wasteful. A filter would be a good fix.

AaronHolbrook:
cc @GhostToast

PR #294, opened 2015-06-16
Include arguments to the ep_get_sites

bordoni:
- New Filter `ep_indexable_sites_args` to allow control over the whole `wp_get_sites`
- Easy usage of --network-wide=INT to change the limit, defaults to 0 where it has no LIMIT

_Resolve:_ #272


tlovett1:
I think this makes sense. We will need to update documentation.

PR #299, opened 2015-06-19
Run unit tests again.

ocean90:
This reverts parts of e8c82bb3c72617dd4db19513bb3eadc58cb748cd since `ABSPATH` isn't (always) defined for tests, especially not in `tests/bootstrap.php`.

### DIFF
--- a/tests/bootstrap.php
+++ b/tests/bootstrap.php
@@ -1,9 +1,5 @@
 <?php
 
- if ( ! defined( 'ABSPATH' ) ) {
-    exit; // Exit if accessed directly.
-}
-
 $_tests_dir = getenv( 'WP_TESTS_DIR' );
 if ( ! $_tests_dir ) {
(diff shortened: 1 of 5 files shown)


tlovett1:
Does it make sense to just define ABSPATH in the bootstrap file?

ocean90:
@tlovett1 Tests and build files shouldn't be on production systems IMO. So they should be remove from the plugin package on wordpress.org.

joelgarciajr84:
What about this? Doesn't make sense? https://travis-ci.org/10up/ElasticPress/builds/67277556 :v:

ocean90:
@joelgarciajr84 It does make sense because `exit` is equal to `exit(0)`, means PHPUnit had finished normally.

joelgarciajr84:
@ocean90  what about insert a empty index.php file in all directories, this is  a good pratice?

tlovett1:
@joelgarciajr84 no need for empty index.php

PR #300, opened 2015-06-19
Add support for category_name queries

ocean90:
WordPress supports searches via the `category_name` argument. The arg is set on pages like example.com/category/foo/?s=bar.
I think ElasticPress should support this type of queries too.

Workaround:
```php
add_filter( 'ep_formatted_args', 'ds_category_name_searches', 10, 2 );
function ds_category_name_searches( $formatted_args, $args ) {

	if ( ! empty( $args[ 'category_name' ] ) ) {
		$terms_obj = array(
			'terms.category.slug' => array( $args[ 'category_name' ] ),
		);

		$formatted_args['filter']['and'][]['bool']['must'] = array(
			'terms' => $terms_obj
		);
	}

	return $formatted_args;
}

### DIFF
--- a/classes/class-ep-api.php
+++ b/classes/class-ep-api.php
@@ -706,6 +706,23 @@ class EP_API {
 		}
 
 		/**
+		 * 'category_name' arg support.
+		 *
+		 * @since 1.5
+		 */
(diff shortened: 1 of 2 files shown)


tlovett1:
Love how you created a test for this.

PR #305, opened 2015-06-25

### LINKED ISSUE #304: Improvement: allow recursively terms indexing

tuanmh:
Related #304 
Subject to unit-tests though.


tuanmh:
@tlovett1 sorry but last second commit should be separated in another PR. I reverted it.

PR #308, opened 2015-07-07
Use the `ep_post_sync_kill` filter for bulk index

Stayallive:
Added the `ep_post_sync_kill` filter when queing posts too, otherwise they get indexed anyway.


tlovett1:
Solid.

PR #314, opened 2015-07-20
Extra check for proper host.

ChrisWiegman:
After running ep_get_host() the elasticsearch_alive() method will still attempt to reach the host even if a wp_error is returned due to a failure of all available hosts. This fix will force a 2nd host check for backups and then return false instead of the wp_error();


allan23:
@tlovett1 This corrects a bug with the fallback hosts.

PR #315, opened 2015-07-21

### LINKED ISSUE #304: Improvement: allow recursively terms indexing

tuanmh:


allan23:
@tuanmh Any chance you can rebase?

tuanmh:
@allan23 rebased from develop, first time I did it so hopefully it is as you wished.

tlovett1:
@tuanmh can we get this squashed into no more than a few commits? Also, there are some formatting issues.

allan23:
@tuanmh This might help in regards to the code formatting: https://make.wordpress.org/core/handbook/best-practices/coding-standards/php/

PR #317, opened 2015-07-21
Allow indexing for posts have "0000-00-00 00:00:00" date value

tuanmh:
- Draft, pending, auto-draft post will have post_date_gmt set to:
“0000-00-00 00:00:00”. This value is not parseable by ElasticSearch,
hence, we will need to set it to null value to get indexing gone
through.
http://cl.ly/image/2J2U0H3u2f3c
- This will also apply for post_modified_gmt

### DIFF
--- a/tests/test-single-site.php
+++ b/tests/test-single-site.php
@@ -34,7 +34,6 @@ class EPTestSingleSite extends EP_Test_Base {
 	public function tearDown() {
 		parent::tearDown();
 
-		remove_filter( 'ep_indexable_post_status', array( $this, 'mock_indexable_post_status' ), 10);
 		$this->fired_actions = array();
 	}
 
@@ -1679,5 +1678,6 @@ class EPTestSingleSite extends EP_Test_Base {
 			$this->assertNull( $post[ 'post_modified_gmt' ] );
 		}
 		$this->assertNotNull( $post );
+		remove_filter( 'ep_indexable_post_status', array( $this, 'mock_indexable_post_status' ), 10);
 	}
 }
--- a/tests/test-single-site.php
+++ b/tests/test-single-site.php
@@ -34,6 +34,7 @@ class EPTestSingleSite extends EP_Test_Base {
 	public function tearDown() {
 		parent::tearDown();
 
+		remove_filter( 'ep_indexable_post_status', array( $this, 'mock_indexable_post_status' ), 10);
 		$this->fired_actions = array();
 	}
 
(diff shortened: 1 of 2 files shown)


tlovett1:
@tuanmh what problem is this solving?

tuanmh:
@tlovett1 sorry did not make it clear. For some ``draft`` posts, the ``post_date_gmt`` will be set to ``0000-00-00 00:00:00``, and this value cannot be parsed by ElasticSearch, hence, the indexing could fail if users allow ``draft`` posts to be indexed. The error could be something like this image:

image

TheLastCicada:
We actually had this same issue where the post_date_gmt was "0000-00-00 00:00:00" on a site with a bunch of posts imported from Joomla - these posts wouldn't index and threw a bunch of errors.  Granted, having the post_date_gmt unset at "0000-00-00 00:00:00" is probably not the "right" thing to have set in the database for these posts, but the only thing it caused trouble with was indexing for ElasticPress.  If ElasticPress handled this gracefully and just indexed the posts without error, it would have been a pretty big help in this situation.  

@tuanmh - I'm curious - in what cases a user would want to index draft posts?  I can't think of a use case where you'd be doing this.  I suppose some sort post manager that is built outside the admin or something.....

tuanmh:
> I'm curious - in what cases a user would want to index draft posts? I can't think of a use case where you'd be doing this.

Ah, in our system, users can not only find published posts but also drafts that they have permission to view/edit - just provide an ease for them to navigate and interact with the system. In addition, since we integrate ElasticSearch deeply in admin dashboard so having drafts indexed is essentially a must.

TheLastCicada:
@tuanmh Sounds like a pretty neat setup :)  Thanks for explaining!

tlovett1:
@tuanmh generally this looks good. Two nit picky things:

1. Please use === instead of == to do strict comparisons.
2. No need to define the constant. Just compare the variable against the time string.

Also, what do you think about doing this same check on `post_date`, `post_modified`, and `post_modified_gmt`?

tuanmh:
> Please use === instead of == to do strict comparisons.

Silly me! Amended :) 



> No need to define the constant. Just compare the variable against the time string.

Done.


> Also, what do you think about doing this same check on post_date, post_modified, and post_modified_gmt?

I added same check for post_date and post_modified. ``post_modified_gmt`` had been added in first commit. Thanks for your feedback!

tests/test-single-site.php
```diff
@@ -34,6 +34,7 @@ public function setUp() {
 	public function tearDown() {
 		parent::tearDown();
 
+		remove_filter( 'ep_indexable_post_status', array( $this, 'mock_indexable_post_status' ), 10);
```
tlovett1: Can you just put this at the end of the test method?

tuanmh: Done!

tlovett1:
Thanks!

PR #319, opened 2015-07-22
Enable indexing of attachments

tlovett1:
Simple PR to enable indexing of attachments.


tlovett1:
Just need to document this properly. As @ChrisWiegman pointed out we may want to have this disabled at first since  it is a change in functionality.

ChrisWiegman:
Hey @tlovett1! Trying this out this morning and while it works when adding a new attachment or editing an existing attachment it does not seem to work when performing a site index (say via WP_CLI)

PR #320, opened 2015-07-24

### LINKED ISSUE #304: Improvement: allow recursively terms indexing

tuanmh:
Relate #304

### DIFF
--- a/classes/class-ep-api.php
+++ b/classes/class-ep-api.php
@@ -530,6 +530,8 @@ class EP_API {
 
 		$terms = array();
 
+		$allow_hierarchy = apply_filters( 'ep_sync_terms_allow_hierarchy', false );
+
 		foreach ( $selected_taxonomies as $taxonomy ) {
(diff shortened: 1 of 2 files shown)


classes/class-ep-api.php
```diff
@@ -537,20 +539,50 @@ private function prepare_terms( $post ) {
...
-					'parent'  => $term->parent
-				);
+				if( ! isset( $terms_dic[ $term->term_id ] ) ) {
+					$terms_dic[ $term->term_id ] = array(
+						'term_id' => $term->term_id,
+						'slug'    => $term->slug,
+						'name'    => $term->name,
+						'parent'  => $term->parent
+					);
+					if( $allow_hierarchy ){
```
tlovett1: Shouldn't this be an if...else since we are immediately overwriting `$terms_dic`?

tlovett1: Ah nevermind

PR #339, opened 2015-08-31
Add configuration to choose to index all attachments at index time

sambonner:
Hi, my apologies if this patch should be applied within https://github.com/10up/ElasticPress/pull/319, I wasn't sure as I see that it has already been merged.

Anyway, this patch essentially just adds a new ElasticPress config  (EP_INDEX_ATTACHMENTS) that, if defined in wp-config.php will cause all attachments to indexed on initial ElasticSearch index. I believe something like this was discussed here - https://github.com/10up/ElasticPress/pull/319#commitcomment-12347070

I'm unclear if this is actually the appropriate place to be checking runtime configuration, so suggestions happily accepted.

Thanks,
Sam

### DIFF
--- a/classes/class-ep-config.php
+++ b/classes/class-ep-config.php
@@ -125,7 +125,12 @@ class EP_Config {
 	 * @return array
 	 */
 	public function get_indexable_post_status() {
-		return apply_filters( 'ep_indexable_post_status', array( 'publish' ) );
+    if ( defined( 'EP_INDEX_ATTACHMENTS' ) ) {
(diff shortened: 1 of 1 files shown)


tlovett1:
I like this idea. A couple things needed here:

1. Not only do we need to check if the constant is defined but also if it evals to true.
2. The constant also needs to filter `ep_indexable_post_types` and add the `attachment` post type (I think).

Thanks for the PR!

tlovett1:
@sambonner any update on this?

tlovett1:
Closing this. Feel free to reopen.

PR #347, opened 2015-09-14
Fix posts_per_page typo, keeping backwards compatibility

### LINKED ISSUE #346: posts_per_page bug in argument handling
https://github.com/10up/ElasticPress/blob/develop/classes/class-ep-api.php#L678

There's a typo, it should be `'posts_per_page'` instead of `'post_per_page'`. Will submit PR.

sc0ttkclark:
Fixes #346

### DIFF
--- a/classes/class-ep-api.php
+++ b/classes/class-ep-api.php
@@ -674,7 +674,10 @@ class EP_API {
 	 * @return array
 	 */
 	public function format_args( $args ) {
-		if ( ! empty( $args['post_per_page'] ) ) {
+		if ( ! empty( $args['posts_per_page'] ) ) {
+			$posts_per_page = $args['posts_per_page'];
+		} elseif ( ! empty( $args['post_per_page'] ) ) {
+			// For backwards compatibility for those using this since EP 1.4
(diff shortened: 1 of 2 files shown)


sc0ttkclark:
Just rereading the code, further down you can see that `posts_per_page` is reference and then set for `size`, so there's some extra logic here that seems unnecessary for `post_per_page`.

sc0ttkclark:
Dockunit passes, just says it's errored.

This doesn't really change anything, it just keeps backwards compatibility for the typo arg name for `post_per_page` and cleans up that logic so it's not at the top anymore.

Should be an easy merge.

tlovett1:
Looks pretty good. Two things:

1. I'd like to do the check before calling `get_option` so we can avoid that if possible.
2. This change would be in EP 1.7

tlovett1:
@sc0ttkclark see above

allan23:
@sc0ttkclark Any update?

sc0ttkclark:
PR updated with get_option conditional added back, except it's now looking at the proper posts_per_page argument.

PR #348, opened 2015-09-14
Use get_post() if we're on the current site

### LINKED ISSUE #345: Pulling post data from current db instead of only from EP
https://github.com/10up/ElasticPress/blob/develop/classes/class-ep-wp-query-integration.php#L277

On this line, we setup a faux post object, but it's not a real `WP_Post` object. I feel like we should either enable or by default use `$post = get_post( $post_array['post_id'] );` and setting `$post->site` and `$post->elasticpress` after that. If `$post` is empty, then continue and skip that post, as EP is returning a post that doesn't exist.

Getting a post from `get_post()` won't work as expected if the scope is a different site than current blog ID, so we could add logic there to use a faux WP_Post object in that case by setting up a `$faux_post` object and using `$post = new WP_Post( $faux_post )` to get a real live WP_Post.

sc0ttkclark:
Open to further tweaks, like making this optional, if you *only* want to pull from EP, and not ping the WP object cache / DB to see if the post actually still exists.

Fixes #345

### DIFF
--- a/classes/class-ep-wp-query-integration.php
+++ b/classes/class-ep-wp-query-integration.php
@@ -329,7 +329,7 @@ class EP_WP_Query_Integration {
 				}
 
 				if ( $post ) {
-					$post->site = $post_site_id;
+					$post->site_id = $post_site_id;
 					$post->elasticsearch = true; // Super useful for debugging
 
 					$new_posts[] = $post;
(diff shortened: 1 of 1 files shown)


sc0ttkclark:
I need to look into the test that failed to further tweak things

allan23:
@sc0ttkclark Any update?

sc0ttkclark:
No update on this one, it's not super high priority for our project since the post data is usually in sync. More work needs to be done to debug exactly why it's failing tests right now.

sc0ttkclark:
Made a refreshed branch and PR at #447

Issue #328, opened 2015-08-07
Query posts by thumbnail not available because it's a protected meta key


sc0ttkclark:
The EP_API::prepare_meta method removes meta keys which are protected. That unfortunately leaves out the ability to still query by posts that have thumbnails (a common occurrence).

Original:

```php
	/**
	 * Prepare post meta to send to ES
	 *
	 * @param object $post
	 *
	 * @since 0.1.0
	 * @return array
	 */
	public function prepare_meta( $post ) {
		$meta = (array) get_post_meta( $post->ID );

		if ( empty( $meta ) ) {
			return array();
		}

		$prepared_meta = array();

		foreach ( $meta as $key => $value ) {
			if ( ! is_protected_meta( $key ) ) {
				$prepared_meta[$key] = maybe_unserialize( $value );
			}
		}

		return $prepared_meta;
	}
```

With filter and _thumbnail_id added would be something like this (with inline hook doc):

```php
	/**
	 * Prepare post meta to send to ES
	 *
	 * @param object $post
	 *
	 * @since 0.1.0
	 * @return array
	 */
	public function prepare_meta( $post ) {
		$meta = (array) get_post_meta( $post->ID );

		if ( empty( $meta ) ) {
			return array();
		}

		$prepared_meta = array();

		$allowed_protected_keys = apply_filters( 'ep_prepare_meta_allowed_protected_keys', array( '_thumbnail_id' ), $post );

		foreach ( $meta as $key => $value ) {
			if ( ! is_protected_meta( $key ) || in_array( $key, $allowed_protected_keys ) ) {
				$prepared_meta[$key] = maybe_unserialize( $value );
			}
		}

		return $prepared_meta;
	}
```

FreshFlesh:
Because this is very specific and someone could argue that some other protected metas should also be indexed by default, may be it is better to set this meta as unprotected using is_protected_meta filter?

```php
add_filter( 'is_protected_meta', 'my_index_thumbnail_meta', 10, 3 );

function my_index_thumbnail_meta( $protected, $meta_key, $meta_type ) {
    if ( $meta_key == '_thumbnail_id' ) {
        $protected = false;
    }
    return $protected;
}
```

sc0ttkclark:
I wouldn't want to change how WordPress it works itself here -- is_protected_meta would do that.

[...]

PR #351, opened 2015-09-16
Adds a filter to allow the indexing of specified private meta keys.

ChrisWiegman:
This is similar to #328 and based off @sc0ttkclark 's code but does not include any default private fields to index.

### DIFF
--- a/classes/class-ep-api.php
+++ b/classes/class-ep-api.php
@@ -625,9 +625,24 @@ class EP_API {
 		 */
 		$allowed_protected_keys = apply_filters( 'ep_prepare_meta_allowed_protected_keys', array(), $post );
 
-		foreach ( $meta as $key => $value ) {
-			if ( ! is_protected_meta( $key ) || true === $allowed_protected_keys || in_array( $key, $allowed_protected_keys ) ) {
-				$prepared_meta[$key] = maybe_unserialize( $value );
+		/**
+		 * Filter non-indexed public meta
+		 *
+		 * Allows for specifying public meta keys that should be excluded from the ElasticPress index.
+		 *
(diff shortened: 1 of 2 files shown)


sc0ttkclark:
Looks good, my vote is STILL to include thumbnails by default, but you know me :)

ChrisWiegman:
I'm not commenting on thumbnails one way or another (personally I would like to include media information but that's a whole other matter). I just want to get potentially any field I need when I need it :smile:

sc0ttkclark:
I humbly request you add the ability in this filter handling to support `add_filter( 'ep_prepare_meta_allowed_protected_keys', '__return_true' );` so that all protected keys can be indexed, no matter the key name.

sc0ttkclark:
Oh hey, I can edit, so I added it :)

ChrisWiegman:
Good catch @sc0ttkclark Fixed.

tlovett1:
Can one of you add a unit test here?

ChrisWiegman:
I've added tests that should be sufficient for this PR and can be expanded to better cover #328 as we move forward.

allan23:
@tlovett1 This looks good.

sc0ttkclark:
Do we need to add this to readme / docs?

sc0ttkclark:
I still strongly believe that _thumbnail_id should be in the default list, but I know that Taylor is against it. Just wanted to note it :)

This will resolve the concerns I was up against in #328, although it still requires a custom filter to enable _thumbnail_id.

allan23:
Tests are confirmed working locally and in Travis; not sure why Dockunit doesn't match. @tlovett1

PR #352, opened 2015-09-16
Aborts sync when post status is auto-draft. Unit test included.

allan23:
When you hit post-new.php, WP will create a post with the post status of auto-draft. When the wp_insert_post action is fired, EP will try to delete the post from the index and will fail. This PR will prevent the HTTP request from occurring if the post status is auto-draft.


sc0ttkclark:
+1 if all tests pass :)

PR #354, opened 2015-09-16
Allow guid to be index and retrieved

sc0ttkclark:
As we now support attachment indexing, allowing guid to be indexed and
retrieved is crucial as this field is utilised in media screen to
display file name.

Making a PR for @tuanmh


tuanmh:
thanks guys!

PR #364, opened 2015-09-21
Add a filter to search scope before search is called.

ChrisWiegman:
Adds a filter to the search scope allowing for easier manipulation over complex sites.

### DIFF
--- a/classes/class-ep-wp-query-integration.php
+++ b/classes/class-ep-wp-query-integration.php
@@ -262,6 +262,17 @@ class EP_WP_Query_Integration {
 				$scope = $query_vars['sites'];
 			}
 
+			/**
+			 * Filter search scope
+			 *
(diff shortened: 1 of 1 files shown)


classes/class-ep-wp-query-integration.php
```diff
@@ -262,6 +262,17 @@ public function filter_posts_request( $request, $query ) {
...
+			/**
+			 * Filter search scope
+			 *
+			 * Allows for easier filtering of search scope to meet the requirements of more complex sites, etc.
+			 *
+			 * @since 1.7
+			 *
+			 * @param string $scope The search scope.
+			 */
+			$scope = apply_filters( 'ep_search_scope', $scope );
```
tlovett1: Not sure how useful this filter is without additional stuff being provided to it.

ChrisWiegman:
yep, I think you're right on that one. This was a premature answer to the problem I was facing at the time

Issue #338, opened 2015-08-28
Complex meta values are lost


ChrisWiegman:
This is partially related to #328 but from a different direction.

Currently prepare_meta() attempts to unserialize the meta values it encounters but there are a couple of issues here.

First, it should never hit a meta value it would need to unserialize as the get_meta call only returns arrays of the meta values (not the single value) which is how it should work and could easily be overcome with a check of each element of the $value array in the function.

However that poses a different issue. Namely if the value is a data structure other than a string it will fail due to the mappings allowing only a single level for any given meta item (an array or object would kill the indexing of the item with an error).

All that said, there are a couple of approaches that would help solve this. We can either drop nested meta data (which may be bad) or we can alter the mappings to allow for the nested data. In either case my biggest concern would be backwards compatibility as there are potentially a lot of indexes that are missing data due to this issue.

Thoughts?

tlovett1:
How does it attempt to unserialize? Doesn't `get_post_meta` do that? Can you create a supporting unit test that demonstrates the problem?

ChrisWiegman:
There are a few issues in how this works.

First, as a quick demonstration take a look at the [feature/meta-poc](https://github.com/10up/ElasticPress/tree/feature/meta-poc) branch which will provide some demonstration on the issue.

To go a little deeper the prepare_meta method in the api class fetches all meta items for a given post with the get_post_meta which it then, and this is the first bug, tries to run each value through the maybe_unserialize function. The issue here is that get_post_meta returns an array of all meta items associated with the post. Because of this it only attempts to unserialize the array it sees as the value rather than the serialized content that may be present in the meta entry. This could be fixed by running maybe_unserialize on $value[0] as opposed to $value (which is already an array) but leads to other issues. If we do properly unserialize the value the meta field elasticsearch will fail to index the post due to the data structure. 

This leads us to a few options:

# Adjust the mapping for a nested data structure and fix prepare_meta to unserialize the appropriate value
# Promote nested meta options to top-level meta items (a subkey would become a primary meta key). This would actually make a lot of sense for searching things like media but probably not much else

In either case, as the current situation is actually broken (existing nested items aren't being returned in search results) we must be careful to risk not breaking existing searches for sites that have grown accustomed to current behavior.

tlovett1:
[...]

PR #365, opened 2015-09-21
Meta Formatting Fixes

tlovett1:
This ticket is in response to #338.

Unfortunately, due to limitations of Elasticsearch we cannot store unserialized meta objects and arrays. We can however serialize them in case anyone needs that information. Right now our mapping is limited to storing boolean, integer, long, double, and string (possibly float not sure).

This PR fixes a line of useless code that was unserializing an array every time. It also does a check on each meta value. If it's an array or object, it serializes. If not, it simply indexes the value. A slew of tests have been added to test this code and the general meta mapping.

Down the road I can see us improving the mapping a bit, but this PR maintains backwards compatibility.

@ChrisWiegman @allan23 can you review?

### DIFF
--- a/README.md
+++ b/README.md
@@ -227,7 +227,7 @@ After running an index, ElasticPress integrates with `WP_Query` if and only if t
 
 * ```meta_query``` (*array*)
 
-    Filter posts by post meta conditions. Takes an array of form:
+    Filter posts by post meta conditions. Meta arrays and objects are serialized due to limitations of Elasticsearch. Takes an array of form:
 
     ```php
     new WP_Query( array(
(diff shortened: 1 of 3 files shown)


tlovett1:
@ChrisWiegman can you review?

ChrisWiegman:
Tried this with a podcast MP3 and there is still an issue. Namely it never unserializes $value_array to perform the check on line 622. That said, in this case it will actually work as it will simply save the serialized string however I don't think this is what you're looking for. In addition, when searching for text in the serialized string that is saved nothing is ever returned.

As a testing selection try this with a media file and watch the _wp_attachment_metadata field (you'll have to comment out lines 617 and 631 in the current API class to view).

tlovett1:
This is tabled until #376 gets merged.

PR #367, opened 2015-09-24
Added unicode char encoding to pass through json_encode()

dmitryshkolnikov:
If posts contain unicode characters the indexer fail to process it. I added encoder of unicode chars in posts before feeding them to json_encode()

### DIFF
--- a/bin/wp-cli.php
+++ b/bin/wp-cli.php
@@ -372,6 +372,8 @@ class ElasticPress_CLI_Command extends WP_CLI_Command {
 
 		// put the post into the queue
 		$this->posts[$post_id][] = '{ "index": { "_id": "' . absint( $post_id ) . '" } }';
+		// Encode unicode to utf8 for entire array
+		array_walk_recursive($post_args, function(&$item) {$item = utf8_encode($item);});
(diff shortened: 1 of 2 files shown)


tlovett1:
A few things:
1. This will probably need to happen in the `index_post` method of the API class
2. Please make sure you code follows the WordPress Coding Standards.
3. Are you trying to index characters in another language? Perhaps you should a different Elasticsearch analyzer.

dmitryshkolnikov:
1. I had that issue while initial indexing. But you're right, the same encoding should be in API class.
2. Fixed.
3. No, I'm trying to index the posts in English but they contain special characters like 1/2 or 3/4 as well as specific names in Spanish and French.

tlovett1:
@ChrisWiegman are you investigating this one?

tlovett1:
@ChrisWiegman ?

ChrisWiegman:
@dmitryshkolnikov Have you tried it with just wp_json_encode? While not while documented the idea is to verify all content is properly UTF-8 (see http://pento.net/2014/12/19/json-encoding-in-wordpress-4-1/)

dmitryshkolnikov:
Yes, with wp_json_encode we have no problem with our data.

PR #368, opened 2015-09-24
Add failure details.

ChrisWiegman:
When running a bulk index troubleshoot the individual errors in posts that fail to index can be problematic. This makes it a bit easier by adding a --show-bulk-errors param to display the error message along with the post that failed to index.


allan23:
@ChrisWiegman Readme will need an update too.

ChrisWiegman:
Good point Allan. Fixed

allan23:
@tlovett1 This looks good.

PR #372, opened 2015-10-03
Add support for orderby => meta_value and meta_value_num

### LINKED ISSUE #329: Adding orderby meta_value and meta_value_num support
In order to add support for a potential meta_value and meta_value_num value in 'orderby' arguments, we have to be able to sort by meta value. But post_meta data sets are all arrays when indexed, as prepared by EP_API::prepare_meta.

`$prepared_meta[$key] = maybe_unserialize( $value );`

Something like this could allow for single values, in which sorting would work for those cases:

``` php
if ( empty( $value ) ) {
    $value = '';
} elseif ( 1 == count( $value ) ) {
    $value = current( $value );
}

$prepared_meta[$key] = $value;
```

As noted above, I am also proposing we drop maybe_unserialize or make it optional. Right now it doesn't do anything because `$value` is always an array, so it never triggers is_serialized in maybe_unserialize anyway. Plus, in testing, maybe_unserialize can cause some trouble when indexing posts with a ton of extra serialized data, but that could just be an edge case.

If we keep maybe_unserialize, we should add it above the if statement proposed above, before the `$value` array gets tweaked by the above logic:

`$value = array_map( 'maybe_unserialize', $value );`

Once that prerequisite is done, an attempt to add a sort could be done. I'm not sure how to force ES to handle a value as an integer, but this would work for meta_value orderby in EP_API::parse_orderby:

``` php
            case 'meta_value':
                $meta_key = $args['meta_key'];

                $sort = array(
                    array(
                        'post_meta.' . $meta_key => array(
                            'order' => $order,
                        ),
                    ),
                );
                break;
```

The only issue with the above code would be that we'd have to add $args to parse_orderby as a parameter, and we'd have to ensure meta_key was set, otherwise use the default switch. We'd want to move the relevance / default case down to the bottom of the switch list and not `break;` in the meta_value case if meta key is empty.

Anyways, looking for direction here and if this all makes sense I'll put the PR together.

sc0ttkclark:
Fixes #329

Something to note, in order to sort by meta_value_num, a sort _script has to kick off. It seems to run okay in terms of speed, but we should note about it's performance in the readme, just as MySQL CAST().

- [x] Unit tests (meta_value / meta_value_num / asc+desc variations)
- [ ] Update `@since` if needed


sc0ttkclark:
I also moved relevance / default switch statement to the bottom, so it's easier to read for the future.

sc0ttkclark:
Added a tasklist to the PR for me to keep track of things that's needed to move PR forward.

sc0ttkclark:
Seems to be passing Travis-CI but failing on Dockunit, not sure yet what's up with that.

tests/test-single-site.php
```diff
@@ -1012,6 +1012,62 @@ public function testSearchPostTitleOrderbyQuery() {
 	}
 
 	/**
+	 * Test meta_value orderby query
+	 *
+	 * @since 1.1
```
sc0ttkclark: To be updated

tests/test-single-site.php
```diff
@@ -1012,6 +1012,62 @@ public function testSearchPostTitleOrderbyQuery() {
...
+		$this->assertEquals( 3, $query->found_posts );
+		$this->assertEquals( 'ordertest 333', $query->posts[0]->post_title );
+		$this->assertEquals( 'ordertest 222', $query->posts[1]->post_title );
+		$this->assertEquals( 'ordertest 111', $query->posts[2]->post_title );
+	}
+
+	/**
+	 * Test meta_value_num orderby query
+	 *
+	 * @since 1.1
```
sc0ttkclark: To be updated

sc0ttkclark:
@tlovett1 What's the pathway towards merge on this one?

sc0ttkclark:
As of Elasticsearch 1.3.8 and 1.4.3, Dynamic scripting for Groovy is disabled by default.

https://www.elastic.co/blog/running-groovy-scripts-without-dynamic-scripting

I see no way to let EP define a script to be used in a non-dynamic way, for the purposes of sorting by a meta field that's numeric, except for using Lucene Expression Scripts which only works with number field types:

https://www.elastic.co/guide/en/elasticsearch/reference/current/modules-scripting.html#_lucene_expressions_scripts

Have to put this feature on hold, it breaks EP if Dynamic Groovy Scripting is not enabled:

`ScriptException[dynamic scripting for [groovy] disabled]`

sc0ttkclark:
Moving forward without dynamic scripting, adding additional mappings to the dynamic templates in addition to .raw, I've included: .long, .double, .date, .datetime, .time (all typed and formatted to their corresponding ES core data types).

sc0ttkclark:
Still more work to do, after reviewing, we need to discuss further with ES contributors to ensure the solution I've come up with is still feasible and not going to cause backwards-compat issues.

tlovett1:
Seems like this will be solved with #376. Correct?

sc0ttkclark:
#376 is somewhat in this PR as a primary, but yes, they both have the same code now in place for the mapping. Just need to know what `@since` to put for this PR.

classes/class-ep-api.php
```diff
@@ -1535,6 +1541,18 @@ protected function parse_orderby( $orderby, $order ) {
 							'order' => $order,
 						),
 					);
+				} elseif ( 'meta_value' === $orderby_clause || 'meta_value_num' === $orderby_clause ) {
+					$meta_key = $args['meta_key'];
+					$orderby_path = 'raw';
+
+					if ( 'meta_value_num' === $orderby_clause ) {
+						$orderby_path = 'double';
```
sc0ttkclark: @tlovett1 I'm using `double` here but in your existing unit tests sorting by numeric you were using `long`, any preference?

sc0ttkclark:
@tlovett1 needs review and adjustment for `@since`

dkotter:
@sc0ttkclark I've reviewed the code here and it looks good but the ordering doesn't appear to actually work.

Digging into this deeper, I think the issue here is that in a normal `WP_Query`, sorting by `meta_value` or `meta_value_num` only works if you specify `'meta_value' => 'some value'`. The issue is that ElasticPress currently doesn't support that, it only supports full meta_queries.

Feel free to correct me if I'm wrong but if not, would love to see support added for `'meta_value' => 'some value'` queries, as part of this PR.

sc0ttkclark:
@dkotter are you using `'meta_key' => 'your_meta_key` in your WP_Query args so it knows what to sort by?

dkotter:
Yep. Here's the test code I was using:

```
new WP_Query( array(
	'ep_integrate' => true,
	'meta_key'     => 'ages',
	'order_by'     => 'meta_value_num',
) );
```

I have 2 posts with the ages meta_key set, but running the query above gives me 10 results. If I take off the `'ep_integrate' => true,` part, I get the proper 2 results.

sc0ttkclark:
Oh I see what you mean, yeah it's sorting on that meta_key but EP doesn't care that there's no value -- it's not restricting results to only those with that value.

Sounds like there's a little more work to be done here then. I'll look at that next.

sc0ttkclark:
Tinkering a bit right now to see if we can automatically restrict meta_query to check if that meta key exists, when using orderby and under the right conditions.

sc0ttkclark:
Looks like it's testing correctly now after some more tweaks to auto-handle meta_query exists.

sc0ttkclark:
@dkotter ready for your review again

dkotter:
@sc0ttkclark Took another look and the code looks fine and doing the same query above now works as expected, giving me the expected 2 results in the correct order, which is great.

But from what I'm seeing, we're still not fully supporting `'meta_key' => 'meta value',` queries, at least not without the `orderby` set.

For instance, if I run just this:

```
new WP_Query( array(
    'ep_integrate' => true,
    'meta_key'     => 'ages',
) );
```

instead of getting the expected 2 results, I get 10 results again. So removing the `orderby` param, that prevents the `meta_key` part from working. So would be nice to have those work with or without the `orderby` set.

sc0ttkclark:
Right, this PR doesn't add support for basic meta_key queries, it's explicitly for orderby. This would be done just before we start processing the meta_query in a separate PR I'd imagine.

dkotter:
Right, but unless I'm not following correctly, to have this ordering work, we need to have a `meta_key` set, so this PR does add some support there.

Thoughts on this @tlovett1? I know you initially said we should have the full support here to make the `orderby` work as expected. The current state of this PR makes the ordering work but doesn't make `'meta_key' => 'meta value'` queries work if you don't have that `orderby` set.

PR #373, opened 2015-10-03
Add support for orderby => post__in

### LINKED ISSUE #330: Adding orderby post__in support

``` php
        }
```

sc0ttkclark:
Fixes #330

- [ ] Unit test for `'orderby' => 'post__in'`
- [ ] Update `@since` if needed


tlovett1:
Will merge once the unit test is added. This just orders by the order in the post__in array?

sc0ttkclark:
That's correct, in a similar way as WP_Query does for it's orderby => post__in

sc0ttkclark:
Another tweak needed for this, will look into it

sc0ttkclark:
New PR posted here: #903

PR #374, opened 2015-10-05
Fetching the test suite from 4.3 branch instead of trunk.

allan23:
New functionality was recently introduced into WP trunk that will break all unit tests regardless of version. This PR updates the install script to fetch the testing suite from the 4.3 branch instead of trunk.

Definitely open to thoughts on this one.

screen shot 2015-10-05 at 10 36 57 am

screen shot 2015-10-05 at 10 37 16 am


danielbachhuber:
#375 is the more precise approach.

PR #376, opened 2015-10-13
Remap meta mappings for extensive `type` usage

sc0ttkclark:
Added new dynamic template mapping for long, double, date, datetime, time as well -- same as #372

- [x] Unit test for all `'type'` options for `'meta_query'`
- [x] Update `@since` if needed

### DIFF
--- a/classes/class-ep-api.php
+++ b/classes/class-ep-api.php
@@ -489,6 +489,9 @@ class EP_API {
 		 * This filter is named poorly but has to stay to keep backwards compat
 		 */
 		$post_args = apply_filters( 'ep_post_sync_args', $post_args, $post_id );
+
+		$post_args['meta'] = $this->prepare_meta_types( $post_args['post_meta'] );
+
 		return $post_args;
 	}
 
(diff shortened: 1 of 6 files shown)


tlovett1:
Is this ready to go?

sc0ttkclark:
Needs tests for each 'meta_query' 'type' option that WP_Query currently supports, to ensure they work as expected for ES (no unexpected exceptions, etc).

classes/class-ep-api.php
```diff
@@ -489,6 +489,9 @@ public function prepare_post( $post_id ) {
 		 * This filter is named poorly but has to stay to keep backwards compat
 		 */
 		$post_args = apply_filters( 'ep_post_sync_args', $post_args, $post_id );
```
tlovett1: Let's create another similarly named filter after `prepare_meta_types`

tlovett1:
This looks really solid. Had one comment.

tlovett1:
We will need lots of documentation updates for this.

ChrisWiegman:
I've added unit tests to test the prep function. I'm thinking of adding an "object" type here to cover data structures. What do you think @sc0ttkclark ?

tlovett1:
@ChrisWiegman if object actually works I think that would be cool. The sense that I got was that it isn't possible to index arbitrary objects.

sc0ttkclark:
From what I gathered, I didn't think objects could be used here because you had to define the specific fields on those objects.

sc0ttkclark:
I also don't think serialized data should be evaluated in this case, unless it was a basic object already (serialized arrays don't sound super easy to interact with in ES)

PR #377, opened 2015-10-13
More comprehensive tests for 'compare' on number meta values to ensure proper comparisons

sc0ttkclark:
Just to ensure range comparisons are working as intended for numeric values. May need follow-up related to #376 to use a numeric mapping instead of `.raw` if it doesn't pass with the added test data.


tlovett1:
Maybe we should merge these tests into #376?

tlovett1:
@sc0ttkclark I think this is irrelevant now. Reopen if I'm wrong.

PR #379, opened 2015-10-14
Fix/bulk index kill

ChrisWiegman:
Currently if bulk indexing and using the ep_post_sync_kill filter the bulk indexing operation will miss posts due to a difference in the number of posts it thinks it should have vs the number of posts present after the kill operation as it doesn't take into effect posts that are no longer there due to the kill filter. This takes care of that by tracking the number of posts killed and making this a bit easier to read in the logs as the post statuses display.

Note that previously EP counted all posts and told you they were all synced as queue_posts always returned true unless there was an actual error. This could be quite confusing as, for example, if you have 150 posts on a site and ep_post_sync_kill eliminates 5 of them (which happens after the initial query) The status will tell you that all posts are synced but because there are now less posts than the trigger accounts for (the post counter doesn't increment until after the kill already returns true) the bulk trigger will never be hit as it the post count will never reach the threshold.

Another option for this would have been to simply move up the post counter but that leads to inaccurate statuses as EP will list all the posts as synced without regard to whether that post data was sent. Now it will tell you posts processed (which should be all of them) and list posts "indexed" when the site completes for a better reflection on what has actually happened. It is actually this confusion in display that made this bug so hard to find in the first place.

### DIFF
--- a/bin/wp-cli.php
+++ b/bin/wp-cli.php
@@ -366,7 +366,7 @@ class ElasticPress_CLI_Command extends WP_CLI_Command {
 	 * @param $bulk_trigger
 	 * @param bool $show_bulk_errors true to show individual post error messages for bulk errors
 	 *
-	 * @return bool
+	 * @return bool|int true if successfully synced, false if not or 2 if post was killed before sync
 	 */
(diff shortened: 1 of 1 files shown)


tlovett1:
Looks good.

PR #385, opened 2015-10-27
Use wp_json functions for encoding and decoding

ChrisWiegman:
I do believe this will address #367. Need to test with some better data however.

### DIFF
--- a/bin/wp-cli.php
+++ b/bin/wp-cli.php
@@ -386,7 +386,16 @@ class ElasticPress_CLI_Command extends WP_CLI_Command {
 
 			// put the post into the queue
 			$this->posts[ $post_id ][] = '{ "index": { "_id": "' . absint( $post_id ) . '" } }';
-			$this->posts[ $post_id ][] = addcslashes( wp_json_encode( $post_args ), "\n" );
+
+			if ( function_exists( 'wp_json_encode' ) ) {
+
+				$this->posts[ $post_id ][] = addcslashes( wp_json_encode( $post_args ), "\n" );
+
+			} else {
+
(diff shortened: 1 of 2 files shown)


ChrisWiegman:
@tlovett1 Sounds like this will solve #367

allan23:
@ChrisWiegman This is failing WP 3.8 tests.

ChrisWiegman:
@allan23 How far back are we looking to go? Looking at the test details it those tests aren't even running. They're giving failures on other dependencies (or am I missing something)?

allan23:
@ChrisWiegman The problem is that wp_json_encode was introduced in 4.1. We still need backwards compatibility to 3.8. You could check if the function exists, if it doesn't you can fallback to json_encode. It won't resolve the problem for WP prior to 4.1, but it will make the plugin still backwards compatible.

@tlovett1 Thoughts?

ChrisWiegman:
Thanks @allan23! I wasn't sure if 3.8 was actually something we were supporting or, as has been my experience, with most plugins, was simply the first version it was built against and would be supported until something broke it. 

That said, yeah, a conditional will fix it. But then we're not supporting other char sets. Do we want to do that or bring the min version up to 4.1?

tlovett1:
Think we need to maintain backwards compat here.

ChrisWiegman:
Will do.

ChrisWiegman:
@tlovett1 I did wrap the functions for back compat on this. Should be ready to merge.

tlovett1:
Probably still need a way to address the UTF8 problem on older versions of WP but for now I think this is fine.

PR #386, opened 2015-10-30
Still sync on publish/update when doing cron

### LINKED ISSUE #322: Scheduled posts are not indexed when transitioning to publish status
Thanks 10up for this great plugin!

I noticed that my posts scheduled for publishing are not indexed at all.

Looking at the EP_Sync_Manager class, post sync seems only to happen on `wp_insert_post` hook, which is probably too simplistic. For instance my posts with scheduled publishing are not indexed because WP don't run this hook on cron.

Also what happens if a post meta or a term is updated? It looks like these change are not sync to Elastic.

sc0ttkclark:
Fixes #322


tlovett1:
Thanks Scott!

PR #389, opened 2015-11-09
(Version 1.6.1) Make mapping compatible with Elasticsearch version 2.0.0

tlovett1:
This fix does two things:

* Removes the fields field type from object typed fields as they should be called properties.
* Remove path from object field types.

Dockunit is upgrading to ES 2.0.0 today. Not sure where Travis CI is at, but at least we can see it's backwards compat.

### DIFF
--- a/elasticpress.php
+++ b/elasticpress.php
@@ -3,7 +3,7 @@
 /**
  * Plugin Name: ElasticPress
  * Description: Integrate WordPress search with Elasticsearch
- * Version:     1.6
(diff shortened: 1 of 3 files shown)


ChrisWiegman:
Tested on EP 2.0.0 with Primary Vagrant and EP 1.7.2 with VVV. Indexing complete on both including mappings. Was also able to delete and reindex. Looks like this should be good to go.

PR #396, opened 2015-12-02
Feature/gui indexing

ChrisWiegman:
While still a work in progress this is close.

What has been completed:

Indexing works great on single site and on a single site in a multisite network. For a full network index there are still a few bugs. Mapping works great but indexing doesn't work well yet in all situations.


ChrisWiegman:
While tests are still failing the GUI indexing is working like it should and I've worked out the remainder of the bugs that were present from it's previous incarnation. Should be ready for discussion.

allan23:
@ChrisWiegman can you rebase?

ChrisWiegman:
@allan23 Rebase complete. If we elect to hold off on this again I'll save further rebases until we're ready to merge

assets/js/elasticpress-admin.js
```diff
@@ -0,0 +1,195 @@
...
+		bar.show();
+
+		if ( resetBar ) {
+
+			var progress = 0;
+
+			if ( parseInt( ep.total_posts ) > 0 ) {
+
+				progress = parseFloat( ep.synced_posts ) / parseFloat( ep.total_posts );
+				status.html( ep.synced_posts + '/' + ep.total_posts + 'items' );
```
dkotter: Should probably use ```.text``` here instead of ```.html```, to strip out any potential tags.

assets/js/elasticpress-admin.js
```diff
@@ -0,0 +1,195 @@
...
+
+					var progress = parseFloat( epTotalIndexed ) / parseFloat( epTotalToIndex );
+
+					bar.progressbar(
+						{
+							value : progress * 100
+						}
+					);
+
+					status.html( epTotalIndexed + '/' + epTotalToIndex + ' ' + ep.items_indexed + sitesCompletedText );
```
dkotter: Same as above, should use ```.text``` here if possible.

assets/js/elasticpress-admin.js
```diff
@@ -0,0 +1,195 @@
...
+
+		//call the ajax
+		$.ajax(
+			{
+				url :      ajaxurl,
+				type :     'POST',
+				data :     data,
+				complete : function ( response ) {
+
+					$( '#ep_site_stats' ).html( response.responseJSON.data );
```
dkotter: As mentioned in a couple comments above, nice to avoid using ```.html``` when possible (see https://vip.wordpress.com/2015/03/25/preventing-xss-in-javascript/). In this case though, we'd need to change the response from the ep_get_site_stats endpoint to just send the data, not markup and then build up the markup here before outputting.

I think this would be a nice to have but not necessarily an absolute requirement before merge.

assets/js/elasticpress-admin.js
```diff
@@ -0,0 +1,195 @@
...
+	var selector = $( '#ep_site_select' );
+
+	/**
+	 * Process changing site stats.
+	 */
+	selector.change( function ( event ) {
+
+		event.preventDefault();
+
+		console.log( selector.val() );
```
dkotter: We'll probably want to remove this.

classes/class-ep-index-gui.php
```diff
@@ -0,0 +1,346 @@
+<?php
+/**
+ * Jovo Indexing Interface
```
dkotter: Assuming we'll want to replace any references to Jovo to ElasticPress

classes/class-ep-index-gui.php
```diff
@@ -0,0 +1,346 @@
...
+			),
+			$ep_settings->options_page,
+			'normal',
+			'core',
+			array( 'index.php' )
+		);
+
+	}
+
+	protected function _run_index() {
```
dkotter: Would be nice to get a docblock here to explain what this function does.

classes/class-ep-index-worker.php
```diff
@@ -0,0 +1,380 @@
+<?php
+/**
+ * Process indexing of posts
+ *
+ * @package Jovosearch
```
dkotter: Another reference to Jovosearch that we'll want to update

classes/class-ep-index-worker.php
```diff
@@ -0,0 +1,380 @@
...
+	/**
+	 * Send any bulk indexing errors
+	 *
+	 * Emails bulk errors regarding any posts that failed to index.
+	 *
+	 * @since 0.1.0
+	 *
+	 * @return void
+	 */
+	protected function send_bulk_errors() {
```
dkotter: Maybe I just missed it but not seeing any option to turn off this feature. I could see a situation where someone might not want emails getting sent out, so would be nice to potentially have a filter around this to disable this functionality.

classes/class-ep-index-worker.php
```diff
@@ -0,0 +1,380 @@
...
+			foreach ( $failed_posts as $failed ) {
+
+				$failed_post = get_post( $failed );
+
+				if ( $failed_post ) {
+					$email_text .= "- {$failed}: " . $failed_post->post_title . PHP_EOL;
+				}
+			}
+
+			wp_mail( get_option( 'admin_email' ), wp_specialchars_decode( get_option( 'blogname' ) ) . esc_html__( ': ElasticPress Index Errors', 'elasticpress' ), $email_text );
```
dkotter: Would be nice to have a filter around the email so that can easily be changed to something else if needed.

includes/settings/status.php
```diff
@@ -0,0 +1,187 @@
+<?php
+/**
+ * Template for displaying Elasticsearch statistics
+ *
+ * @since   0.1.0
+ *
+ * @package Jovosearch
```
dkotter: Another reference to Jovosearch we'll want to update.

dkotter:
@ChrisWiegman This looks really good! Left some comments on the PR for some minor fixes. Also, we definitely want to try and get this in version 1.9, so all @since blocks should be updated to reflect that (saw quite a few of these that will need updating).

Also, having issues running the unit tests locally, keeps crashing my install of elasticsearch, but according to the above checks, some unit tests are failing, so we'll need to get those fixed up.

And finally, since we do want to put this in 1.9 if at all possible, at some point would be nice to get all recent changes into this PR (rebase/merge). 

Let me know if you have questions/concerns on any of these items. Once the rebase has been done, I'm going to be running more tests on the indexing feature, making sure it works fine on multisite and sites with lots of content, but so far things have been working great.

dkotter:
@ChrisWiegman This looks really good and now with all unit tests passing, I think it's mostly just refinements at this point. To that point, I've been testing this in multiple situations and have a few minor things I've found so far I think would be nice to change:

* On a site where Elasticsearch isn't started/setup but the `EP_HOST` constant is set, on the settings page, the ElasticSearch host field is filled in and disabled (which is fine) but in the current status box, I get the message: `A host has not been set. You must set a host to continue.` Not a major thing but the host is actually set, just not available. And right after this message I get the correct error. So I think either remove the line about the host not being set if a host really is set but not available (since the exact error is mentioned after) or modify the error message to say the host is set but not available.

ElasticPress
* Similarly, on a site where Elasticsearch isn't started/setup and no `EP_HOST` constant is found, I get the `ERROR: ElasticSearch Host is not available` message. Again, not a huge deal but if no host has been added, there's no host to check availability for, so I think the error message shouldn't show until a host has been entered.

ElasticPress
* In both of these situations, I still get the option to index the site, which causes JS errors. So until a host is set and we've verified the host works, I think the index option should be hidden.

I'm going to continue testing and will let you know if I notice anything else. Thanks for all your hard work here.

dkotter:
If I have a correct host set and Elasticsearch is setup properly, I then get the option to Use ElasticSearch. If I don't check this box, I get the message: `ElasticPress is not activated and cannot override WP search. You can activate it on the form to the left `. Again, nothing majorly wrong with that but it is a little confusing because ElasticPress the plugin is active but it's not being used. So not sure the best wording to use here but could be as simple as swapping activated with enabled: `ElasticPress is not enabled and cannot override WP search. You can activate it on the form to the left`

ChrisWiegman:
Hi @dkotter 

I've been unable to reproduce the "A host has not been set. You must set a host to continue" message on a new site. Setting up in Single or multisite gives me the following:

screen shot 2016-01-27 at 09 24 50

I could see how this might happen after host has switched from GUI to constant due to transients. Perhaps that is what you're seeing there?

ChrisWiegman:
I've updated the copy on the remaining messages. I do have a bit of concern on the "activated" message though as this is how we refer to it in CLI. For now I've changed it to say "enabled" but this isn't consistent to what we're using elsewhere. That said, I'm thinking a better way would be to update CLI to "enabled" but that would touch a number of functions. Thoughts?

dkotter:
@ChrisWiegman Newest changes look good. And yeah, I agree on wanting to be consistent. I know this was something mentioned by Taylor in another issue: #64. That said, not sure if we want to make changes to other places not related to this PR, even if it is just text changes. Not a huge deal but maybe those should be made separately from this PR, after it's merged in? If so, can easily create another issue we can track those changes against.

For the first message, the issue I had was when I had the EP_HOST constant set but elasticsearch itself wasn't activated (just had done a new `vagrant up`). So the host was set properly but the elasticsearch instance couldn't be reached, resulting in a wrong error message. Again, not a huge deal and probably an edge case, just something I noticed in testing.

dkotter:
I've done lots of testing on single sites and haven't run across any issues yet. Do have some questions on multisite though. It appears this has been set up to only run properly if the plugin is network activated, is that correct? I initially activated the plugin on a single site on the network but none of the settings would save. After some debugging, found out that it would only work if the plugin was network activated, and then at that point, you have to go to the network admin page to get to the settings.

If we want to force this to only be accessed from the network settings, I think that's fine but we should make sure it doesn't ever show up on a single site in the network, as again, I'm not seeing it work if that's the case. I do think there's probably use cases where someone would only want certain sites in the network to use Elasticsearch, so that's something to consider. As-is, appears you have to use ElasticPress/Elasticsearch on all sites or none. This might not be something we add for this release but definitely a feature that would be nice to have soon.

When I finally realized the plugin had to be network activated, I did that and then I tried syncing a large multisite (not sure if the size matters here or not) and the progress bar had some issues. Showed having 0 items and 0 sites, even though there were definitely some of both.



If I refreshed the page, I got the correct number of items (no sites listed though) but needs a space between the amount and the word items:

ChrisWiegman:
Thanks @dkotter For the first, I'll keep trying. I had actually added the constant to a brand new site but didn't get the issue....

...For the network stuff, that's a bug. I've been jumping back and forth between network activation and single site so much that I think that case got mixed. I'll get that fixed up next

ChrisWiegman:
@dkotter This seems to be working for me now on multisite if the plugin is only activated on one site or on more than one. Do you think it would be advisable for a feature/filter to exclude individual sites on a network activation? This wouldn't be too difficult.

dkotter:
@ChrisWiegman Awesome! Just tested and can confirm you can now activate this on individual sites on the network and index from there or you can network activate it and run from the network settings page.

I think it would be nice to choose what sites get indexed if you network activate it. If that's something that can easily be done for this release, great. If not, definitely something to add in the future. I can see this either being a filter where someone can modify what sites get either included or excluded or some sort of UI around this where you choose what sites to include/exclude. I think the filter is probably easiest to implement, especially for networks that might have lots of sites, but I can see it being nice to be able to choose straight from the settings screen.

ChrisWiegman:
- [ ] Remove EP API Key

- [ ] Add notice if AJAX request fails (currently just freezes)

dkotter:
I think this looks really good. Definitely some things we can probably add/clean-up, but I think those are things that should be in separate PR's at this point. So merging this in now for a 1.9 release.

Issue #397, opened 2015-12-02
Fuzziness on by default ? How do I get exact results from ElasticSearch ?


paulgv:
Hi,
I would like to know how I should build WP_queries in order to get exact matches from ES.
Performing a basic search query like this :
```
$my_query = new WP_Query( array(
    's'             => 'query',
    'search_fields' => array(
        'post_title',
    )
) );
```
Returns too much results for my need. This is caused by the `fuzziness` parameter in `class-ep-api.php` :
```
$query = array(
    'bool' => array(
        'should' => array(
            array(
                'multi_match' => array(
                    'query' => '',
                    'fields' => $search_fields,
                    'boost' => apply_filters( 'ep_match_boost', 2 ),
                    'fuzziness' => 0,
                )
            ),
            array(
                'multi_match' => array(
                    'fields' => $search_fields,
                    'query' => '',
                    'fuzziness' => 2,
                    'operator' => 'or',
                ),
            )
        ),
    ),
);
```

Why is this parameter set to 2 ? How do I set it to 0 for specific queries ?
Thanks !

PR #398, opened 2015-12-04
Allow setting fuzziness in search query arguments

paulgv:
Hi,
This is related to issue #397.
I didn't find a way to change the fuzziness level when performing a search query so here is an attempt to make this option configurable with a `fuzziness` argument that you can pass in your `WP_Query` :
```
$query = new WP_Query( array(
    's'             => 'term',
    'fuzziness'		=> 0
) );
```

### DIFF
--- a/classes/class-ep-api.php
+++ b/classes/class-ep-api.php
@@ -1277,6 +1277,8 @@ class EP_API {
 
 		$search_fields = apply_filters( 'ep_search_fields', $search_fields, $args );
 
+        $fuzziness = (!isset($args['fuzziness'])) ? 2 : $args['fuzziness'];
+
 		$query = array(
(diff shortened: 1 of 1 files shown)


classes/class-ep-api.php
```diff
@@ -1292,7 +1294,7 @@ public function format_args( $args ) {
 						'multi_match' => array(
 							'fields' => $search_fields,
 							'query' => '',
-							'fuzziness' => 2,
+							'fuzziness' => $fuzziness,
```
aheinz-sg: For fuzziness = 0, this second multi_match becomes redundant. Otherwise I like this patch, as I have the same requirement for matches without any fuzziness.

tlovett1:
See #408

PR #400, opened 2015-12-06
(Closes #390) Fix --no-bulk option

### LINKED ISSUE #390: Cannot use --no-bulk indexing parameter.
When I use `--no-bulk` index parameter I get following error.

```
Error: Parameter errors:
 unknown --bulk parameter
```

lukaspawlik:


allan23:
@lukaspawlik Can you please update the readme since this changes the CLI argument name?

lukaspawlik:
@allan23 sure as I am not native speaker my "update" may not be perfect. Feel free to suggest something more appropriate :smile:

tlovett1:
Does this break backwards compat? What do the logical changes do?

allan23:
@lukaspawlik After running a few tests, it appear that the default index command does so in bulk and by adding the ``--bulk`` argument it still indexes in bulk.

Wouldn't it be easier to change the ``--no-bulk`` argument to ``--nobulk`` by updating lines 184 and 299?
That way functionality would stay the same. 

Just curious on your thoughts.

lukaspawlik:
@allan23 yes it would be easier to rename --no-bulk to --nobulk. I decided to prepare patch which is backward compatible. If everyone is okay with that I can modify this PR.

lukaspawlik:
@allan23 I've renamed to --nobulk as you suggested

dkotter:
Most recent changes look good to me and I've tested this out and it seems to work fine. Wasn't sure why the argument wasn't being recognized so dug into the actual wp-cli code and looks like any params that start with `--no-` get that part stripped. Couldn't find that documented anywhere and couldn't find any reasoning, so not sure why.

But this should fix the issue. @lukaspawlik There are some merge conflicts. Can you resolve those so I can get this merged in? Thanks!

lukaspawlik:
@dkotter rebased

dkotter:
Looks good, merging this in for a 1.9 release.

PR #401, opened 2015-12-06
Add option not to disable ElasticPress while indexing

lukaspawlik:
Sometimes it would be very useful not to disable ElasticPress search integration during wp-cli indexing operation. The reason behind this feature is that sometimes there is a need to reindex site and keep using ElasticPress turned on as reindexing may take very long time (1-2 hours).

### DIFF
--- a/README.md
+++ b/README.md
@@ -489,7 +489,7 @@ The following are special parameters that are only supported by ElasticPress.
 
 The following commands are supported by ElasticPress:
 
-* `wp elasticpress index [--setup] [--network-wide] [--posts-per-page] [--no-bulk] [--offset] [--show-bulk-errors] [--post-type]`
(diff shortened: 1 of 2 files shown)


tlovett1:
@allan23 can you review?

tlovett1:
I think this makes sense but I don't understand the point of the `$keep_active` variable. Why not just call deactivate in the conditional if there is no `--keep-active`.

lukaspawlik:
@tlovett1 I've refactored this

sc0ttkclark:
This would be nice :)

dkotter:
This looks good to me and I've tested this out without any issues. This change is only for the cli command, but now that we have the admin GUI in place, would be nice to replicate this feature there. That's separate enough from this PR though that I'm fine with merging this in and handling the other separately.

fnpen:
Is it possible to keep working while indexing with the last version?

PR #406, opened 2015-12-15
Created filter 'ep_wp_query_vars' to add meta_query to ElasticSearch

imcodingideas:
Added a filter to add meta_query to ElasticSearch.


codingideas:
@tlovett1 this is a error in the time it took to build. Right?

dkotter:
@codingideas Can you maybe add some clarification on what you're trying to achieve with this filter? We already support doing a normal `meta_query`, so I guess just trying to figure out the use case here. Thanks!

allan23:
@codingideas Can you please advise?

PR #407, opened 2015-12-16
Extending PR #398 with default fuzziness option.

aheinz-sg:
Extending PR #398 with default fuzziness option.


tlovett1:
Thoughts on 408? Same intention.

aheinz-sg:
Closing in favor of #408.

PR #408, opened 2015-12-16
Add a filter around the second fuzziness argument

dkotter:
In version 1.6.2, the ```fuzzy_like_this``` param was removed, as it's deprecated in elasticsearch. The ```min_similarity``` field was also removed and replaced with the ```fuzziness``` field. This field had a filter around it so you could modify these results a bit. This adds a new filter around the new ```fuzziness``` field to give you the same freedom.

### DIFF
--- a/classes/class-ep-api.php
+++ b/classes/class-ep-api.php
@@ -1292,7 +1292,7 @@ class EP_API {
 						'multi_match' => array(
 							'fields' => $search_fields,
 							'query' => '',
-							'fuzziness' => 2,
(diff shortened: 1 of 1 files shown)


tlovett1:
We should be passing search_fields and args to the filter (for that matter the previous filter as well). Can you make that change?

dkotter:
Made that change.

aheinz-sg:
I do like the flexibility of this approach versus single value get_option. Unclear to me why the last Travis build failed for latest Wordpress.

PR #412, opened 2015-12-23
Sync searchable_post_types with indexable_post_types.

rossluebe:
Switching this out allows the `any` query to match the indexed post types. Because the [function](https://github.com/rossluebe/ElasticPress/blob/develop/classes/class-ep-config.php#L115-L119) already returns `get_post_types( array( 'exclude_from_search' => false ) );`, this will only affect users who have made changes to the indexable post types through the `ep_indexable_post_types` filter.

### DIFF
--- a/classes/class-ep-config.php
+++ b/classes/class-ep-config.php
@@ -285,6 +285,10 @@ function ep_get_indexable_post_types() {
 	return EP_Config::factory()->get_indexable_post_types();
 }
 
+function ep_get_searchable_post_types() {
+	return EP_Config::factory()->get_searchable_post_types();
+}
+
 function ep_get_indexable_post_status() {
 	return EP_Config::factory()->get_indexable_post_status();
 }
--- a/classes/class-ep-config.php
+++ b/classes/class-ep-config.php
@@ -128,6 +128,18 @@ class EP_Config {
 	}
 
 	/**
+	 * Returns searchable post types for the current site
+	 *
+	 * @since 1.9
+	 * @return mixed|void
+	 */
+	public function get_searchable_post_types() {
+		$post_types = get_post_types( array( 'exclude_from_search' => false ) );
+
+		return apply_filters( 'ep_searchable_post_types', $post_types );
+	}
+
+	/**
 	 * Return indexable post_status for the current site
 	 *
 	 * @since 1.3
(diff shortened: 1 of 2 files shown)


tlovett1:
@allan23 thoughts here?

rossluebe:
@tlovett1 @allan23 This PR seems to be hanging in limbo, so I wanted to provide a bit context on how this PR is useful.

Currently my search results return incorrect post types because the [`filter_post_request` function](https://github.com/10up/ElasticPress/blob/develop/classes/class-ep-wp-query-integration.php#L214) returns [`$searchable_post_types` as `get_post_types( array( 'exclude_from_search' => false ) );`](https://github.com/10up/ElasticPress/blob/develop/classes/class-ep-wp-query-integration.php#L229). The issue with this is there are post types that are not included by this core function that have been added in the [`ep_indexable_post_types` filter](https://github.com/10up/ElasticPress/blob/develop/classes/class-ep-config.php#L115-L119). It's possible that users of EP might not want to search against the indexable post types, in which case it would be useful to introduce a new filter at `$searchable_post_types` so the user can decide if they want to sync indexable and searchable. I'd be happy an amend my pull request if this is preferred instead of using the same filter for both.

dkotter:
@rossluebe I understand the issue here and the solution in this PR is good, but I am worried about it having unintended consequences on sites where they've modified indexable post types using that [ep_indexable_post_types](https://github.com/10up/ElasticPress/blob/develop/classes/class-ep-config.php#L127) filter.

So I think your most recent suggestion sounds good, adding a filter around the [$searchable_post_types](https://github.com/10up/ElasticPress/blob/develop/classes/class-ep-wp-query-integration.php#L229), allowing users to change that if needed.

Let me know if that makes sense and if so, would love to see an updated PR here.

Thanks!

rossluebe:
@dkotter Thanks for confirming my suspicions, I've updated the PR and rebased to the current state of the `develop` branch. One thing to note is I'm not sure which version of ElasticPress will get this PR, but I went ahead put the version number as `1.9` in the helper function. Please amend as needed.

dkotter:
@rossluebe This looks pretty good but the actual `ep_get_searchable_post_types` isn't defined anywhere. Just needs to be setup the same as `ep_get_indexable_post_types`, something like:

```
function ep_get_searchable_post_types() {
	return EP_Config::factory()->get_searchable_post_types();
}
```

in `classes/class-ep-config.php`

rossluebe:
@dkotter Sorry for missing that. I forgot these weren't namespaced files.

dkotter:
Thanks! This looks good, merging in for a 1.9 release.

PR #416, opened 2016-01-05
Only download MySQLi dropin when it's needed

johnpbloch:
WordPress supports mysqli by default as of 3.9, so we only need this dropin for 3.8 (at least, that's the lowest version we test for in our travis matrix). We also need to make sure we're grabbing the legacy version (the actual dropin) since the one downloaded only defines a constant telling WordPress not to use the mysql_* functions.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,8 +8,8 @@ php:
 env:
     - WP_VERSION=latest WP_MULTISITE=0
     - WP_VERSION=latest WP_MULTISITE=1
-    - WP_VERSION=3.8 WP_MULTISITE=0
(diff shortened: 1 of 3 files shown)


johnpbloch:
ಠ_ಠ

It's failing anyway because of a deprecated notice about PHP4-style constructors.

tlovett1:
What about WP 3.7?

johnpbloch:
Ah. So then shouldn't we be testing 3.7.1 instead of 3.8 in Travis, since that's our minimum supported version of WordPress?

tlovett1:
Yea, we should. Add to PR?

PR #418, opened 2016-01-12
Log queries for debugging

tlovett1:
This let's us do the following:
* Track the time taken for each query
* Log broken hosts
* Track the amount of EP queries per page
* Track actual arguments being sent to Elasticsearch

### DIFF
--- a/classes/class-ep-api.php
+++ b/classes/class-ep-api.php
@@ -1662,7 +1662,7 @@ class EP_API {
 				$query['url'] = $request_url;
 			}
 
-			$request = wp_remote_request( esc_url( $request_url, $args ); //try the existing host to avoid unnecessary calls
+			$request = wp_remote_request( $request_url, $args ); //try the existing host to avoid unnecessary calls
 		} else {
 			if ( defined( 'WP_DEBUG' ) && WP_DEBUG ) {
 				$query['failed_hosts'][] = $host;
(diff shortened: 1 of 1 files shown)


classes/class-ep-api.php
```diff
@@ -1622,6 +1629,15 @@ public function elasticsearch_alive( $host = null ) {
 	 */
 	public function remote_request( $path, $args = array() ) {
 
+		$query = array(
+			'time_start' => microtime( true ),
+			'args' => $args,
+			'blocking' => true,
+			'failed_hosts' => array(),
+			'request' => false,
+			'host' => false,
```
ChrisWiegman: I feel like time_finished should be initialized here and maybe set to false indicating nothing has been run. This will prevent index not found errors should anything be hit at weird times. Thoughts?

tlovett1: Totally agree.

ChrisWiegman:
I like it but I feel like the usability is a bit bulky. Would it make sense to include a "get_failed_queries" or similar method? For that matter, if logging in general, a few methods including get_failed_queries, get_all_queries, get_slowest_queries, etc seem like they might be helpful but at that point we can go on forever with options. Perhaps a simple get_queries with the right parameters would be the way to go. Thoughts?

tlovett1:
@ChrisWiegman let me know what you think of that

ChrisWiegman:
Perfect. That seems much more consistent with the way we've done everything else.

PR #419, opened 2016-01-12
Fix for wp_cache_flush

jstensved:
Fixes error "Fatal error: Cannot access private property WP_Object_Cache::$cache". Internally this methods does the same thing, see https://developer.wordpress.org/reference/classes/wp_object_cache/flush/


tlovett1:
@allan23 can you review?

PR #431, opened 2016-01-26
Ensures both PHPUnit and WP-CLI are available allowing a simplificati…

ChrisWiegman:
Ensures both PHPUnit and WP-CLI are available allowing a simplification of the development environment.


dkotter:
@ChrisWiegman Checked this out to test it and was getting an error when composer was trying to install wp-cli. I fixed it by changing the version from master to dev-master, which I think is correct but if not, let me know. And then I fixed the merge conflict.

If it sounds correct to be dev-master, I'll get this merged in. Thanks!

ChrisWiegman:
That sounds right. I hadn't received the error but that makes perfect sense.

PR #435, opened 2016-01-27
Use lower-case for composer name

johnpbloch:
Packagist requires lower-case names, so this needs to happen before we can submit to packagist. (I think this branch needs to be merged into master, but it probably also needs to be merged into develop, so I left the PR against develop)


dkotter:
This looks good, thanks! Fixed a merge conflict but will merge this in now for a 1.9 release.

Issue #420, opened 2016-01-12
Checking WP_Query->query instead of WP_Query->query_vars


tlovett1:
In `ep_elasticpress_enabled`, we are checking `WP_Query->query` instead of `WP_Query->query_vars`. I feel like we should be checking `WP_Query->query_vars`. Thoughts? @ChrisWiegman @allan23

ChrisWiegman:
After poking around a bit I don't see any reason why this was implemented this way. I've opened PR #436 to address. In testing the switch seems to work just fine.

dkotter:
Tested out the PR and worked great. This has been merged so I'm closing this out now. Thanks!

PR #436, opened 2016-01-28
Look in query_vars instead of query to determine enabled status.

ChrisWiegman:
Addresses #420


dkotter:
Code looks good and I've tested this out without seeing any issues. Merging this in for a 1.9 release.

PR #440, opened 2016-02-02
Support wordpress' default taxonomy query behavior

tuanmh:
We're are now having support for ``category_name`` in WP query parameter (https://github.com/10up/ElasticPress/blob/develop/classes/class-ep-api.php#L955), however we're still missing other things like `tag` or custom taxonomies. For example if I have a custom taxonomy called ``license``, searching for things under that taxonomy using the url pattern: `/license/*` will implicitly set `license` parameter in WP query parameters. The same behaviour is applied for `tag` with url pattern: `tag/*`

This PR is to support all the taxonomy query implicitly set in WP query including [built-in taxonomies](https://codex.wordpress.org/Class_Reference/WP_Query#Taxonomy_Parameters) and custom taxonomies.


tuanmh:
any feedback on this guys?

tlovett1:
@ivanlopez @ChrisWiegman needs review

ChrisWiegman:
This looks great. Simple "tag" and other queries now work and it isn't adding any extra queries to do it. @tuanmh can you please check the conflicts? @tlovett1 when conflicts are resolved this is good to merge.

tuanmh:
> @tuanmh can you please check the conflicts?

Okay done.

classes/class-ep-api.php
```diff
@@ -969,25 +988,6 @@ public function format_args( $args ) {
 		}
 
 		/**
-		 * 'category_name' arg support. Skip entirely if `tax_query` has already been provided. WP_Query
-		 * has some quirky backwards compat functionality that will fill `category_name` if the category
-		 * taxonomy is provided within `tax_query` which leads to duplicate filters.
-		 *
-		 * @since 1.5
-		 */
-		if ( ! empty( $args[ 'category_name' ] ) && empty( $tax_filter ) ) {
```
tlovett1: Why are we getting rid of this?

tuanmh: hey @tlovett1, this is covered in this section (see above code as well).
```php
//set tax_query if it's implicitly set in the query
		//e.g. $args['tag'], $args['category_name']
		if ( empty( $args['tax_query'] ) ) {
			$taxonomies = get_taxonomies();
			$taxonomies = $this->sanitize_taxonomy_names($taxonomies); //fix it up
			foreach( $taxonomies as $tax => $taxName ){
				if( isset( $args[ $taxName ] ) && ! empty( $args[ $taxName ] ) ){
					$args['tax_query'][] = array(
						'taxonomy' => $tax,
						'terms' =>  array($args[ $taxName ]),
						'field' => 'slug'
					);
				}
			}
		}
```

tlovett1:
This looks awesome. Thanks!

PR #443, opened 2016-02-09
Add the ability to pause indexing from the admin GUI

dkotter:
Adds the ability to pause indexing from the admin GUI introduced in #396. When the pause button is clicked, indexing will pause once it's done with the current indexing batch. At this point, you should be able to resume indexing at any time, even if the page is refreshed.


dkotter:
Added a button to also restart indexing, that shows once indexing has been paused. That way if someone wants to start indexing over but doesn't want to wait until the current index is done, they can pause, then click restart, which at that point will allow them to start the index over again.

PR #445, opened 2016-02-10
Reduce memory consumption during indexing

lukaspawlik:
Those changes reduces memory consuption during indexing. wp-cli memory usage was growing until process was killed by Out Of Memory exception. I've added query logging only when WP_DEBUG is enabled. RIght now wp-cli doesn't use more than 50M even when indexing large datasets.


dkotter:
Code looks good and I've tested this out and haven't seen any issues.

@tlovett1 Do you have any thoughts here? My only concern was with the idea of only logging queries if `WP_DEBUG` is on. I'm okay with that but wasn't sure if that's something we would want to always leave on for some reason.

tlovett1:
Looks good to me.

PR #450, opened 2016-02-16
Add configuration to choose to index all attachments at index time

sambonner:
This is an update to my PR from #339 which I'm afraid I let languish last year (apologies for this). I've modified the constant conditional to check that it is set to true as per @tlovett1's suggestion. With regards to point 2 on #339, I checked and it seems like the attachment post type is already available so I'm not sure this needs to be added. Happy to be corrected if I'm misunderstanding anything.

Thanks,
Sam

### DIFF
--- a/classes/class-ep-config.php
+++ b/classes/class-ep-config.php
@@ -125,7 +125,7 @@ class EP_Config {
 	 * @return array
 	 */
 	public function get_indexable_post_status() {
-    if ( defined( 'EP_INDEX_ATTACHMENTS' ) ) {
+    if ( defined( 'EP_INDEX_ATTACHMENTS' ) && EP_INDEX_ATTACHMENTS == true ) {
 		  return apply_filters( 'ep_indexable_post_status', array( 'publish', 'inherit' ) );
     }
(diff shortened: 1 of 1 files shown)


dkotter:
Hi @sambonner, thanks for the PR.

I think this looks pretty good and looking into `ep_indexable_post_types`, it will index any post types where `'exclude_from_search' => false`. By default, attachments are set up this way, so I don't believe anything special needs to be done there.

I do think the code could be cleaned up a bit. Nothing major, just not sure we want to have the `ep_indexable_post_status` filter there twice. I think just setting a variable that holds our post stati that we can add to if the `EP_INDEX_ATTACHMENTS` constant is set is probably the way to go. Also, to keep things consistent across the project, we don't need to check if the constant is equal to true, just something like this:

```
$post_stati = array( 'publish' );

if ( defined( 'EP_INDEX_ATTACHMENTS' ) && EP_INDEX_ATTACHMENTS ) {
	$post_stati[] = 'inherit';
}
```

Let me know if that makes sense. Thanks!

PR #455, opened 2016-02-25
Adds WP-API endpoint to fetch post_types. Fixes #334

### LINKED ISSUE #334: Theme specific CPTs / Taxonomies not indexed when using --network-wide flag
Take the example of a custom taxonomy "review-type" being registered by theme code on blog_id 2 - since that site's theme is not loaded with the --network-wide flag, the taxonomy is never indexed into elasticsearch. By using the --url flag and doing site by site, you do get this data (since wp-cli sets up the environment as if you are running that site specifically when using --url)

Same thign for CPTs that are registered in site-specific themes - they do not get indexed.

The following command gets around this for now by indexing each site individually. 

```
#!/bin/bash

for url in $(wp site list --field=url)
do
    echo "Indexing for $url..."
    wp elasticpress index --url=$url
done
```

Ideas to actually solve this problem long term...
- `wp_remote_post()` to each site on the network from the main wp-cli process - return a list of public CPTs and Taxonomies in the POST, and then continue to index those as normal from the main process (remote post only for getting the list of CPTs and Taxonomies avoids potential timeout issues trying to actually index on the POST request) - Could be an issue on some setups or where the url is not publicly accessible

allan23:
Utilizing the WP-API incorporated into WP 4.4, this PR adds a route that returns all post_types that may have been registered from within a theme or plugin that was otherwise not accessible with a CLI indexing process.

### DIFF
--- a/bin/wp-cli.php
+++ b/bin/wp-cli.php
@@ -203,7 +203,7 @@ class ElasticPress_CLI_Command extends WP_CLI_Command {
 		} else {
 			$assoc_args['offset'] = 0;
 		}
-
+		
 		if ( empty( $assoc_args['post-type'] ) ) {
 			$assoc_args['post-type'] = null;
 		}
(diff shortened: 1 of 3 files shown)


ChrisWiegman:
Hey @allan23. Finally took a look at this...

... Two issues.

1.) As a basis for perhaps more endpoints in the future can we abstract the api version for easier reuse.
2.) On sites with more sensitive data, particularly custom post types, I'm concerned that there is no restriction on access to the data. That said, perhaps a way to determine this is from a local cli request or some similar host restriction would be the way to go? Definitely open to thoughts on that one.

PR #456, opened 2016-02-28
Adds Elasticearch Shield username:password login authentication. Fixes #441

### LINKED ISSUE #334: Theme specific CPTs / Taxonomies not indexed when using --network-wide flag

### LINKED ISSUE #441: ElasticSearch Protected with Shield
I was wondering if it is possible to connect EP to an ElasticSearch Server that has been protected with the Shield plugin?

rveitch:
#### Adds username:password basic authentication headers.
Define the constant ES_SHIELD in your wp-config.php
**Format:** `username:password` (colon separated)
**Example:** `define( 'ES_SHIELD', 'es_admin:password' );`

### DIFF
--- a/classes/class-ep-api.php
+++ b/classes/class-ep-api.php
@@ -272,6 +272,20 @@ class EP_API {
 			$headers['X-ElasticPress-API-Key'] = EP_API_KEY;
 		}
 
+    /**
+     * ES Shield Username & Password
+     * Adds username:password basic authentication headers
+     *
+     * Define the constant ES_SHIELD in your wp-config.php
+     * Format: 'username:password' (colon separated)
+     * Example: define( 'ES_SHIELD', 'es_admin:password' );
+     *
+     * @since 1.9
+     */
+		if ( defined( 'ES_SHIELD' ) && ES_SHIELD ) {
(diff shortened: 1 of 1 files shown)


dkotter:
Code looks good and just tested this out, installing Shield locally and it works great. Merging this in for a 1.9 release. Thanks!

PR #464, opened 2016-03-16
localization problem fixed

mustafauysal:
When I try to create language pack for this plugin, realized that language packs couldn't load correctly.


tlovett1:
Awesome, thanks!

Issue #467, opened 2016-03-29
ElasticPress improperly indexes oembed content


ChrisWiegman:
Rather than just indexing the URL, as is stored in the WordPress posts table for post content, the application of the "the_content" filter at https://github.com/10up/ElasticPress/blob/develop/classes/class-ep-api.php#L503 causes WordPress to index the oembed output, not just the URL. This has the result of breaking things like TinyMCE when it is used to retrieve backend content.

AaronHolbrook:
Think this is somewhat related to #293

tlovett1:
@Ritesh-patel anything worth looking at here?

Ritesh-patel:
@tlovett1 ElasticPress still indexes oembed content due to `the_content` filter. I wonder if use of `ep_post_sync_args` or  `ep_post_sync_args_post_prepare_meta` filters can do the work.

tlovett1:
Not sure why indexed EP content would be shown in the editor. Closing for now.

PR #469, opened 2016-03-29
Do not pre-filter content.

ChrisWiegman:
As the_content filter handles items such as oembed it can break things, such as TinyMCE when EP data is used instead of raw data in the database. 

This addresses #467 in the most basic of ways.


tlovett1:
I'm worried about this one. This is radical change from what we're doing.

What about creating another wp_content unfiltered field?

ChrisWiegman:
Backing this out entirely for now. Let me re-approach. This is actually a big one for some users

PR #475, opened 2016-04-07
Add ep_post_index_id filter to _id

Just-Ghali:
Add "ep_post_index_id" filter to allow other plugins to filter the "_id" value of the indexed post in elastic.

On a recent project, we needed the ability to control the index and had no way to edit the _id on elastic. This is useful if the wp application needs to change the mapping to support other applications or optimize the indexing.


dkotter:
@Just-Ghali Reviewed this and looks good, but the formatting is slightly off on a couple of the statements. Basically some of the lines are tabbed in far enough, so they don't line up with the rest. If you could fix those minor formatting issues, I'll get this merged in. Thanks!

PR #476, opened 2016-04-08
Fix meta query paths

mmcachran:
Update meta query paths to post_meta.[KEY].raw to avoid search breaking with meta queries.

### DIFF
--- a/classes/class-ep-api.php
+++ b/classes/class-ep-api.php
@@ -1103,18 +1103,18 @@ class EP_API {
 
 					// Comparisons need to look at different paths
 					if ( in_array( $compare, array( 'exists', 'not exists' ) ) ) {
-						$meta_key_path = 'meta.' . $single_meta_query['key'];
+						$meta_key_path = 'post_meta.' . $single_meta_query['key'];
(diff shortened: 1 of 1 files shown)


allan23:
Hi @mmcachran it appears that this change causes unit tests to fail.

dkotter:
@mmcachran As mentioned above, unit tests are now failing. If you could get those fixed up, that would be great.

Also, can you provide an example of what this fixes? You mention "avoid search breaking with meta queries" but I guess I'm not 100% clear what was actually breaking. Code looks good, besides unit tests needing fixed up, but not exactly sure what I should be testing to make sure this works as expected.

Thanks!

mmcachran:
@dkotter the updates will fix backwards compatibility issues with the key getting changed in newer versions of the plugin. The mapping now contains fields for meta.* and postmeta.* but the search is only for meta.*. For sites using the older version of the plugin the meta.* object doesn't exist. 

If the field name is going to stay changed moving forward it may be easier to just reindex sites when upgrading the plugin rather than changing the code.

tlovett1:
Closing this one.

We moved to the `meta` key. When upgrading from an older version you just need to reindex.

PR #481, opened 2016-04-12
Fix/reduce overhead

ChrisWiegman:
On a client project we realized just how inefficient calls per page can be with as many as 8 on every admin page load. This reduces that by properly caching some calls and removing others.

Note there is a unit test failure currently only with WordPress 4.4.2 and only on PHP 5.3. I've gone as far as installing a box with PHP5.3 yet I still can't reproduce. If you look at the difference between build 1355 and the current there should not be anything affecting that call yet post search with post types disabled is still returning posts only on Travis. Would love a second set of eyes who might be able to see what's going on there.

### DIFF
--- a/classes/class-ep-api.php
+++ b/classes/class-ep-api.php
@@ -1926,6 +1926,18 @@ class EP_API {
 	 */
 	public function get_index_status( $blog_id = null, $force = false ) {
 
+		if ( null === $blog_id ) {
+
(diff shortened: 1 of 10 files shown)


tlovett1:
Adding caching to get_index_status has introduced problems.

tlovett1:
This needs rethinking and testing before it gets merged.

PR #483, opened 2016-04-13
(Closes #478) Enable query logging

### LINKED ISSUE #478: Query log only possible with WP_DEBUG
I think in 1.9 we only add to the query log when `WP_DEBUG` is on. This makes it extremely difficult to debug production sites. I realize it's a performance fix, but we need a way around this ASAP.

lukaspawlik:
This PR solved issue reported in #478 

This seems to be the simplest solution to enable query logging even if `WP_DEBUG` is not enabled. 

I think it would be better if `Debug Bar ElasticPress` plugin would have defined some named constant like `define( 'DEBUG_BAR_ELASTICPRESS_ENABLED', true );` then this PR could use it instead of checking if `Debug Bar ElasticPress` plugin is enabled.

@tlovett1 & @allan23  if you don't like this approach I can go with proposed one - new option on GUI settings page with enabling debug mode via option.

### DIFF
--- a/classes/class-ep-api.php
+++ b/classes/class-ep-api.php
@@ -2053,7 +2053,7 @@ class EP_API {
 	 * @return void Method does not return.
 	 */
 	protected function _add_query_log( $query ) {
-		if ( defined( 'WP_DEBUG' ) && WP_DEBUG ) {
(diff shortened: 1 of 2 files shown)


tlovett1:
Ahhh I love this.

PR #485, opened 2016-04-14
Post meta updates not being indexed

tlovett1:
Right now, when post meta is updated using `update_post_meta`, the updated meta is not sent to Elasticsearch. This is unexpected behavior and causes a huge problem for WooCommerce when stock counts are updated for purchased products.

This PR queues posts for syncing every time whitelisted meta is updated on a whitelisted post type. On shutdown all queued posts are synced. We do the sync on shutdown in case the same post has multiple meta keys updated.

Please review and post feedback on the problem as well as my solution. Pay special attention to the permission checking I'm doing. Also this needs unit tests if anyone wants to write them.

### DIFF
--- a/classes/class-ep-sync-manager.php
+++ b/classes/class-ep-sync-manager.php
@@ -14,6 +14,14 @@ class EP_Sync_Manager {
 	public function __construct() { }
 
 	/**
+	 * Save posts for indexing later
+	 * 
+	 * @since  2.0
+	 * @var    array
+	 */
+	protected $sync_post_queue = array();
+
+	/**
 	 * Setup actions and filters
 	 *
 	 * @since 0.1.2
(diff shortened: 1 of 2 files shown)


tlovett1:
@ChrisWiegman @ivanlopez review this one please

classes/class-ep-sync-manager.php
```diff
@@ -43,6 +53,75 @@ public function destroy() {
...
+
+		// If we have an importer we must be doing an import - let's abort
+		if ( ! empty( $importer ) ) {
+			return;
+		}
+		
+		$indexable_post_statuses = ep_get_indexable_post_status();
+		$post_type               = get_post_type( $object_id );
+
+		if ( ( defined( 'DOING_AUTOSAVE' ) && DOING_AUTOSAVE ) || 'revision' === $post_type ) {
```
ChrisWiegman: perhaps a filterable array of post_types here as well as post_status? I thought we had a filter already in place for post_type but I can't find it for the purpose here.

ChrisWiegman:
Not sure I have a lot of feedback here. The hook seems appropriate and all is working for me locally. My only concern is it might be more efficient to check meta early for protected/index status (ie ep_prepare_meta_allowed_protected_keys filter and related).

PR #493, opened 2016-04-22
Small patch for elasticpress.php, should use require_once instead of require. 

tuanmh:
Should use `require_once` instead of `require` to avoid `PHP Fatal error: Cannot redeclare class` error. This is something actually has happened for me.


dkotter:
Looks good, thanks! Merged in for a 2.0 release.

PR #494, opened 2016-04-24
Support post_parent, fixed #462

### LINKED ISSUE #462: Support post_parent
Title says it all.

tuanmh:


dkotter:
@tuanmh Looks good! Code looks good and I've checked it out and tested locally and it works as expected. Only things I see that need fixing are:

1. The `@since` on the post_parent filter can be set to 2.0, as that's when this will be added.

1. The `@since` on the unit test can also be 2.0

1. We'll want to make sure this support gets documented in our `README.md`, so if you could add that, that would be great.

tlovett1:
Can someone add the documentation to the readme? @ChrisWiegman @ivanlopez

tuanmh:
> The @since on the post_parent filter can be set to 2.0, as that's when this will be added.
> 
> The @since on the unit test can also be 2.0
> 
> We'll want to make sure this support gets documented in our README.md, so if you could add that, that would be great.



> Can someone add the documentation to the readme?

All done via 3f9db24

PR #496, opened 2016-04-26
Filter to not print out post index status

tuanmh:
In our case, we don't really need to show this index status to our users, except administrators, this filter will does the job.


dkotter:
Looks good. Merging in for a 2.0 release.

PR #498, opened 2016-04-26
added do_action( 'ep_add_query_log' ) to _add_query_log

petenelson:
After running into some issues with debugging requests/responses from ElasticSearch, this seems like it would be a handy addition to allow other plugins to hook into this to perform more robust logging without requiring debug turned on or the user to be logged in (based on https://github.com/10up/ElasticPress/commit/82acd71e0a948782da42ae2505ea0d0c4b415755)

### DIFF
--- a/classes/class-ep-api.php
+++ b/classes/class-ep-api.php
@@ -2150,7 +2150,10 @@ class EP_API {
 	}
 
 	/**
-	 * Query logging. Don't log anything when WP_DEBUG is not enabled.
+	 * Query logging. Don't log anything to the queries property when
+	 * WP_DEBUG is not enabled. Calls action 'ep_add_query_log' if you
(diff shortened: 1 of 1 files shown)


dkotter:
@petenelson I think this is fine to add in, and I do agree could make debugging easier at times. My only suggested change or I guess addition would be some documentation. The docblock for the `_add_query_log` function mentions it doesn't log unless `WP_DEBUG` is enabled, but in this case, that actually wouldn't be fully true anymore. So I could see adding another line to the docblock, mentioning how this filter could be used for other logging without the need for `WP_DEBUG` (not sure on exact wording but I think something would be nice).

ChrisWiegman:
Just following up. Doc looks good to me (it makes sense now). This could be ready for merge @dkotter

PR #502, opened 2016-05-02
Support OR relation for tax_query

tlovett1:
Title says it all. Right now we just support `and`.


tlovett1:
This passes with the merging of #501

PR #504, opened 2016-05-03
(Closes #503) Removed slash

### LINKED ISSUE #503: Unable to delete post from ES index
This was fixed in https://github.com/10up/ElasticPress/pull/415/files however it seems to be an issue again here https://github.com/10up/ElasticPress/blob/develop/classes/class-ep-api.php#L242

lukaspawlik:
Double slash // prevents ES from deleting documents from index.


dkotter:
Looks good. Merging in for a 2.0 release.

PR #508, opened 2016-05-13
Tweak search algorithm for more relevant results

tlovett1:
This PR is not ready for merge but rather review. @allan23 @ChrisWiegman @lukaspawlik @dkotter @AaronHolbrook 

Here are the proposed changes:

1. Decrease fuzzy clause fuzziness to 1 from 2. With a fuzziness of 2, the word yell matches yeti since they are two levenshtein changes apart. The point of fuzziness is to account for misspellings which 1 does for the most part. 2 is resulting in way too many bad results.
2. Add `operator => 'and'` to `multi_match` clause. Right now any document containing ANY of the search terms returns a match with a 2x boost. Let's only give the 2x boost to docs that contain ALL the terms.
3. Finally, add a `multi_match` query of type `phrase`. This clause gives a 4x boost to documents containing the terms exactly as is in the EXACT order.

P.S. If anyone wants a cool project, we need a way to unit test relevancy accuracy.

### DIFF
--- a/classes/class-ep-api.php
+++ b/classes/class-ep-api.php
@@ -1337,7 +1337,7 @@ class EP_API {
 							'fields' => $search_fields,
 							'query' => '',
 							'fuzziness' => apply_filters( 'ep_fuzziness_arg', 2, $search_fields, $args ),
-							'operator' => 'or',
+							'operator' => 'and',
 						),
 					)
 				),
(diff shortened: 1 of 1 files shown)


tlovett1:
Also, please review/test this for merging into core: https://github.com/10up/elasticpress-weight-recent

ChrisWiegman:
The search tweaks here look good (we've actually had to filter fuzziness down to one on client sites for the same reasons as you state above). 

Tested with wp_test content and a bit more and it all seems a bit more accurate now with this new tweak (I can actually find my last name mispelled "Weigman" without issue). I think this is fine to merge to core when you're ready.

PR #509, opened 2016-05-17
495 fix version number

### LINKED ISSUE #495: Version 1.9 tells me, it is '1.8'
The tag 1.9 has the version number `1.8` in it's plugin header which leads to confusion when using this plugin via composer:
https://github.com/10up/ElasticPress/blob/1.9/elasticpress.php#L6
Also the constant `EP_VERSION` is defined to `1.8`.

Also, semantic version numbers would be nice. E.g. `1.9.0` instead of `1.9`.

dnaber-de:
This would fix #495 if merged and then tagged (and released) with `1.9.1`. :)


tlovett1:
Fixing this manually now.

Issue #507, opened 2016-05-10
Admin GUI Usability Issues


tlovett1:
This is an ongoing list of admin GUI usability issues:

1. If no index has occurred, the following message is shown: `ElasticPress is not enabled and cannot override WP queries. You can activate it on the form to the left.`. I spent hours trying to understand that message just to find out I needed to index. This is not an acceptable experience.
1. Showing index status within the publish meta box is not user friendly. No one understands what that means except a developer. This should only be showing if `WP_DEBUG` is enabled.
1. On a multisite instance, after inputting a usable host and indexing, the Activate/deactivate checkbox is not shown. If you then save settings, since the checkbox is not shown, search integration is deactivated. The only way to fix is to reindex. This makes the admin GUI completely unusable.
4. The "do not deactivate" checkbox confuses people. They think they are supposed to check it and click save. This experience needs rethinking.
5. `EP_IS_NETWORK` constant. Is this needed? What does it do? There is no documentation.

ivanlopez:
I was unable to reproduce your issue #3 exactly. After inputting a usable host and indexing the use Elasticsearch checkbox was there but not checked. If I hit save changes at that time ElasticPress was disabled. So my solution to that problem was auto select the use Elasticsearch checkbox once indexing is complete. I'll be submitting a pull request shortly with the rest of the updates.

tlovett1:
I just reproduced #3 again locally (multisite only).

First I added a host to the text field and saved. Then I clicked the index button resulting in this screenshot:
![one](https://cloud.githubusercontent.com/assets/1844351/15437557/c70d6516-1e95-11e6-9563-a52f6062806e.png)

Then I clicked save button again resulting in this:
![two](https://cloud.githubusercontent.com/assets/1844351/15437556/c70cb468-1e95-11e6-9add-10a5fd47abce.png)

tlovett1:
Adding a #4 to the list above:

4. The "do not deactivate" checkbox confuses people. They think they are supposed to check it and click save. This experience needs rethinking.

ivanlopez:
What version of WordPress are you using?

tlovett1:
4.5.2

ivanlopez:
I'm still not able to reproduce  but it looks like ep_check_host might be returning an error on your environment. How many sites do you have on your multisite and what plugins do you have enabled?

tlovett1:
Going to close this.

PR #511, opened 2016-05-20
Fix Admin GUI Usability Issues 

ivanlopez:
This fixes issue #507

1. I updated the copy that is displayed when a usable host is entered but Elasticsearch has not been set to override the default WordPress search.  
2. I added a WP_DEBUG check before adding the `post_submitbox_misc_actions` action to prevent the display of the index status to users that don't have WP_DEBUG enabled. 
3. When indexing is complete via the GUI I had it auto select the "Use Elasticsearch" checkbox in order to prevent accidentally disabling Elasticsearch if a user click "Save Changes".


ChrisWiegman:
@tlovett1 Tested and working. This should be ready to be merged.

tlovett1:
@ivanlopez should we merge this or wait for the UI revamp?

ChrisWiegman:
Is a revamp something I can help with? I have nothing but time this week.

ivanlopez:
I would wait for UI revamp.

tlovett1:
Ok, let's merge this stuff now so we don't block 2.0

ivanlopez:
Sure, I'll reach out to you shortly on Slack.

PR #517, opened 2016-05-31
Cache the index names in the DB (Feedback & opinions please)

danielhomer:
As Elasticpress generated index names on the fly, it doesn't play nicely with domain mapping plugins.

For example, we have multiple subdomains across multiple environments:
- uat1.www.site.com.sites.abc.net
- uat2.www.site.com.sites.abc.net
- preprod1.www.site.com.sites.abc.net
- prod4.www.site.com.sites.abc.net
- www.site.com

The above example should use the same index name, regardless of the current domain. However, as the indexes are generated on the fly, accessing one of the sites from a UAT url will generate something along the lines of `uat1wwwsitecomsitesabcnet-2`. By persisting the index names in the DB, we can eliminate this issue.

This still needs some test coverage, but I'm going to hold off until we've managed to have a chat about it, any feedback / opinions appreciated!

### DIFF
--- a/bin/wp-cli.php
+++ b/bin/wp-cli.php
@@ -158,6 +158,57 @@ class ElasticPress_CLI_Command extends WP_CLI_Command {
 	}
 
 	/**
+	 * Clear the index cache for one or all of the blogs
+	 *
+	 * @subcommand clear-index-cache
+	 *
+	 * @param  array $args
+	 * @param  array $assoc_args
+	 */
+	public function clear_index_cache( $args, $assoc_args ) {
+
+		WP_CLI::line( __( 'Clearing index cache...', 'elasticpress' ) );
+		
+		$flushed = ep_flush_index_name_cache( $blog_id );
+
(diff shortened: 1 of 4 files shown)


tlovett1:
Daniel, before you go to far down this road, why not just use the ep_index_name filter?

danielhomer:
I have :) but I thought it may be useful in the plugin and wanted an opinion from you guys.

tlovett1:
Why would you need to cache the index name?

danielhomer:
Because `ep_get_index_name` returns a different index name if you're accessing the same site on a different URL. My thought is that the ES index should be linked to the site, not the domain.

tlovett1:
Can't you handle that logic within the `ep_index_name` filter?

danielhomer:
We can and do, but I was in two minds whether this should be part of the plugin instead. I guess the sort of user that's using domain mapping and ES is capable of using the filter, though...

PR #523, opened 2016-06-24
Order by post__in ASC/DESC support

### LINKED ISSUE #330: Adding orderby post__in support

sc0ttkclark:
#373 strikes back!

Fixes #330 

Now with tests!


sc0ttkclark:
Tests passing, this is golden :)

sc0ttkclark:
@tlovett1 is this in core now? or should I draft a new PR?

tlovett1:
I don't think it is. Can you add a new PR?

sc0ttkclark:
Will do in the coming week or so.

sc0ttkclark:
New PR posted here: #903

PR #530, opened 2016-07-13
Add support for query by post_status

### LINKED ISSUE #529: post_status from WP_Query not respected
There's a filterable list of indexable post statuses for EP, however even if you enable them the post_status of WP_Query is not respected.

By adding a clause below the post_type filtering in class-ep-api.php we can solve that problem.

sc0ttkclark:
Fixes #529 

- [x] Add unit test for WP_Query with post_status
- [ ] Add documentation for post_status argument coverage

### DIFF
--- a/tests/test-single-site.php
+++ b/tests/test-single-site.php
@@ -1008,6 +1008,8 @@ class EPTestSingleSite extends EP_Test_Base {
 	 * @since 2.1
 	 */
 	public function testPostStatusQueryDraft() {
+		add_filter( 'ep_indexable_post_status', array( $this, 'mock_indexable_post_status' ), 10, 1 );
+
 		ep_create_and_sync_post( array( 'post_content' => 'findme test 1', 'post_status' => 'draft' ) );
 		ep_create_and_sync_post( array( 'post_content' => 'findme test 2' ) );
 		ep_create_and_sync_post( array( 'post_content' => 'findme test 3', 'post_status' => 'draft' ) );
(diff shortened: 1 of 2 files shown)


tlovett1:
Is this up to date? There is already code like this here: https://github.com/10up/ElasticPress/blob/develop/classes/class-ep-api.php#L1452

tlovett1:
Nevermind.

sc0ttkclark:
womp

tests/test-single-site.php
```diff
@@ -980,6 +980,101 @@ public function testNoPostTypeSearchQuery() {
...
+		ep_refresh_index();
+
+		$args = array(
+			's'         => 'findme',
+			'post_status' => 'draft',
+		);
+
+		$query = new WP_Query( $args );
+
+		$this->assertEquals( 1, $query->post_count );
```
tlovett1: Should this be 2?

sc0ttkclark: Updated (below)

tests/test-single-site.php
```diff
@@ -980,6 +980,162 @@ public function testNoPostTypeSearchQuery() {
...
+			's'         => 'findme',
+			'post_status' => array(
+				'draft',
+				'publish',
+			),
+		);
+
+		$query = new WP_Query( $args );
+
+		$this->assertEquals( 1, $query->post_count );
```
tlovett1: This should be 3. Using an array for post status is an OR query not AND.

tests/test-single-site.php
sc0ttkclark: You'll have to refresh the diff, the lines moved a little I think

tests/test-single-site.php
```diff
@@ -980,6 +980,101 @@ public function testNoPostTypeSearchQuery() {
...
+	public function testPostStatusQueryMulti() {
+		ep_create_and_sync_post( array( 'post_content' => 'findme test 1', 'post_status' => 'draft' ) );
+		ep_create_and_sync_post( array( 'post_content' => 'findme test 2' ) );
+		ep_create_and_sync_post( array( 'post_content' => 'findme test 3', 'post_status' => 'draft' ) );
+
+		ep_refresh_index();
+
+		$args = array(
+			's'         => 'findme',
+			'post_status' => array(
```
tlovett1: Sorry this should be 3. OR not AND.

tests/test-single-site.php
sc0ttkclark: This should technically only return 3 if 'draft' was a whitelisted post status, which it isn't in this method.

`testPostStatusQueryMultiDefault()` tests we get the proper single result back (1) even though we've passed multiple post statuses because 'draft' is not whitelisted as an indexable post status.

`testPostStatusQueryMulti()` tests we get the proper results back (3) when multiple post statuses are set AND 'draft' is whitelisted as an indexable post status

sc0ttkclark: 1* on that first sentence

PR #532, opened 2016-07-14
Allow caching results of WP_Query during index. Fixes #453

### LINKED ISSUE #453: bulk indexing extremely slow
We upgraded to version 1.8 and realized that bulk indexing extremely slow. The reason of this, calling `wp_cache_flush()` in wp-cli.php

Is there anybody have same problem? (We are using memcached for object caching)

Thanks.

cmmarslender:
There was no significant memory savings by having this here because of
the stop_the_insanity method, and removing it speeds up indexing
significantly


tlovett1:
@mustafauysal this okay with you?

mustafauysal:
@tlovett1 Yes, totally agree with @cmmarslender

PR #533, opened 2016-07-18
Fix query when sites is set in query string.

jaisgit:
When specifying 1 or more site ID's via "sites" var in the query string it is ignored.


tlovett1:
Why not just use an array?

jaisgit:
Was thinking only in terms of strings and overlooked that... Passing an array works.

However passing only one site_id fails conditional as $scope seems to be a string and not and int. Changing is_int to is_numeric seems t fix this.

tlovett1:
A PR for that would be good.

jaisgit:
#534

PR #534, opened 2016-07-18
Fix query string for sites={id}

jaisgit:


tlovett1:
Can we cast to int before passing it along?

classes/class-ep-api.php
```diff
@@ -142,7 +142,7 @@ public function query( $args, $scope = 'current' ) {
 
 		if ( 'all' === $scope ) {
 			$index = ep_get_network_alias();
-		} elseif ( is_numeric( $scope ) ) {
+		} elseif ( is_int( (int) $scope ) ) {
 			$index = ep_get_index_name( $scope );
```
tlovett1: Here is what I meant:

```php
} elseif ( is_numeric( $scope ) ) {
    $index = ep_get_index_name( (int) $scope );
}
```

jaisgit: My thought was to catch it in filter_posts_request() as this is likely to be the only time the argument would be passed as a string. But your suggestion works as well.

tlovett1:
Can you rebase this PR? This should just be one simple commit. Sorry!

PR #539, opened 2016-08-02
Stop loading the EP_Settings class on the front end

craigmillerdev:
Hello

Looks like there's a bug when loading ElasticPress (based on release 2.0) which results in a significant amount of unnecessary remote requests to the ElasticSearch server.

[`ep_loader`](https://github.com/10up/ElasticPress/blob/master/elasticpress.php#L53), hooked on `plugins_loaded`, instantiates a new EP_Settings regardless of need. This triggers a call to [`ep_check_host();`](https://github.com/10up/ElasticPress/blob/master/classes/class-ep-settings.php#L39) resulting the remote request to the ElasticSearch server.

It seems that EP_Settings should be limited to load only in admin. The class loads a handful of admin pages, along with this call to `ep_check_host();`

### DIFF
--- a/tests/includes/class-ep-test-base.php
+++ b/tests/includes/class-ep-test-base.php
@@ -11,7 +11,15 @@ class EP_Test_Base extends WP_UnitTestCase {
 		if ( property_exists( __CLASS__, 'ignore_files' ) ) {
 			self::$ignore_files = true;
 		}
+		$this->plugin_path = str_replace( '/tests/includes', '', dirname( __FILE__ ) );
 	}
+
+	/**
+	 * Stores the root path for the plugin
+	 *
+	 * @var string
+	 */
+	protected $plugin_path = '';
 
 	/**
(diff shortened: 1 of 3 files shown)


tests/includes/class-ep-test-base.php
```diff
@@ -11,9 +11,17 @@ public function __construct() {
 		if ( property_exists( __CLASS__, 'ignore_files' ) ) {
 			self::$ignore_files = true;
 		}
+		$this->plugin_path = str_replace( '/tests/includes', '', __DIR__ );
```
tlovett1: __DIR__ isn't available in PHP 5.2

tlovett1:
This is a huge bug. Mind fixing the one thing real quick so I can merge?

craigmillerdev:
I've updated `__DIR__` for a good ol' `dirname( __FILE__ )`

PR #563, opened 2016-08-28
[cli] Do not activate plugin integration after indexing IF it was dea…

wkw:
…ctivated at start.
When indexing from wp-cli, if EP integration is inactive, do not reactivate upon index completion.


tlovett1:
Not really seeing what this does. The `activate` method was removed in 2.1 (beta).

wkw:
OK, sorry. Didn't see that. I'm still on 2.0.1.

PR #564, opened 2016-08-29
Fixed: support post_parent = 0

tuanmh:
- Important for searching unattached media (post_parent = 0)
- empty( $args['post_parent’]): returns true if $args['post_parent’] = 0


Ritesh-patel:
@tlovett1 this looks good to me.

PR #565, opened 2016-08-29
Fix/560

Ritesh-patel:
- Check remote request response code of Elasticsearch query and fallback to WP query if not allowed valid request response code
- Add PhpUnit test case for "If index is deleted, search should fallback to WP Query"

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 /vendor/
 composer.lock
 node_modules
+.idea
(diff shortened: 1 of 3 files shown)


tlovett1:
`ep_remote_request` is used for a lot of things. I have a feeling we need to accept all 200's not just 200 itself.

Ritesh-patel:
In 933a2f8d64f53128d43999c4ebf1f0d950370af2 make all remote requests valid with 2xx status.

tlovett1:
Nice work!

PR #569, opened 2016-08-31
Do not activate the module if does not meet dependency requirement

Ritesh-patel:


tlovett1:
I'm confused on this PR. If a module's dependencies aren't met, the activate button is not available. Correct?

Ritesh-patel:
Correct, it's not available. But can be available if we just alter HTML from source and remove a class property.

tlovett1:
Why is that an issue? Is the use case is someone is modifying elements in the DOM manually?

Ritesh-patel:
Yes, if someone modifies elements in DOM manually.

Ritesh-patel:
I know it's very rare that someone will do that but upon doing that, it gives fatal error.

tlovett1:
Can you update the PR not to update the .pot file?

Ritesh-patel:
PR updated.

PR #583, opened 2016-09-23
Abort action queue meta sync if EP is not alive

### LINKED ISSUE #582: action_queue_meta_sync function attempts to index even when EP is not alive
Noticed this during a migration that does one-off importing, the function `action_queue_meta_sync` attempts to process even if the Elasticsearch host application is unreachable.

This causes increased processing time on each item as the remote request must timeout before proceeding.

AaronHolbrook:
Fixes #582


tlovett1:
Thanks, @AaronHolbrook!

Issue #580, opened 2016-09-20
Support for orderby rand


jonathan-dejong:
Hi!

Just wondering wether it's even possible to support orderby rand? 
I've got a query where I set orderby to `meta_value rand`. While meta_value works rand does not and I notices it's not supported. 

This is not a pretentious request. I'm wondering wether there is a way to support rand at all? I have no insight in how the querying towards elasticsearch works :)

tlovett1:
Definitely possible. Not something we've added yet.

jonathan-dejong:
Cool! Great to hear.. I spent about 30min trying to figure out the code of the plugin but I could not find the relevant locations to tinker with :)

Ritesh-patel:
@tlovett1 @jonathan-dejong this has been added in #584 however, orderby rand will work standalone, `orderby => meta_value rand` won't give expected results.

jonathan-dejong:
Hi @Ritesh-patel 

Thanks! Is it due to a technical incompatibility? Or just not "fixed" yet? 
I very much appreciate you working on this but for it to be useful in my particular situation I'd need the ability to filter by `meta_value rand` :)

Ritesh-patel:
Hi @jonathan-dejong 

I think it's technical incompatibility. In Elasticsearch I can't find any way to sort on random order along with other fields.

jonathan-dejong:
Damn that's not good news.. :/

PR #584, opened 2016-09-25
Added support for "orderby rand"

Ritesh-patel:
Fix for #580 
- Added support for "orderby rand"
- This won't work if there is any other field to orderby with "rand", Elasticsearch doesn't have mechanism yet using which we can sort on multiple fields in conjunction with random order.


tlovett1:
Not sure why you would want to order by any other field in conjunction with rand.

tlovett1:
Can you push a unit test?

Ritesh-patel:
@tlovett1 you are right, we don't want any other field with rand. It's just that WP_Query does allow that but upon doing that Elasticsearch ignores random order and just order the results by another field provided in WP_Query.

So `'orderby' => 'rand post_date'` will only sort results by post_date in Elasticsearch.

Ritesh-patel:
@tlovett1 added test case in dc2f46ac69603f3722f42c1c7b5f52f21a2bbd80

PR #600, opened 2016-10-12
Fix post_type array on bulk action filtered by category

danielhuesken:
In WooCommerce products backend bulk actions not working, if they filtered by a category.
Than the form (for bulk actions) filed post_type has a value of Array.
Also in the Array is product defined twice.


tlovett1:
@Ritesh-patel, can you test this one?

Ritesh-patel:
@tlovett1 this looks good to me.

PR #601, opened 2016-10-13
removing unused function ep_elasticsearch_alive

mustafauysal:


tlovett1:
Not sure if it makes sense to remove this (backwards compat)?

mustafauysal:
but method doesn't exist – https://github.com/10up/ElasticPress/commit/b9075df#diff-9703716231f66c3ddbc0030ef4dc582cL1713

tlovett1:
You're right! I missed that.

PR #610, opened 2016-10-25
Merge Network enabled modules with Site enabled modules

craigmillerdev:
Currently if you network enable the plugin, you cannot activate any modules on a site by site basis. This is a suggested PR which will enable the plugin to be network active, and sites to maintain their own configuration of active modules.

The wp-cli command list-modules may need tweaking to list modules and how they are activated for the current site.

### DIFF
--- a/classes/class-ep-modules.php
+++ b/classes/class-ep-modules.php
@@ -66,10 +66,9 @@ class EP_Modules {
 	 * @since  2.1
 	 */
 	public function activate_module( $slug ) {
+		$modules = get_option( 'ep_active_modules', array() );
 		if ( defined( 'EP_IS_NETWORK' ) && EP_IS_NETWORK ) {
-			$modules = get_site_option( 'ep_active_modules', array() );
-		} else {
-			$modules = get_option( 'ep_active_modules', array() );
+			$modules = array_merge( $modules, get_site_option( 'ep_active_modules', array() ) );
 		}
 
 		if ( false === array_search( $slug, $modules ) ) {


tlovett1:
Can you explain how this scenario occurs? If the plugin is network enabled, how would you site-by-site activate modules?

craigmillerdev:
Of the top of my head because of this:
https://github.com/10up/ElasticPress/blob/develop/bin/wp-cli.php#L71-L75

Network enable the plugin, then individually activate modules on different sites.

tlovett1:
I'm still not seeing how the situation you are proposing to fix even occurs. @Ritesh-patel, am I missing something?

craigmillerdev:
Huh... maybe it's me missing something...

Check: https://github.com/10up/ElasticPress/blob/develop/classes/class-ep-modules.php#L69-L73

If you network activate the plugin `EP_IS_NETWORK` gets defined as `true` therefore loaded modules get read from `get_site_options()`. If you use the WP CLI commands to enable a module without the `--network` flag the option is set using `update_option()`.

It might be the opposite of this PR is required depending on what was intended. i.e. check if the plugin is network activated and set the option appropriately when using WP CLI. In our case it's useful to have the elasticpress network-enabled but maintain control over loaded modules on a site-by-site basis.

Ritesh-patel:
@tlovett1 the scenario I am thinking that maybe on certain sites in network one want admin module activated while not on others. Not sure if that's what you trying to achieve @craigmillerdev

craigmillerdev:
It was a while ago I was working on this now... I expect I would have been activating and deactivating the `search` module on a site by site basis whilst having the plug-in network activated. We have the admin interface permanently disabled on all sites and manage EP via CLI.

This PR will allow a module to be activated either network wide (as the admin interface seems to) or on individual site basis (as the CLI commands do)

However! We just deactivated the plugin network wide and activated it on a site by site basis avoiding this confusion entirely. Therefore you might instead want to consider updating the CLI commands to first check if the plug-in is network activated and refusing to activate/deactivate modules unless the `--network` flag is passed .

Feel free to close this PR if you don't consider this necessary, this should have been created as an issue for discussion in the first place. :)

tlovett1:
I would definitely be open to a PR that throws an error if a network activated EP receives a CLI command to single site activate a module. Thanks!

PR #611, opened 2016-10-28
Fix admin gui locked if wp cli indexing fails by using transient

### LINKED ISSUE #602: If indexing via WPCLI fails the admin gui is locked with the WPCLI in syncing message
Today while indexing a large site I had WPCLI time out. When I went to the GUI to try and index from there the GUI was disabled with the WPCLI is syncing message. I was able to get around it by opening up the chrome inspector. I'm thinking we need to implement a try catch so that if anything goes wrong while indexing via WPCLI the admin GUI does not get locked down.

Ritesh-patel:
Fix #602


bin/wp-cli.php
```diff
@@ -804,4 +832,21 @@ private function _connect_check() {
 			WP_CLI::error( __( 'Unable to reach Elasticsearch Server! Check that service is running.', 'elasticpress' ) );
 		}
 	}
+	
+	/**
+	 * Maybe reset transient while indexing
+	 *
+	 * @since 2.1.1
+	 */
+	private function maybe_reset_transient() {
```
tlovett1: I don't see the point of "maybe" reseting the transient. Why not just reset it every time?

tlovett1:
Can we prioritize this one @Ritesh-patel ?

bin/wp-cli.php
Ritesh-patel: I am resetting transient only when it expired but yeah that won't make much difference if we reset it during each sync request.

Ritesh-patel:
@tlovett1 I have updated PR.

PR #617, opened 2016-10-31
When using the 'ep_wp_query_search_cached_posts' filter, if it return…

jvolfson:
…s a non empty response $ep_query is never set and as a result line 303 generates a PHP warning indicating so. Added a check to see if it is set before attempting to use it.


Ritesh-patel:
@jvolfson you can also set default value of $ep_query to empty array before if( count( $new_posts ) < 1 ) statement at line 220. This way any future reference to $ep_query will have a default value.

jvolfson:
Modified approach to set default value.

Ritesh-patel:
@tlovett1 I thinks it's good to merge.

Issue #621, opened 2016-11-02
Respect actuall indexed post types in Woocommerce module (ep_wc_translate_args)


dnaber-de:
The function `ep_wc_translate_args()` [assumes](https://github.com/10up/ElasticPress/blob/d1f8ffcbb2bb95f78ffaad2a208ac798ab3b63f9/modules/woocommerce/woocommerce.php#L291) that the post types `shop_order`, `shop_order_refund`, `product` and `product_variations` are always be indexed. It would be nice to apply the `ep_indexable_post_types` filter here in order to be compatible to possible changes from third-party code.

**Update** A even better solution would be to ask for `ep_get_indexable_post_types()`.

allan23:
Hi @dnaber-de,

Thank you for reporting this issue and submitting a PR for it.

dnaber-de:
You're welcome.

PR #623, opened 2016-11-03
Integrate ES only on indexed post types #621

dnaber-de:
Description see #621 

This fix limits integration to actuall indexed post types.
If the post types aren't filtered, the behavior doesn't
change at all.

Actually it would be enough to ask just for `ep_get_indexable_post_types` IMO but the module itself indexes more post types than this code wants to integrate ES with. That's why I used `array_intersect()` to remove all not-indexed post types from `$supported_post_types`.


modules/woocommerce/woocommerce.php
```diff
@@ -288,11 +288,14 @@ function ep_wc_translate_args( $query ) {
...
-		'shop_order_refund',
-		'product_variation'
+	$supported_post_types = array_intersect(
+		array(
+			'product',
+			'shop_order',
+			'shop_order_refund',
+			'product_variation'
+		),
+		ep_get_indexable_post_types()
```
tlovett1: Can't you just check `ep_get_indexable_post_types()` since we are filtering that early on?

dnaber-de: Not without changing the current behavior of the module. The module currently defines more indexed post types than it actually adapts the query for:

```
function ep_wc_post_types( $post_types ) {
	return array_unique( array_merge( $post_types, array(
		'shop_order' => 'shop_order',
		'shop_coupon' => 'shop_coupon',
		'shop_order_refund' => 'shop_order_refund',
		'product_variation' => 'product_variation',
		'product' => 'product',
	) ) );
}
```

```
function ep_wc_translate_args( $query ) {
	/* … */
	$supported_post_types = array(
		'product',
		'shop_order',
		'shop_order_refund',
		'product_variation'
	);
```

In fact, if you just ask for `ep_wc_post_types()` as supported post types, you would then also adapt the query for `shop_coupon` post type. I don't know what's your intention here. Thats why I created the PR in a way, that it doesn't change the behavior besides the obvious bug. Thats what I meant with:

> Actually it would be enough to ask just for ep_get_indexable_post_types IMO but the module itself indexes more post types than this code wants to integrate ES with. That's why I used array_intersect() to remove all not-indexed post types from $supported_post_types.

tlovett1:
Mind fixing the merge conflict?

dnaber-de:
Sure… but not today :)

dnaber-de:
Fixed

Issue #632, opened 2016-11-15
QA Version 2.2


tlovett1:
2.2 beta has been pushed to the development branch (see readme.txt for list of updates).

We need to QA all the feature API changes. Also, the `must not` tax query code seems to be off. I left a todo in the API class about it.

Ritesh-patel:
I found 2 issues while doing QA

- In the changelog it says, "Modules are now auto-on", but when I update from 2.1.2 to 2.2, all the features were off.
- While doing index using CLI, I got following notice in error log 
```
Notice: ep_elasticsearch_alive is <strong>deprecated</strong> since version ElasticPress 2.1! Use ep_elasticsearch_can_connect() instead.
```

#633 fixes deprecated notice and tax query issue.

Ritesh-patel:
ed3de5b7d1ac23109e1ecc0012f43bb06e29f36a fixes the feature auto-update issue on plugin update.

j3j5:
I am testing this branch and I've encountered a lot of this errors on the ES servers while building the index:

```
Nov 18 17:24:53 : org.elasticsearch.index.mapper.MapperParsingException: failed to parse
Nov 18 17:24:53 :         at org.elasticsearch.index.mapper.DocumentParser.wrapInMapperParsingException(DocumentParser.java:175) ~[elasticsearch-5.0.1.jar:5.0.1]
Nov 18 17:24:53 :         at org.elasticsearch.index.mapper.DocumentParser.parseDocument(DocumentParser.java:69) ~[elasticsearch-5.0.1.jar:5.0.1]
Nov 18 17:24:53 :         at org.elasticsearch.index.mapper.DocumentMapper.parse(DocumentMapper.java:274) ~[elasticsearch-5.0.1.jar:5.0.1]
Nov 18 17:24:53 :         at org.elasticsearch.index.shard.IndexShard.prepareIndex(IndexShard.java:529) ~[elasticsearch-5.0.1.jar:5.0.1]
Nov 18 17:24:53 :         at org.elasticsearch.index.shard.IndexShard.prepareIndexOnPrimary(IndexShard.java:506) ~[elasticsearch-5.0.1.jar:5.0.1]
Nov 18 17:24:53 :         at org.elasticsearch.action.index.TransportIndexAction.prepareIndexOperationOnPrimary(TransportIndexAction.java:174) ~[elasticsearch-5.0.1.jar:5.0.1]
[...]

PR #633, opened 2016-11-15
2.2 fixes

Ritesh-patel:
- Fix tax_query "OR" relation support in EP API
- Use ep_elasticsearch_can_connect() instead of ep_elasticsearch_alive()


Ritesh-patel:
Fixes #632

elasticpress.php
```diff
@@ -55,37 +55,46 @@
...
 
 /**
- * On activate, all features that meet their requirements with no warnings should be activated.
+ * All features that meet their requirements with no warnings should be activated.
  *
- * @since  2.1
+ * @since 2.2
  */
-function ep_on_activate() {
+function ep_auto_activate_features() {
```
tlovett1: Does it make sense to move this to class-ep-features.php?

Ritesh-patel: Yes, we can move that as it's feature specific.

tlovett1:
For the auto activate piece, are you just moving the code to another file? If so, let's remove that part which should solve the conflict.

Ritesh-patel:
2 different functions uses the same code so I have moved that code into a wrapper function. PR has been updated accordingly.

tlovett1:
Ritesh please just separate this into multiple PRs. The auto-activate function you created does two completely different things (2nd is setting an upgrade needed option)

Ritesh-patel:
@tlovett1 running the plugin activation code on `plugins_loaded` hook doesn't sound good. I think maintaining an option which can tell whether to upgrade or not is still a valid option. This is the case basically for plugin updates where `register_activation_hook` won't be called.

tlovett1:
Let's table this one until our next meeting so we can discuss.

tlovett1:
I pushed the taxonomy fix as a separate commit as it was breaking tests.

Issue #639, opened 2016-11-20
WooCommerce Feature Issues


tlovett1:
Received an email with the following issues:

- [x] New product Drafts are not showing in the product list 
- [x] Image search in the media area and when searching an image in the feature product not working
- [x] Searching in the customer order screen not returning results
- [ ] Running the woocommerce google feed plugin …when google fetches it will only go to 250 products then stop.  I deactivate elsaticpress and It will fetch the remaining products.
- [ ] Not Displaying draft previews correctly
- [ ] The search for products when in a customer orders not going through elastic search

Ritesh-patel:
> New product Drafts are not showing in the product list

Not showing in All Products, this issue is for all the post types. The fix I have think of is to set post_status filter in EP query if in admin dashboard. We can make it to not include post statuses using `get_post_stati` function.

> Image search in the media area and when searching an image in the feature product not working

It's a confirmed bug and I have fixed it in #639.

> Searching in the customer order screen not returning results

I have tested this but it's working fine for me all the time.

> Running the woocommerce google feed plugin …when google fetches it will only go to 250 products then stop. I deactivate elsaticpress and It will fetch the remaining products.

I will check this once I have access to this plugin.

> Not Displaying draft previews correctly

Not sure if related to EP but working fine for me. Need to ask for any specific scenario or use case.

> The search for products when in a customer orders not going through elastic search

WooCommerce doesn't utilise WP_Query for this search neither it has any filter using which we can search through Elasticsearch.

PR #641, opened 2016-11-22
PR for #639

Ritesh-patel:


classes/class-ep-api.php
```diff
@@ -1622,7 +1622,7 @@ public function format_args( $args ) {
 		if ( ! empty( $args['post_status'] ) ) {
 			// should NEVER be "any" but just in case
 			if ( 'any' !== $args['post_status'] ) {
-				$post_status = (array) $args['post_status'];
+				$post_status = (array) ( is_string( $args['post_status'] ) ? explode( ',', $args['post_status'] ) : $args['post_status'] );
```
tlovett1: Probably need to strip this of spaces too :/

Ritesh-patel: I have applied `trim` function mapped to each array element. It should fix the issue.

PR #642, opened 2016-11-23
Fix post_type product getting set in any WP_Query if tax_query is provided in WooCommerce feature

### LINKED ISSUE #640: Issue with Polylang plugin and localized home pages showing error 404 when ElasticPress activated
Hello,
I am using Wordpress 4.6.1 with Polylang plugin version 2.0.9. 
I just installed ElasticPress version 2.1.2 and I meeting an issue with all the localized home pages which now are displaying an 404 error : https://mydomain/fr/ or https://mydomain/en/ etc...
When I deactivate ElasticPress, my home pages are displaying well again.
Could you please check this? Thank you in advance.
Best regards

Ritesh-patel:
Fixes #640


tlovett1:
This doesn't seem right to me. What if the query contains other tax queries that aren't WC? It wouldn't integrate unless a WC taxonomy was found.

Ritesh-patel:
You are right, it wouldn't integrate unless there's a WC taxonomy and might strip other taxonomies. The Issue here is that for any tax_query, post_type is getting set as product. I have updated commit and now it will check for integration, if it can integrate than and only then it will set post_type to product.

tlovett1:
I see only one line change in the commit?

Ritesh-patel:
Yep, that should fix the issue. We are setting `$integrate` flag but weren't using it anywhere.

Ritesh-patel:
@tlovett1 I have updated PR which will now set query vars if we have WooCommerce specific query.

tlovett1:
@Ritesh-patel I pushed a fix. Can you make sure it works?

Ritesh-patel:
@tlovett1 this is working fine at my end.

Issue #644, opened 2016-11-29
Call to undefined method EP_Feature::dependencies_met()


lukaspawlik:
When I tried to execute

```
wp elasticpress activate-feature search
```

I got error as follows:

> Fatal error: Uncaught Error: Call to undefined method EP_Feature::dependencies_met() in wp-content/plugins/elasticpress/bin/wp-cli.php:59

Ritesh-patel:
Thanks Lukas for bringing this to our notice and for PR.

PR #645, opened 2016-11-29
Fix for wp-cli issues

lukaspawlik:
It fixes https://github.com/10up/ElasticPress/issues/644 and issues with features activation/deactivation/listing using wp-cli.

### DIFF
--- a/bin/wp-cli.php
+++ b/bin/wp-cli.php
@@ -47,9 +47,9 @@ class ElasticPress_CLI_Command extends WP_CLI_Command {
 		}
 
 		if ( ! empty( $assoc_args['network-wide'] ) ) {
-			$active_features = get_site_option( 'ep_active_features', array() );
+			$active_features = get_site_option( 'ep_feature_settings', array() );
 		} else {
-			$active_features = get_option( 'ep_active_features', array() );
+			$active_features = get_option( 'ep_feature_settings', array() );
 		}
 
 		if ( $feature->is_active() ) {
@@ -60,7 +60,8 @@ class ElasticPress_CLI_Command extends WP_CLI_Command {
 			WP_CLI::error( __( 'Feature depedencies are not met', 'elasticpress' ) );
 		}
 
-		$active_features[] = $feature->slug;
+		$active_features[ $feature->slug ]           = wp_parse_args( $active_features[ $feature->slug ], $feature->default_settings );
+		$active_features[ $feature->slug ]['active'] = true;
 
 		$feature->post_activation();
 
@@ -69,9 +70,9 @@ class ElasticPress_CLI_Command extends WP_CLI_Command {
 		}
 
 		if ( ! empty( $assoc_args['network-wide'] ) ) {
-			update_site_option( 'ep_active_features', $active_features );
+			update_site_option( 'ep_feature_settings', $active_features );
 		} else {
-			update_option( 'ep_active_features', $active_features );
+			update_option( 'ep_feature_settings', $active_features );
 		}
 
 		WP_CLI::success( __( 'Feature activated', 'elasticpress' ) );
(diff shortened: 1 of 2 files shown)


Ritesh-patel:
Hi @lukaspawlik 

When I checked this PR for WooCommerce feature it says `Success: Feature activated` even if WooCommerce plugin wasn't active. I think instead of `dependencies_met` we need to check for `requirements_status`.

Regards,
Ritesh

lukaspawlik:
Hi @Ritesh-patel ,

My assumption was that this kind of check should be done by `dependencies_met_cb` callback defined in the feature rather than using `requirements_status` function inside `EP_Feature` class. I believe that both approaches are correct - only final place where things are checked is different.

Ritesh-patel:
Hi @lukaspawlik 

`ep_register_feature` doesn't support `dependencies_met_cb` parameter so I think we should stick to using `requirements_status`. We also need to update registration of WooCommerce feature.

Regards,
Ritesh

lukaspawlik:
Hi @Ritesh-patel ,

Why do you think that `ep_register_feature` doesn't support `dependencies_met_cb` ?

Constructor of `EP_Feature` does as follows:

```
/**
	 * Initiate the feature, setting all relevant instance variables
	 *
	 * @since  2.1
	 */
	public function __construct( $args ) {
		foreach ( $args as $key => $value ) {
			$this->$key = $value;
		}

		do_action( 'ep_feature_create', $this );
	}
```

So if `dependencies_met_cb` is set as an argument in `ep_register_feature` it's correctly set in `EP_Feature` instance class the same way as `requirements_status_cb` is.

Am I missing something here?

lukaspawlik:
@Ritesh-patel I think that the more important question is if we really need to call `$feature->dependencies_met()` in `wp-cli`? Maybe it's a typo and we actually should call `$feature->requirements_met()`? My assumption was that these two have different purposes however they can do the same. If it was a bigger typo then let's use only one `requirements_met` in `wp-cli` instead of `dependencies_met`.

Ritesh-patel:
@lukaspawlik See supported array parameter of $feature_args arguments of EP_Features->register_feature.
Yes, you are right `dependencies_met` and `requirements_met` doing the same thing, and we should use `requirements_met` in CLI. I think `requirements_met` should call `requirements_status` and need to check for status code, if status code is 2, we shouldn't activate that feature.

lukaspawlik:
@Ritesh-patel I see docblock information.

Right now I am confused as README.md says as follows :smile: 

## Custom Features

ElasticPress has a robust API for registering your own features. Refer to the code within each feature for detailed examples. To register a feature, you will need to call the `ep_register_feature()` function like so:

```php
add_action( 'plugins_loaded', function() {
    ep_register_feature( 'slug', array(
        'title' => 'Pretty Title',
        'setup_cb' => 'setup_callback_function',
        'feature_box_summary_cb' => 'summary_callback_function',
        'feature_box_long_cb' => 'long_summary_callback_function',
        'requires_install_reindex' => true,
        'dependencies_met_cb' => 'dependencies_meta_callback_function',
        'post_activation_cb' => 'post_activation_callback_function',
    ) );
} );
```

The only arguments that are really required are the `slug` and `title` of the associative arguments array. Here are descriptions of each of the associative arguments:

* `title` (string) - Pretty title for feature
* `requires_install_reindex` (boolean) - Setting to true will force a reindex after the feature is activated.
* `setup_cb` (callback) - Callback to a function to be called on each page load when the feature is activated.
* `post_activation_cb` (callback) - Callback to a function to be called after a feature is first activated.
* `feature_box_summary_cb` (callback) - Callback to a function that outputs HTML feature box summary (short description of feature).
* `feature_box_long_cb` (callback) - Callback to a function that outputs HTML feature box full description.
* `dependencies_met_cb` (callback) - Callback to a function that determines if the features dependencies are met. True means yes, WP_Error means no. If no, WP_Error message will be printed to the screen.

I guess we need to clarify what exactly is needed here :confused:

tlovett1:
Hey guys, `dependencies_met` is a relic of 2.1. I fixed all of this. Please confirm.

Ritesh-patel:
@lukaspawlik I think we should open a new PR.

Issue #643, opened 2016-11-28
"Fielddata is disabled on text fields by default." Admin Feature


bcumber:
I get the following exception from Elasticpress when I load the "Pages" in the admin screen:

```json
{
    "error": {
        "root_cause": [
            {
                "type": "illegal_argument_exception",
                "reason": "Fielddata is disabled on text fields by default. Set fielddata=true on [post_title.sortable] in order to load fielddata in memory by uninverting the inverted index. Note that this can however use significant memory."
            }
        ],
        "type": "search_phase_execution_exception",
        "reason": "all shards failed",
        "phase": "query",
        "grouped": true,
        "failed_shards": [
            {
                "shard": 0,
                "index": "devboxbennet-1",
                "node": "VbXshdquSwqnaeSD02K7vQ",
                "reason": {
                    "type": "illegal_argument_exception",
                    "reason": "Fielddata is disabled on text fields by default. Set fielddata=true on [post_title.sortable] in order to load fielddata in memory by uninverting the inverted index. Note that this can however use significant memory."
                }
            }
        ],
        "caused_by": {
            "type": "illegal_argument_exception",
            "reason": "Fielddata is disabled on text fields by default. Set fielddata=true on [post_title.sortable] in order to load fielddata in memory by uninverting the inverted index. Note that this can however use significant memory."
        }
    },
    "status": 400
}
```

If I understand this error correctly (and I'm new to Elasticsearch, Wordpress, and ElasticPress so I could very well be misunderstanding) version 5.x of Elasticsearch does not index text fields for sorting by default. (This does seem to differ from previous versions [5.x](https://www.elastic.co/guide/en/elasticsearch/reference/current/fielddata.html#_fielddata_is_disabled_on_literal_text_literal_fields_by_default) vs [2.4](https://www.elastic.co/guide/en/elasticsearch/reference/2.4/fielddata.html))

The 5.x documentation provides a way to enable fielddata for text fields. I tried to do this:
```json
PUT devboxbennet-1/_mapping/post
{
  "properties": {
    "post_title": { 
      "type":     "text",
      "fields": {
        "post_title": {
          "type": "text",
          "analyzer": "standard"
        },
        "raw": {
          "type": "keyword"
        },
        "sortable": {
          "type": "text",
          "analyzer": "ewp_lowercase",
          "fielddata": true
        }
      }
    }
  }
}
```

[...]

PR #648, opened 2016-11-30
Potential mapping solution for #643

bcumber:
This mapping seems to work for me. I tried to keep the `analyzer` parameter in the `sortable` field, but Elasticsearch doesn't like it. Not sure if there is an alternative. Perhaps the standard analyzer is sufficient?


tlovett1:
So my issue is that sortable intentionally lowercased things. We doing want case sensitivity in sorting. How do we address that?

tlovett1:
@Ritesh-patel will create another PR to lowercase things before going into the sortable field.

PR #662, opened 2016-12-15
Update ep_delete_post, include $post_type argument

timbunch:
I found it useful to add an optional $post_type to ep_delete_post(). This small change helps me process large amounts of custom post types quickly.


tlovett1:
Glad this helped! I don't think this makes sense for a PR. Happy to add a filter to the function to help you.

timbunch:
Thank you!

PR #667, opened 2017-01-03
elasticpress_enabled to return false if host is not set

Svanmark:


tlovett1:
It probably makes more sense to handle this in `elasticpress.php`, only including certain classes if a host is set.

Svanmark:
Yepp, makes sense in terms of saving a bunch of bytes when loading the plugin. 

I'll have it updated. 
Thank you.

tlovett1:
Any update?

Svanmark:
Nothing to verify just yet, just got my fork up to date.

Svanmark:
Seems like it's fixed already, closing. Thanks.

Issue #660, opened 2016-12-13
Media Indexing


tlovett1:
Functionality to index binary media content such as PDFs.

elvismdev:
@tlovett1 I see [here](https://github.com/10up/ElasticPress/commit/32bb16eb5502cfd47a92f94d73dc125ca2bfcbec#diff-04c6e90faac2675aa89e2176d2eec7d8R59) you guys added this feature recently. 

Is there any extra work needed to be done on the Elasticsearch instance for this to function properly? 

Or just by installing and configuring ElasticPress, users will be able to search for texts contents inside PDF or Doc files, and in WordPress search results will return the links for those?

Thank you

tlovett1:
You will need the [ingest attachment](https://www.elastic.co/guide/en/elasticsearch/plugins/master/ingest-attachment.html) plugin for the feature to work.

PR #672, opened 2017-01-05
media indexing feature #660

Ritesh-patel:


tlovett1:
Moved to feature plugin.

Issue #680, opened 2017-01-17
meta_query issue


Ritesh-patel:
ElasticPress ignores two dimension array in meta_query. For example, consider following meta_query parameter of WP_Query:

```
'meta_query' => array(
	'relation' => 'AND',
	array(
	  'key' => 'meta_key_1',
	  'value' => '1',
	  'compare' => 'NOT EXISTS',
	),
	array(
	    'relation' => 'OR',
		array(
		  'key' => 'meta_key_2',
		  'value' => date( 'Y-m-d' ),
		  'compare' => '>=',
		),
		array(
		  'key' => 'meta_key_3',
		  'value' => '1',
		  'compare' => 'NOT EXISTS',
		),
	),
),
```

The ElasticPress query derived from this WP_Query contains only `meta_key_1` filter.

tlovett1:
CC @pereirinha

PR #681, opened 2017-01-17
Fix multidimensional meta_query array #680

Ritesh-patel:


tlovett1:
How many levels of nesting does this support?

classes/class-ep-api.php
```diff
@@ -1728,6 +1494,294 @@ public function format_args( $args ) {
...
+				 * 	 	...
+				 * 	 )
+				 * 	)
+				 * )
+				 *
+				 * Somehow WordPress (WooCommerce) handles that case so we need to as well.
+				 *
+				 * @since  2.1
+				 */
+				if ( is_array( $single_meta_query ) && empty( $single_meta_query['key'] ) ) {
```
tlovett1: Wondering if this would conflict with your code?

tlovett1:
@Ritesh-patel can you tell me what has actual changed? Moving the code into a recursive method made it impossible to see.

classes/class-ep-api.php
Ritesh-patel: It will return true for this but won't affect. For a safe side we can add `empty( $single_meta_query['relation'] )` check.

Ritesh-patel:
@tlovett1 there isn't any limit for nesting.

classes/class-ep-api.php
```diff
@@ -1728,6 +1494,294 @@ public function format_args( $args ) {
...
+							}
+							
+							break;
+					}
+					
+					// Add the meta query filter
+					if ( false !== $terms_obj ) {
+						$meta_filter[] = $terms_obj;
+					}
+				} elseif ( is_array( $single_meta_query ) && isset( $single_meta_query[0] ) && is_array( $single_meta_query[0] ) ) {
```
Ritesh-patel: @tlovett1 this is the exact change. It will check if single query is array, it will build nested object for that array with bool should/must filter.

Ritesh-patel:
```
'meta_query' => array(
	'relation' => 'AND',
	array(
	  'key' => 'meta_key_1',
	  'value' => '1',
	),
	array(
		'relation' => 'OR',
		array(
		  'key' => 'meta_key_2',
		  'value' => '2',
		  'compare' => '>=',
		),
		array(
		  'key' => 'meta_key_3',
		  'value' => '4',
		),
	),
),
```
will become
```
{
    "bool": {
        "must": [
            {
                "terms": {
                    "meta.meta_key_1.raw": [
                        "1"
                    ]
                }
            },
            {
                "bool": {
                    "should": [
                        {
                            "bool": {
                                "must": [
                                    {
                                        "range": {
                                            "meta.meta_key_2.double": {
                                                "gte": "2"
                                            }
                                        }
                                    }
                                ]
                            }
                        },
                        {
                            "terms": {
                                "meta.meta_key_3.raw": [
                                    "4"
                                ]
                            }
                        }
                    ]
                }
            }
        ]
    }
}
```

classes/class-ep-api.php
```diff
@@ -1728,6 +1494,294 @@ public function format_args( $args ) {
...
+	 * Build Elasticsearch filter query for WP meta_query
+	 *
+	 * @since 2.2
+	 *
+	 * @param $meta_queries
+	 *
+	 * @return array
+	 */
+	public function build_meta_query( $meta_queries ){
+		$meta_filter = array();
```
tlovett1: Why is this defined twice?

Ritesh-patel: Right, I have corrected it.

Issue #677, opened 2017-01-15
Page admin shows no pages when admin module activated.


teddarling:
I'm reporting this on a fresh install with 2017 theme and only ElasticPress installed.

When the admin module is deactivated, I see the list of pages on the site (See first image)

<img width="849" alt="screen shot 2017-01-15 at 3 17 39 pm" src="https://cloud.githubusercontent.com/assets/3044359/21966252/d2aa66dc-db35-11e6-8629-83fe68c9e75d.png">

Upon activating the admin module, I see that it reports 2 pages "All (2)" and "Published (2)" but "no pages found" listed as in the next screen shot:

<img width="840" alt="screen shot 2017-01-15 at 3 19 27 pm" src="https://cloud.githubusercontent.com/assets/3044359/21966268/2c8f3614-db36-11e6-8d5b-ea28bc1ab43a.png">

Expected behavior is to see the list of available pages for editing, viewing, etc... as in the first screenshot. I setup on a fresh install as I was noticing this on other sites that I was trying to install the plugin on.

If it helps, tested on Cloudways hosting with Varnish on and off, re-sync and still the same issue. PHP memory limit 256MB (tested originally at 32MB).

tlovett1:
What version of EP?

ekandreas:
Workaround is making a index command `wp elasticpress index`and pages are back. This has bothering me for some month now and occurs after a whole indexing with the setup flag sometimes.

teddarling:
@tlovett1 

plugin version: 2.1.2
WP: 4.7.1

@ekandreas 
I did do a run of `wp elasticpress index` with no change as well as deleting the index and trying again.

Host is Cloudways if that makes any difference (I've had other plugins break due to NGINX to Varnish to Apache). But again, I have tried with Varnish enabled and disabled.

Also if it helps, I've installed the EP Debug Bar which provides the following output (some data replaced for security reasons with xxxxx):


TOTAL ELASTICPRESS QUERIES:
1
TOTAL BLOCKING ELASTICPRESS QUERY TIME:
9 ms
Host: http://xxxxx:9200
Time Taken: 9 ms
URL: http://xxxxx:9200/wordpress2540054247195413cloudwaysappscom-1/post/_search
Method: POST
Query Args: 
array(67) {
  ["order"]=>
  string(3) "ASC"
  ["orderby"]=>
  string(16) "menu_order title"
  ["posts_per_page"]=>
  int(-1)
  ["posts_per_archive_page"]=>
[...]

PR #697, opened 2017-02-03
Fix hierarchical queries in WP4.6+

craigmillerdev:
Suggested fix for: https://github.com/10up/ElasticPress/issues/677

Issue is due to the early return when `fields => "id=>parent"` and never reaching the `get_posts` filter.
https://github.com/WordPress/WordPress/blob/4.7.2/wp-includes/class-wp-query.php#L2785

There doesn't appear to be any issue hooking the `posts_pre_query` (as long as it's available... 4.6.0+ only)

### DIFF
--- a/classes/class-ep-wp-query-integration.php
+++ b/classes/class-ep-wp-query-integration.php
@@ -47,8 +47,12 @@ class EP_WP_Query_Integration {
 		// Nukes the FOUND_ROWS() database query
 		add_filter( 'found_posts_query', array( $this, 'filter_found_posts_query' ), 5, 2 );
 
-		// Query and filter in EP_Posts to WP_Query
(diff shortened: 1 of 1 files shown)


tlovett1:
We are just going to pull admin page table query integration for now.

craigmillerdev:
Admin page was a nice thing to fix alongside this so I linked the PR to that issue - I was actually aiming for this: https://github.com/10up/ElasticPress/issues/724

Could use this for WP 4.6+ and for WP 3.7.1+ simply fallback to DB search when `'ids' == $q['fields']` or `'id=>parent' == $q['fields']`. Maybe not ideal for older WP versions, depends how complicated you want to make your plugin logic. If you're interested I can update the PR.

PR #703, opened 2017-02-08
make sure the version value exist in get_elasticsearch_version 

### LINKED ISSUE #702: Notice message is not correct when first install
Steps to reproduce:
1. Do fresh install
2. In plugin page it shows correct notice
3. In settings page it says "Your Elasticsearch version is below the minimum required Elasticsearch version 1.3. ElasticPress may or may not work properly."

I haven't set any host yet.

Trace down code:
class-ep-dashboard.php in function maybe_notice()

```
$es_version = ep_get_elasticsearch_version();

/**
 * Check Elasticsearch version compat
 */

if ( false !== $es_version ) {
	// First reduce version to major version i.e. 5.1 not 5.1.1
	$major_es_version = preg_replace( '#^([0-9]+\.[0-9]+).*#', '$1', $es_version );

	if ( -1 === version_compare( EP_ES_VERSION_MAX, $major_es_version ) ) {
		$notice = 'above-es-compat';
	} elseif ( 1 === version_compare( EP_ES_VERSION_MIN, $major_es_version ) ) {
		$notice = 'below-es-compat';
	}
}
```

$es_version will return null, thus $notice will set into 'below-es-compat'

ivankristianto:
If there is no host set, the get_elasticsearch_info will return false.
Thus we need to check the value of the array, otherwise it will return null
Fixed #702

### DIFF
--- a/classes/class-ep-api.php
+++ b/classes/class-ep-api.php
@@ -159,7 +159,7 @@ class EP_API {
 
 		$info = $this->get_elasticsearch_info();
 
-		return $info['version'];
+		return isset( $info['version'] )? $info['version'] : false;
 	}
 
 	/**


tlovett1:
Pushed a similar fix to this issue using a different approach.

Issue #552, opened 2016-08-16
EP used even with ep_skip_query_integration


rcespa:
Hi, I have a problem with a couple of plugins (FacetWP and Excel-Like Product Manager for WooCommerce).
Both the plugins stop working with EP activated.
The problem is that they run a standard query with the argument
`'field' => 'id'`

EP does its magic but always returns an array of WP_Post objects instead the expected array of post ids.

I tried with 

`add_filter('ep_skip_query_integration', '__return_true');`

but it still does not work.

Any suggestion?

Thanks in advance,
Roberto

tlovett1:
What version are you using?

rcespa:
EP: 2.1
WP: 4.5.3
WC: 2.6.4

tlovett1:
Can you try to trace the issue? `ep_skip_query_integration` should work

mdorchain:
I was testing EP with the same configuration (EP, Facetwp, woocommerce) and observed the same 2 issues about the array of posts returned when a an array of int is expected, but also the ep_skip_query_integration filter having no effect with the woocommerce module.

tlovett1:
`'field' => 'id'` won't work in EP due to `WP_Query` limitations.

jdmadison:
Sample filter implementation for the WooCommerce / StoreFront 404/featured products issue (Requires patch noted in Pull Request #705):

```php
function my_filter_ep_skip_query_integration( $skip, $query ) {

    $product = ( $query->get( 'post_type' ) === array( 'product', 'product_variation' ) );
    $featured = in_array( '_featured', array_column( $query->meta_query->queries, 'key' ) );

    return ( $skip || ( $product && $featured ) );

}

add_filter( 'ep_skip_query_integration', 'my_filter_ep_skip_query_integration', 10, 2 );
```

PR #705, opened 2017-02-08
Prevent EP from altering $query object when ep_skip_query_integration…

jdmadison:
… filter returns true. Fixes issue #552 

This issue was the source of thousands of `Object of class WP_Post could not be converted to string` PHP errors. A standard WooCommerce / StoreFront theme install includes a 404.php template which displays featured products in addition to the 'Page Not Found' error, which can be traced to `/wp-content/plugins/woocommerce/includes/wc-product-functions.php -> wc_get_featured_product_ids(202)`.

The `array_merge` call is expecting associative arrays, but is getting and array of WP_Post objects via the `$featured` variable when EP is enabled. This is due to `ep_wc_translate_args` changing the `fields` parameter from `id=>parent` to `default`. 

By making `ep_wc_translate_args` apply the `ep_skip_query_integration` filter, we can selectively disable this behavior.

### DIFF
--- a/features/woocommerce/woocommerce.php
+++ b/features/woocommerce/woocommerce.php
@@ -168,6 +168,10 @@ function ep_wc_translate_args( $query ) {
 		return;
 	}
 
+	if ( ! ep_elasticpress_enabled( $query ) || apply_filters( 'ep_skip_query_integration', false, $query ) ) {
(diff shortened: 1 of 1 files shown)


tlovett1:
Thanks!

PR #707, opened 2017-02-09
Add no-host notice in settings page when fresh install or no host setted

ivankristianto:
In host setting page, if there is no host set, we don't have a notice message.
This PR is just to add that messsage


tlovett1:
I don't understand what this does. When I have no host set, I get: `There is a problem with connecting to your Elasticsearch host. You will need to fix it for ElasticPress to work.`

Issue #708, opened 2017-02-09
Irrelevant search results


ivankristianto:
I'm using:
WP: 4.7.2
EP: 2.2
ES: 5.0.0
Content from theme-unit-test-data.xml

`wp elasticpress stats` result

```
====== Stats for: 10updev-1 ======
Documents:  184
Index Size: 929.10 KB
====== End Stats ======

```

**Keyword: "Rob Dyrdek"**

![search results for rob dyrdek 10up](https://cloud.githubusercontent.com/assets/656006/22786245/4addb0f4-ef0a-11e6-9713-db6128dcd06d.jpg)

**POST Query:**
```
{
	"explain": true,
    "from": 0,
    "size": 10,
    "sort": [
        {
            "_score": {
                "order": "desc"
            }
        }
    ],
    "query": {
        "function_score": {
            "query": {
                "bool": {
                    "should": [
                        {
                            "multi_match": {
                                "query": "Rob Dyrdek",
                                "type": "phrase",
                                "fields": [
                                    "post_title",
                                    "post_content",
                                    "post_excerpt",
                                    "author_name",
                                    "terms.post_tag.name",
                                    "terms.category.name",
                                    "post_author.login"
                                ],
                                "boost": 4
                            }
                        },
                        {
                            "multi_match": {
                                "query": "Rob Dyrdek",
                                "fields": [
                                    "post_title",
                                    "post_content",
                                    "post_excerpt",
                                    "author_name",
                                    "terms.post_tag.name",
                                    "terms.category.name",
                                    "post_author.login"
                                ],
                                "boost": 2,
                                "fuzziness": 0,
                                "operator": "and"
                            }
                        },
                        {
                            "multi_match": {
                                "fields": [
                                    "post_title",
                                    "post_content",
                                    "post_excerpt",
                                    "author_name",
                                    "terms.post_tag.name",
                                    "terms.category.name",
                                    "post_author.login"
                                ],
                                "query": "Rob Dyrdek",
                                "fuzziness": 1
                            }
                        }
                    ]
                }
            },
            "exp": {
                "post_date_gmt": {
                    "scale": "14d",
                    "decay": 0.25,
                    "offset": "7d"
                }
            }
        }
    },
    "post_filter": {
        "bool": {
            "must": [
                {
                    "terms": {
                        "post_type.raw": [
                            "post",
                            "page",
                            "attachment",
                            "product"
                        ]
                    }
                },
                {
                    "term": {
                        "post_status": "publish"
[...]

PR #720, opened 2017-02-14
Date Decay Fix – First Pass

mattonomics:
First pass at fixing the date decay issue. Currently, date decay is completely usurping relevancy scores and taking otherwise perfect matches to relevancy of zero because they are too far away from the decay origin. By adding score_mode of avg, a relevancy score is calculated using a weighted average which is preferable since we have large boost values for some parts of the query. We want to maintain that. By setting boost_mode to sum, the scores are added instead of the default multiply which seems to have been the primary driver of lowering relevancy scores to zero.

### DIFF
--- a/features/search/search.php
+++ b/features/search/search.php
@@ -164,6 +164,8 @@ function ep_weight_recent( $formatted_args, $args ) {
 						'offset' => apply_filters( 'epwr_offset', '7d', $formatted_args, $args ),
 					),
 				),
+				'score_mode' => 'avg',
+				'boost_mode' => 'sum'
 			),
 		);
 


mattonomics:
This is directly related to @ivankristianto's issue in #708. Will have to analyze scores, but the current date decay is pulling scores down unacceptably low.

Note the following default EP decay function:
```json
"exp": {
        "post_date_gmt": {
          "scale": "14d",
          "decay": 0.25,
          "offset": "7d"
        }
      }
```
The origin property is set to `now` by default. A `scale` of `14d` with a `decay` of `0.25` means that results 14 days from the origin will have a relevancy score of .25. Here's a visual for reference:
Elasticsearch date decay graph

The default scoring mode for decay is multiply, which means that perfectly relevant content (e.g. evergreen content) that is far out on the tail date-wise will have a decay score of `0`. So `0 * $high_relevancy_score = 0`, which is why we're seeing completely relevant content disappear to the bottom of results.

mattonomics:
Alternatively, date decay functionality could (and arguably _should_) be put into a module, rather than default functionality. While it may make sense for _some_, I don't believe that it makes sense for _all_.

tlovett1:
This is awesome. Thanks @mattonomics !

mattonomics:
Thank you, @tlovett1!

Quick word of caution if using EP on some sites: this will change the results ordering, likely drastically. In the case of an organization who may not want to surface older, though highly relevant content, their query may need a bit of tweaking to adjust for the shift. I recommend extra scrutiny before pushing this change to production on a site that values recency over relevancy.

That said, I surmise that the vast majority of sites will prefer the updated query since it will only boost more relevant content, given two otherwise similar relevancy scores.

rveitch:
I agree with @mattonomics that date decay functionality would make most sense in a module or filter as it would not apply to all use cases.  I add guassian scored date decay using a filter, which makes it very easy to tweak and activate/deactivate as needed.

In regards to changes in result ordering, I typically include a toggle to choose between "Date" and "Relevance" (or "Newest, "Relevance", and "Oldest") with "Date" being the default. Entirely optional, but can help skirt the issue of recency over relevancy.

image image

kevinlisota:
@rveitch Any chance you have a gist available for those toggles. Am looking to do exactly that on our site.

rveitch:
@kevinlisota  Sorry, I don't have any WordPress integrated examples handy. I've been working primarily with decoupled front ends (Vue/React/Ember) lately and would have to dig up safe references from past client work, but I may be able to point you in the right direction. 

The ElasticPress filter `ep_formatted_args` is a good place to add to, update or change the generated and formatted query before it is sent by WordPress/ElasticPress to your Elasticsearch cluster. A typical WordPress php theme search will redirect to a route matching the theme's search template with the search term in a query string parameter with a key of `s` and a value of the search term. 

- Example: `http://areavoices.com/?s=jupiter`

I would pass the value of the 'order-by' select form as a query string parameter.

- Example: `http://areavoices.com/?s=jupiter&orderby=relevance`

It's up to you how you pass the parameter value along, but ultimately you want to use it with a filter like `ep_formatted_args` to update the query accordingly before it is sent out.

Sorry I don't have any specific code examples to share at the moment, but hope that helps some!

PR #730, opened 2017-02-23
Add WP_Query $query_vars / EP API format_args $args filter

### LINKED ISSUE #728: Add WP_Query $query_vars / EP API format_args $args filter
For _some plugin integrations_, it is necessary to create customized WP_Query query vars for ElasticPress to make use of versus keeping in place certain SQL overrides being hooked into WP_Query. This ensures a more natural experience for integrating plugins with EP that require a more complicated WP_Query SQL integration (because,.. MySQL).

In order to customize the WP_Query query vars used, a filter needs to be added just before the `ep_wp_query_search_cached_posts` filter call.

Something like this:

```php
/**
 * Filter WP_Query query vars specifically for ElasticPress to use
 *
 * @since x.x.x
 *
 * @param array     $query_vars WP_Query query vars to use
 * @param \WP_Query $query WP_Query object
 */
$query_vars = apply_filters( 'ep_wp_query_vars', $query_vars, $query );
```

Alternatively (or in addition to the above filter), something within EP_API::format_args() at the top of the method would also acceptable:

```php
/**
 * Filter formatted arguments specifically for ElasticPress to use
 *
 * @since x.x.x
 *
 * @param array $args Arguments to format
 */
$args = apply_filters( 'ep_pre_formatted_args', $args );
```

For the interim, the only thing you can do is add a hook to `ep_skip_query_integration` not to change the value returned but instead to interact with the `$query->query_vars`.

sc0ttkclark:
For some plugin integrations, it is necessary to create customized WP_Query query vars for ElasticPress to make use of versus keeping in place certain SQL overrides being hooked into WP_Query. This ensures a more natural experience for integrating plugins with EP that require a more complicated WP_Query SQL integration (because,.. MySQL).

Fixes #728


classes/class-ep-wp-query-integration.php
```diff
@@ -234,6 +234,16 @@ public function filter_posts_request( $request, $query ) {
...
 
+		/**
+		 * Filter WP_Query query vars specifically for ElasticPress to use
+		 *
+		 * @since x.x.x
+		 *
+		 * @param array     $query_vars WP_Query->query_vars to use
+		 * @param \WP_Query $query      WP_Query object
+		 */
+		$query_vars = apply_filters( 'ep_wp_query_vars', $query_vars, $query );
```
tlovett1: Why not run this filter at the beginning of the method?

sc0ttkclark: We want EP to do the normal $query_var stuff it does with post_type up above. Is there a reason to not take advantage of the logic built there? post_type has a hook of it's own, checking if EP is enabled or not for the query has it's own handling. This makes sense right here before caching occurs so we can take advantage of the caching part.

sc0ttkclark:
Any further thoughts or problems with this PR?

sc0ttkclark:
I've accomplished what I needed to do without these filters. I am okay with closing this, but open to feedback on whether the hooks would still be handy. Up to you, I'll let your team make that decision.

Issue #735, opened 2017-02-27
Fatal Error when indexing a post with oEmbeds


ivankristianto:
```
[27-Feb-2017 14:21:49 UTC] PHP Fatal error:  Maximum function nesting level of '100' reached, aborting! in /srv/www/10up/htdocs/wp-includes/kses.php on line 1491
[27-Feb-2017 14:21:49 UTC] PHP Stack trace:
[27-Feb-2017 14:21:49 UTC] PHP   1. {main}() /srv/www/10up/htdocs/wp-admin/admin-ajax.php:0
[27-Feb-2017 14:21:49 UTC] PHP   2. do_action($tag = 'wp_ajax_ep_index', $arg = *uninitialized*) /srv/www/10up/htdocs/wp-admin/admin-ajax.php:91
[27-Feb-2017 14:21:49 UTC] PHP   3. WP_Hook->do_action($args = array (0 => '')) /srv/www/10up/htdocs/wp-includes/plugin.php:453
[27-Feb-2017 14:21:49 UTC] PHP   4. WP_Hook->apply_filters($value = '', $args = array (0 => '')) /srv/www/10up/htdocs/wp-includes/class-wp-hook.php:323
[27-Feb-2017 14:21:49 UTC] PHP   5. EP_Dashboard->action_wp_ajax_ep_index('') /srv/www/10up/htdocs/wp-includes/class-wp-hook.php:298
[27-Feb-2017 14:21:49 UTC] PHP   6. ep_prepare_post($post_id = 3802) /srv/www/10up/htdocs/wp-content/plugins/ElasticPress/classes/class-ep-dashboard.php:493
[27-Feb-2017 14:21:49 UTC] PHP   7. EP_API->prepare_post($post_id = 3802) /srv/www/10up/htdocs/wp-content/plugins/ElasticPress/classes/class-ep-api.php:2411
[27-Feb-2017 14:21:49 UTC] PHP   8. apply_filters($tag = 'the_content', $value = 'https://10up.com/blog/2017/elasticpress-hosted-search/') /srv/www/10up/htdocs/wp-content/plugins/ElasticPress/classes/class-ep-api.php:579
[...]

PR #741, opened 2017-02-28
disable oembed while indexing which cause fatal error

ivankristianto:
Fixed for #735


tlovett1:
Nice!

tlovett1:
@ivankristianto I force reverted this. Issues:
* Prepare post is called many times. This filter will get added over and over again.
* The filter should be removed after it is no longer needed.

ivankristianto:
Ah okay. will more careful next time.

PR #748, opened 2017-03-02
make sure the orderby is correct for default sorting

### LINKED ISSUE #745: Woocommerce option ruins sorting
I use product sort on homepage by popularity (sales)
When I turn on woocommerce option in elasticpress all products sorting by name default.
If I try to go open %SHOPHOMEPAGE%/?orderby=popularity everything's OK.

ivankristianto:
The source of the problem is from the WC_Query:
https://github.com/woocommerce/woocommerce/blob/master/includes/class-wc-query.php#L453-L483

if there is no querystring orderby then it fall back to the default settings.
But it set the orderby to menu_order title in line https://github.com/woocommerce/woocommerce/blob/master/includes/class-wc-query.php#L449

We can only check from the meta_key, and translate that into correct orderby

Fixed #745

### DIFF
--- a/features/woocommerce/woocommerce.php
+++ b/features/woocommerce/woocommerce.php
@@ -405,6 +405,26 @@ function ep_wc_translate_args( $query ) {
 		}
 
 		/**
+		 * For default sorting by popularity (total_sales) and rating
+         * Woocommerce doesn't set the orderby correctly.
+         * These lines will check the meta_key and correct the orderby based on that.
+         * And this won't run in search result and only run in main query
+		 */
+		$meta_key = $query->get( 'meta_key', false );
+		if ( $meta_key && empty( $s ) && $query->is_main_query() ){
+			switch ( $meta_key ){
+				case 'total_sales':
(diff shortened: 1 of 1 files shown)


features/woocommerce/woocommerce.php
```diff
@@ -405,6 +405,26 @@ function ep_wc_translate_args( $query ) {
 		}
 
 		/**
+		 * For default sorting by popularity (total_sales) and rating
+         * Woocommerce doesn't set the orderby correctly.
+         * These lines will check the meta_key and correct the orderby based on that.
+         * And this won't run in search result and only run in main query
+		 */
+		$meta_key = $query->get( 'meta_key', false );
+		if ( $meta_key && empty( $s ) && $query->is_main_query() ){
```
tlovett1: I'm not sure we need this. We not just set order to `desc` when ordering by popularity?

ivankristianto: This line is to handle the default product sorting, not in search query.
You can set default sorting in WooCommerce > Settings > Products > Display
And for the popularity and average_rating, woocommerce doesn't set the correct orderby and order for wp_query.
These lines is to handle those case.

PR #755, opened 2017-03-05
make sure oembed won't mess with the indexing anymore

ivankristianto:
Make sure other oembed like instagram, twitter, youtube, etc... won't cause the problem when indexing


tlovett1:
Was there another error happening? I thought we fixed this?

ivankristianto:
@tlovett1 yes I found it yesterday.
Where Instagram url embed will cause the indexing with infinite loop again.
I disable all the oembed filter from the the_content to make sure it doesn't happen again

tlovett1:
Does removing the_content fix this?

ivankristianto:
Yes, if we merge this #764 
then this PR is not necessary

PR #756, opened 2017-03-05
woocommerce related product use elasticpress search

### LINKED ISSUE #736: Add ElasticPress Related Products Widget when Woocommerce feature activated
We have "ElasticPress Related Posts" widget, it would be great if we add Related Products when woocommerce feature activated.

ivankristianto:
Change query for related product in woocommerce related product. Fixed #736


ivankristianto:
Close this PR, for WC > 3, it's impossible to extend their `wc_get_related_products` to get the related product from ES

PR #758, opened 2017-03-08
Add filter for max results window size

kallehauge:
Added a more detailed explanation of what is going on with the `posts_per_page` conversion and added the option to alter the max results window size hardcoded in EP.

You might be wondering: "Hey, why do we need this?" – well, it is basically documentation with the added functionality of increasing/decreasing the size so you can optimise it for your own server (some might have cheaper ES instances where the window size have been reduced etc.).

### DIFF
--- a/classes/class-ep-api.php
+++ b/classes/class-ep-api.php
@@ -969,7 +969,7 @@ class EP_API {
 				 *
 				 * @return int The max results window size.
 				 *
-				 * @since 2.2.1
+				 * @since 2.3.0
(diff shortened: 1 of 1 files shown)


tlovett1:
Awesome. We'll add this in 2.3.

kallehauge:
Brilliant @tlovett1 :+1:

I can see that the tests that fail also fail in ex #763 so I assume that they are not related. /poke me if this is not the case.

PR #763, opened 2017-03-09
Index Prefixing Constant

allan23:
In order to leverage the full capabilities of wildcard permissions offered by X-Pack, it is required to be able to set a unique index prefix that is applied to all indexes.

This PR allows developers to define a constant in wp-config.php that will add a prefix to their index and alias names without the need to use filters.

`define( 'EP_INDEX_PREFIX', 'abc-' );`

### DIFF
--- a/classes/class-ep-config.php
+++ b/classes/class-ep-config.php
@@ -125,6 +125,10 @@ class EP_Config {
 			$index_name = false;
 		}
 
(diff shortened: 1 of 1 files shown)


tlovett1:
We should probably check if the constant is not empty vs. just checking if it exists

Issue #714, opened 2017-02-10
Indexing using wpcli prematurly terminated


ivankristianto:
I have 950 posts.

I run:
```
vagrant@vvv:/srv/www/10up/htdocs$ wp elasticpress index --setup
Adding mapping...
Deleting index...
Success: Index deleted
Success: Mapping sent
Indexing posts...
Request string length: 1.65 MB
Processed 350/950 entries. . .
Request string length: 1.48 MB
Processed 700/950 entries. . .
vagrant@vvv:/srv/www/10up/htdocs$
```

The `wp elasticpress stats` return:
====== Stats for: 10updev-1 ======
Documents:  700
Index Size: 3.60 MB
====== End Stats ======

Environment:
WP 4.7.2
EP 2.2
ES 5.0.0

ivankristianto:
Stack trace:

```
PHP Fatal error:  Uncaught Error: Call to a member function get_permalink() on boolean in /srv/www/10up/htdocs/wp-content/plugins/woocommerce/includes/class-wc-post-data.php:55
Stack trace:
#0 /srv/www/10up/htdocs/wp-includes/class-wp-hook.php(300): WC_Post_Data::variation_post_link('http://10up.dev...', Object(WP_Post))
#1 /srv/www/10up/htdocs/wp-includes/plugin.php(203): WP_Hook->apply_filters('http://10up.dev...', Array)
#2 /srv/www/10up/htdocs/wp-includes/link-template.php(298): apply_filters('post_type_link', 'http://10up.dev...', Object(WP_Post), false, false)
#3 /srv/www/10up/htdocs/wp-includes/link-template.php(148): get_post_permalink(Object(WP_Post), false, false)
#4 /srv/www/10up/htdocs/wp-content/plugins/ElasticPress/classes/class-ep-api.php(585): get_permalink(Object(WP_Post))
#5 /srv/www/10up/htdocs/wp-content/plugins/ElasticPress/classes/class-ep-api.php(2408): EP_API->prepare_post(42)
#6 /srv/www/10up/htdocs/wp-content/plugins/ElasticPress/bin/wp-cli.php(576): ep_prepare_post(42)
#7 /srv/www/10up/htdocs/wp-cont in /srv/www/10up/htdocs/wp-content/plugins/woocommerce/includes/class-wc-post-data.php on line 55
```

related: #694

ivankristianto:
Just to add another information:
this only happens when it has pagination or posts > 350,
[...]

PR #764, opened 2017-03-13
change post content filter to just do_shortcode

ivankristianto:
Do `apply_filter` on `the_content` when doing indexing might cause an infinite loop.
Someone might use apply_filter the_content to parse anything and try to update_post_meta, this will cause infinite loop doing indexing. I've experience this a lot.
For example oembeds.

This PR might rise some concern and need discussion.
cc: @tott and @tlovett1

### DIFF
--- a/classes/class-ep-api.php
+++ b/classes/class-ep-api.php
@@ -568,9 +568,6 @@ class EP_API {
 			}
 		}
 
-		// Turn off oEmbed auto discovery as this will create an error while indexing
-		add_filter( 'embed_oembed_discover', '__return_false' );
-
 		$post_args = array(
 			'post_id'           => $post_id,
 			'ID'                => $post_id,
@@ -579,7 +576,7 @@ class EP_API {
 			'post_date_gmt'     => $post_date_gmt,
 			'post_title'        => $this->prepare_text_content( get_the_title( $post_id ) ),
 			'post_excerpt'      => $this->prepare_text_content( $post->post_excerpt ),
-			'post_content'      => $this->prepare_text_content( apply_filters( 'the_content', $post->post_content ) ),
+			'post_content'      => $this->prepare_text_content( do_shortcode( $post->post_content ) ) ,
 			'post_status'       => $post->post_status,
 			'post_name'         => $post->post_name,
 			'post_modified'     => $post_modified,
@@ -607,9 +604,6 @@ class EP_API {
 		$post_args['meta'] = $this->prepare_meta_types( $post_args['post_meta'] );
 
 		$post_args = apply_filters( 'ep_post_sync_args_post_prepare_meta', $post_args, $post_id );
-
-		// Turn back on oEmbed discovery
-		remove_filter( 'embed_oembed_discover', '__return_false' );
 
 		return $post_args;
 	}


tlovett1:
@allan23 have you looked at this?

allan23:
@ivankristianto can you let me know how to reproduce the infinite loop issue?

ivankristianto:
@allan23 currently what i know cause the infinite loop is wp-oembed.
Sample:
https://10up.com/blog/2017/six-year-anniversary/
and
https://www.instagram.com/p/57fx3dsB2I/

What cause the infinite loop is:
When WordPress render the wp-oembed, it will update postmeta, and when postmeta is running, it will trigger another elasticpress sync.

see this issue: #714

tlovett1:
@allan23 ???

allan23:
@tlovett1 - while my initial testing of this resolves the oembed issue, I cannot confirm yet that this won't mess with some implementations. Can you think of a use-case where apply_filters() is actually needed on indexing?

ivankristianto:
@allan23 here is my setup:

WP 4.7.3
ES 2.4.0
EP 2.2.1 ( from branch develop )
WC 2.6.14

Here is my exported file:
https://drive.google.com/a/get10up.com/file/d/0B1WrBwInJmWiRlBmdkZxYkR6UGc/view?usp=sharing

This is my debug.log:
https://drive.google.com/a/get10up.com/file/d/0B1WrBwInJmWiZWlvTEJOMDkzZlE/view?usp=sharing

Here is to reproduce:
1. If you are not imported my exported file. This is how i setup it:
Create a post and put this in the content: https://10up.com/blog/2017/six-year-anniversary/
Create another post and put this in the content: https://www.instagram.com/p/57fx3dsB2I/
Both are wp-embed link and you need to wait for some times until it is resolved.

2. Activate Elasitcpress & Woocommerce plugin ( don't know if WC is related, but I just turn it one for this setup )

3. Configure EP and Enable All modules

4. Run Sync either from dashboard or from WP-CLI


------------
To be discussed:
There are 2 options I propose:
1. Fullly disable wp-oembed from the apply_filter the_content, here is the PR for that
https://github.com/10up/ElasticPress/pull/755

2. Remove apply_filter the_content from indexing, and just use do_shortcode. That would be this PR. But We need to know the implication of this.

drebbits:
It's happening on my local as well. Here's my current setup:
WP 4.7.3
ES 2.3.3
EP 2.2.1 (up to date with the `develop` branch)

The process stuck when it tries to render the embed as @ivankristianto describes. I tested his patch and works well on my end.

ivankristianto:
@tlovett1 / @allan23 
Me and @drebbits have a discussion for this issue and have a solution that might work without removing the apply_filter the_content.

The infinite loop is happen when in apply_filter the_content is updating a post_meta, which will trigger another ep_prepare_post in this function: https://github.com/10up/ElasticPress/blob/develop/classes/class-ep-sync-manager.php#L94

We can disable that function during the dashboard sync or wpcli syng, because we are having the transient data `ep_wpcli_sync` or `ep_dashboard_sync`.

in that case the ep_prepare_post will stop running and it would prevent the infinite loop.

tlovett1:
@allan23 did you ever poll our team and investigate this?

allan23:
@tlovett1 per the responses from the team, the overall consensus appears that this PR would not break anyone's implementation. I don't see any issue with moving this forward because there are options for filtering content if it is absolutely needed.  What do you think?

TheLastCicada:
We ran into this issue on a site just yesterday, so an official solution would be very welcome!

ivankristianto:
@tlovett1 and @allan23 
I did push a PR for infinite loop fixed, without remove the filter the_content.
Please check this out: #798

PR #765, opened 2017-03-13
fix feature auto-activate when the feature is manually disabled

### LINKED ISSUE #743: Feature auto-activates even when disabled in the past
Title says it all.

ivankristianto:
Fixed #743 

Add manual_deactivate flag in options to store the preference of the feature.
And it will prevent the feature to auto activated when it is disabled before


tlovett1:
Fixed in a different PR.

PR #769, opened 2017-03-16
Add support for custom taxonomies in WooCommerce

kallehauge:
It might be useful to automate the product taxonomies but, for now, I would love to see support for "custom taxonomies" for the WooCommerce feature.

The reason is that custom taxonomies added to "products" by plugins are not sortable in the admin view if they are not added to the `$supported_taxonomies` array.


tlovett1:
Solid.

PR #770, opened 2017-03-16
Add default support for "product_type" (Woo)

kallehauge:
It is currently not possible to filter by product type in the backend. This PR should solve that by adding `product_type` taxonomy support by default.


tlovett1:
Awesome!

PR #778, opened 2017-03-21
Added configuration constant for ElasticSearch index name

RichElse:
Allows users to define the name for the ES index rather than using siteurl and blog id and/or prefix. 

Makes index naming more readable and brief, and great if using a single WP instance and an ES cluster with many indices.

### DIFF
--- a/classes/class-ep-config.php
+++ b/classes/class-ep-config.php
@@ -118,9 +118,11 @@ class EP_Config {
 
 		$site_url = get_site_url( $blog_id );
 
-		if ( ! empty( $site_url ) ) {
+		if ( ! empty( $site_url ) && ! defined( 'EP_INDEX_NAME') ) ) {
 			$index_name = preg_replace( '#https?://(www\.)?#i', '', $site_url );
 			$index_name = preg_replace( '#[^\w]#', '', $index_name ) . '-' . $blog_id;
+		} elseif(EP_INDEX_NAME) {
+			$index_name = EP_INDEX_NAME;
 		} else {
 			$index_name = false;
 		}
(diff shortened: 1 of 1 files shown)


tlovett1:
This would be problematic for multisite since each site needs a unique index.

ivankristianto:
I think Allan made a PR same think like this one: #763

RichElse:
I thought it might be ommitted because of multisite. How about asking the question on set-up? Is the more usual case a single instance or multisite?

RichElse:
Yes, prefix is good in #763 (for Kibana usage too) but would be good to have less verbose and more descriptive index names where EP index part of cluster with other indices. Makes for more readable ES queries when querying across indices e.g. searching both WP content and other data

tlovett1:
@RichElse since the prefix PR has been merged. Do you still need this? There is a filter on `ep_get_index_name` - https://github.com/10up/ElasticPress/blob/develop/classes/class-ep-config.php#L132

RichElse:
@tlovett1 it would be really handy for our use case.

We use several single instances of WP as content management only with content delivery apps using WP API and ElasticSearch. The pre-fix is good - as it means we can search indices with a wild card - however, when querying specific indices,  [prefix]-[site url]_[blog id] will make readability difficult, especially as we are combining WP content with content from other sources.

I'm happy to make it more elegant if required but wouldn't want to over complicate what is a very usable and simple plug-in installation process with additional fields on set-up screen. Also, pho isn't my first language (as you might have noticed from commits!) so any suggestions amendments welcome! :)

tlovett1:
@RichElse just use code like this:

```php
add_filter( 'ep_get_index_name', function() {
   return 'new_index_name';
} );

PR #780, opened 2017-03-22
Disable dashboard sync with constant. Closes #701

### LINKED ISSUE #701: Disable dashboard sync with constant
This would be helpful for sites we manage where clicking the dashboard sync problem would cause major problems (deleting the index).

rveitch:
## Disable Dashboard Sync

Dashboard sync can be disabled by defining the constant `EP_DASHBOARD_SYNC` as `false` in your wp-config.php file.

```php
define( 'EP_DASHBOARD_SYNC', false );
```

* This can be helpful for managed sites where users initiating a sync from the dashboard could potentially cause issues such as deleting the index and limiting this control to WP-CLI is preferred.
* When disabled, features that would require reindexing are also prevented from being enabled/disabled from the dashboard.

## Additional Details
When `EP_DASHBOARD_SYNC` is defined as `false`:
* Attempting to manually trigger a dashboard sync via `admin.php?page=elasticpress&do_sync` will not initiate a sync. This is to prevent any tomfoolery, intentional or otherwise.
* The `resume`, `pause` and `stop` buttons are not disabled by this constant. This is to prevent situations where a dashboard sync my be in progress (or paused) when sync is disabled and will allow the in-progress sync to be managed.  If no sync is in-progress, these buttons will not appear.
* Dashboard notifications for `no-sync`, `upgrade-sync` and `auto-activate-sync` will be replaced with new dismissible notifications tailored to their respective conditions.
  * These notifications are unlikely to display on the dashboard as most use cases for disabling dashboard sync should (in theory) already be handling their respective situations from WP-CLI and their trigger conditions should already be cleared, but I thought it was safer to include them for situations such as someone updating enabling a feature but forgetting to reindex.

### New Notifications
`sync-disabled-no-sync`
image

`sync-disabled-upgrade`
screenshot5028

`sync-disabled-auto-activate`
image

### DIFF
--- a/classes/class-ep-dashboard.php
+++ b/classes/class-ep-dashboard.php
@@ -191,6 +191,16 @@ class EP_Dashboard {
 			if ( false === $last_sync && ! isset( $_GET['do_sync'] ) ) {
 				$notice = 'no-sync';
 			}
+
+			if ( defined( 'EP_DASHBOARD_SYNC' ) && ! EP_DASHBOARD_SYNC ) {
+				if ( 'auto-activate-sync' == $notice ) {
+					$notice = 'sync-disabled-auto-activate';
+				} elseif ( 'upgrade-sync' == $notice ) {
+					$notice = 'sync-disabled-upgrade';
+				} elseif ( 'no-sync' == $notice ) {
+					$notice = 'sync-disabled-no-sync';
+				}
+			}
 		}
 
 		$es_version = ep_get_elasticsearch_version( $force );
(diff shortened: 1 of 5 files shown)


rveitch:
I'm fairly certain that the failed checks will be due to changes in the notification strings, and I believe the tests would need to be updated to reflect new conditions introduced here.

classes/class-ep-feature.php
```diff
@@ -162,6 +162,12 @@ public function requirements_status() {
 			$status = call_user_func( $this->requirements_status_cb, $status, $this );
 		}
 
+		if ( true === $this->requires_install_reindex && defined( 'EP_DASHBOARD_SYNC' ) && ! EP_DASHBOARD_SYNC ) {
+      $status->code = 2;
+      $status->message = "Dashboard sync is disabled.
```
tlovett1: Need to use proper localization functions here.

classes/class-ep-dashboard.php
```diff
@@ -660,7 +716,7 @@ public function action_admin_enqueue_dashboard_scripts() {
 					$index_meta['wpcli_sync'] = true;
 				}
 
-				if ( isset( $_GET['do_sync'] ) ) {
+				if ( isset( $_GET['do_sync'] ) && defined( 'EP_DASHBOARD_SYNC' ) && EP_DASHBOARD_SYNC ) {
```
tlovett1: This conditional is wrong.

classes/class-ep-feature.php
tlovett1: Oh nevermind.

tlovett1:
Very nice. Just one small tweak needed.

classes/class-ep-dashboard.php
rveitch: Were you thinking of more strict type checking on the truthiness of EP_DASHBOARD_SYNC? 

Something like this:
```php
if ( isset( $_GET['do_sync'] ) && defined( 'EP_DASHBOARD_SYNC' ) && true === EP_DASHBOARD_SYNC ) {
```

tlovett1:
Can you fix the remaining issue?

PR #782, opened 2017-03-23
Adds read-only option

psorensen:
Wanted to get this submitted for user testing. **Unit tests on the way!** Will update readme after any changes post testing/feedback.

This PR adds an option to the settings page to enable read only mode. Much like the ep_host setting, this takes into account whether the environmental variable is set (EP_READ_ONLY)

Would love some user testing feedback on this.

screen shot 2017-03-23 at 12 09 31 pm
screen shot 2017-03-23 at 12 08 47 pm


psorensen:
This should probably be better integrated with #780

psorensen:
This is still very much a work in progress. As it stands, EP still tries to connect to the index name given to the local install resulting in a 404. Proposed roadmap:

- [ ] Add input to declare remote index name
- [ ] Filter the index name with the value from above https://github.com/10up/ElasticPress/blob/49df14f5dc01df466e881b81645db617745195cd/classes/class-ep-config.php#L132

tlovett1:
Hey @psorensen, can you run me through the use case for this? We added the ability to disable dashboard syncing in the last version.

Thanks!

psorensen:
@tlovett1 Sure. My intention for this was to disable all syncing (GUI and wp-cli) and filter the EP index name value to reflect a remote ES server. This way, a development environment can query against a live production ES server without the fear of altering it. 

Read-only mode can be set via either a constant (EP_READ_ONLY) or a site option (ep_read_only). When the constant is defined true, the GUI checkbox is disabled in the on position. (see second screenshot above)

Benefits include:
- Realtime data to query against
- no need to set up/sync local ES

Potential issues:
- performing memory intensive tasks may affect live ES server

This is only about 75% complete--feeling the room as to whether this is a worthwhile addition to EP. The filters already in place make it easy enough to do this on their own.

tlovett1:
@psorensen I think this could be useful but don't want us to duplicate functionality. We already have read more functionality for the dashboard; all that's left is read only for WP-CLI and disabling syncing on post update.

tlovett1:
Closing for now. Feel free to reopen.

PR #794, opened 2017-04-03
Fix/wpcli nobulk

### LINKED ISSUE #792: When using --nobulk parameter, the indexing is success, but the return message is 0
Using --nobulk parameter in WP-CLI indexing.
The indexing is success but the message return 0 index:
`Number of posts indexed on site 1: 0`

ivankristianto:
Fixed #792


allan23:
@ivankristianto This PR fixes the issue in #792.

(CC: @tlovett1 )

PR #801, opened 2017-04-17
Add support for comma separated shorthand taxonomies

kallehauge:
The current implementation of handling "supported taxonomies" assumes that the `query_var` argument is a single value (example: `'product_brand' => 'adidas'`) - It does not support multiple terms separated by a comma (example: `'product_brand' => 'adidas,nike'`).

### DIFF
--- a/features/woocommerce/woocommerce.php
+++ b/features/woocommerce/woocommerce.php
@@ -247,7 +247,14 @@ function ep_wc_translate_args( $query ) {
 		if ( ! empty( $term ) ) {
 			$integrate = true;
 
-			$terms = explode( ',', $term );
+			// Shorthand taxonomies can be one of two data types: array and
+			// string. So we assume that if the provided value is not a string,
+			// then it must be an array.
+			// If it is a string then we convert it to an array with one or more
+			// values to handle it with the same logic.
+			if ( is_string( $term ) ) {
+				$terms = explode( ',', $term );
+			}
 
 			// to add child terms to the tax query
 			if ( is_taxonomy_hierarchical( $taxonomy ) ) {
--- a/features/woocommerce/woocommerce.php
+++ b/features/woocommerce/woocommerce.php
@@ -247,19 +247,20 @@ function ep_wc_translate_args( $query ) {
 		if ( ! empty( $term ) ) {
 			$integrate = true;
 
-			$terms = array( $term );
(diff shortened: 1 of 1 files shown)


features/woocommerce/woocommerce.php
```diff
@@ -247,19 +247,20 @@ function ep_wc_translate_args( $query ) {
 		if ( ! empty( $term ) ) {
 			$integrate = true;
 
-			$terms = array( $term );
+			$terms = explode( ',', $term );
```
tlovett1: What if `term` is an array? Seems like we need to check that case too.

tlovett1: See this?

kallehauge: Ah, sorry @tlovett1 - I cannot remember why I did not check if it was an array. I guess it's because the existing code assumes that it is a string.
But you're correct, there should be a check for "array".

I correct in in the near future :+1:

kallehauge:
Sorry this took so long @tlovett1 - please take a look at it again

features/woocommerce/woocommerce.php
```diff
@@ -247,19 +247,27 @@ function ep_wc_translate_args( $query ) {
...
 			$integrate = true;
 
-			$terms = array( $term );
+			// Shorthand taxonomies can be one of two data types: array and
+			// string. So we assume that if the provided value is not a string,
+			// then it must be an array.
+			// If it is a string then we convert it to an array with one or more
+			// values to handle it with the same logic.
+			if ( is_string( $term ) ) {
+				$terms = explode( ',', $term );
```
tlovett1: Seems like this only gets defined if it's a string

kallehauge:
Brain fart time – great catch @tlovett1. It should be ready for merge, this time ☺️

kallehauge:
@lukaspawlik can you confirm that this have been fixed with your work in: https://github.com/10up/ElasticPress/commit/41f9f18106245be18bb241abaadeda57a3c6dea6 ? ☺️

kallehauge:
@tlovett1 @lukaspawlik I've finally updated some of our projects to the latest version of ElasticPress and discovered that this is still not covered. It's the same case as above.
I might make a patch in the following days - if not, then this just serves as a reminder :-)

allan23:

PR #803, opened 2017-04-18
Fix/woocommerce search order

### LINKED ISSUE #802: Search by phone number in Woocommerce Order Admin
the title says it all.
I open this issue for tracking purpose.

ivankristianto:
WooCommerce as default cannot search by phone number, billing email and billing name.
This PR address that issue. Fixed #802

### DIFF
--- a/features/woocommerce/woocommerce.php
+++ b/features/woocommerce/woocommerce.php
@@ -546,6 +546,10 @@ function ep_wc_bypass_order_permissions_check( $override, $post_id ) {
 
 /**
  * Enhance WooCommerce search order by order id, email, phone number, name, etc..
+ * What this function does:
+ * 1. Reverse the woocommerce shop_order_search_custom_fields query
+ * 2. If the search key is integer and it is an Order Id, just query with post__in
+ * 3. If the search key is integer but not an order id ( might be phone number ), use ES to find it
  *
  * @param $wp
  */
@@ -553,7 +557,7 @@ function ep_wc_search_order( $wp ){
 	global $pagenow;
 
 	if ( 'edit.php' != $pagenow || 'shop_order' !== $wp->query_vars['post_type'] ||
-	     ! isset( $wp->query_vars['shop_order_search'] ) || true !== $wp->query_vars['shop_order_search'] ) {
+	     ( empty( $wp->query_vars['s'] ) && empty( $wp->query_vars['shop_order_search'] ) ) ) {
 		return;
 	}
 
@@ -561,10 +565,15 @@ function ep_wc_search_order( $wp ){
 	$order    = wc_get_order( absint( $search_key_safe ) );
 
 	//If the order doesn't exist, fallback to other fields
-	if( ! $order ){
+	if ( ! $order ) {
 		unset( $wp->query_vars['shop_order_search'] );
 		unset( $wp->query_vars['post__in'] );
 		$wp->query_vars['s'] = $search_key_safe;
+	} else {
+		//we found the order. don't query ES
+		unset( $wp->query_vars['s'] );
+		$wp->query_vars['post__in'] = array( $order->get_id() );
+		$wp->query_vars['ep_integrate'] = true;
 	}
 }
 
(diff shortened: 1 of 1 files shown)


features/woocommerce/woocommerce.php
```diff
@@ -544,6 +544,30 @@ function ep_wc_bypass_order_permissions_check( $override, $post_id ) {
...
+ */
+function ep_wc_search_order( $wp ){
+	global $pagenow;
+
+	if ( 'edit.php' != $pagenow || 'shop_order' !== $wp->query_vars['post_type'] ||
+	     ! isset( $wp->query_vars['shop_order_search'] ) || true !== $wp->query_vars['shop_order_search'] ) {
+		return;
+	}
+
+	$search_key_safe = str_replace( array( 'Order #', '#' ), '', wc_clean( $_GET['s'] ) );
```
tlovett1: I'm not understanding this

LiamMcArthur:
I still can't search by order. When attempting search for a customer search name of McCabe, the URL changes to `wp-admin/edit.php?s=McCabe&post_status=all&post_type=shop_order&action=-1&m=0&_customer_user&paged=1&action2=-1` The page refreshes, and shows `Search results for “”`.

features/woocommerce/woocommerce.php
ivankristianto: It's meant for cleaning user input in Order Admin dashboard.
sometimes they put: 'Order #123456' or '#123456' since the order admin area has the prefix '#'.
screenshot: https://www.dropbox.com/s/iygvo0j5ri1sle1/Screenshot%202017-05-02%2018.51.33.png?dl=0
WooCommerce also does the same: https://github.com/woocommerce/woocommerce/blob/master/includes/wc-order-functions.php#L745

tlovett1:
I don't see how this PR addresses phone number, billing email or billing name.

ivankristianto:
@tlovett1 yes I understand that this is a little bit confusing with woocommerce search.
This is How woocommerce does admin order search:
code: https://github.com/woocommerce/woocommerce/blob/master/includes/admin/class-wc-admin-post-types.php#L1372
1. Through the `wc_order_search` ( eventually go to: https://github.com/woocommerce/woocommerce/blob/master/includes/data-stores/class-wc-order-data-store-cpt.php#L497 ), it will check if the **search key is numeric or not (doesn't care if it is order id or phone number )**. And if it is numeric, it just query it with `post__in` and unset the `s` query_var. This will disable our auto `ep_integrate`
2. But when the keyword is not a numeric, then it will just do search query as usual, and EP will catch it. 


Here is what this PR does:
1. It reverse the `shop_order_search_custom_fields` from woocommerce
2. Empty the search_fields from the woocommerce `search_orders`, by return empty array to filter `woocommerce_shop_order_search_fields`, so it won't query the database. https://github.com/woocommerce/woocommerce/blob/master/includes/data-stores/class-wc-order-data-store-cpt.php#L506
3. Try to find the `order_id`, if order found then just use `post__in`
4. If not found fallback to use search query through EP

features/woocommerce/woocommerce.php
```diff
@@ -560,6 +593,9 @@ function ep_wc_setup() {
 		add_filter( 'ep_sync_taxonomies', 'ep_wc_whitelist_taxonomies', 10, 2 );
 		add_filter( 'ep_post_sync_args_post_prepare_meta', 'ep_wc_remove_legacy_meta', 10, 2 );
 		add_action( 'pre_get_posts', 'ep_wc_translate_args', 11, 1 );
+
+		add_filter( 'woocommerce_shop_order_search_fields', '__return_empty_array' );
```
tlovett1: This is already happening in the code above.

ivankristianto: This part is to empty `$search_fields`: https://github.com/woocommerce/woocommerce/blob/master/includes/data-stores/class-wc-order-data-store-cpt.php#L518
so it will skip querying the database.

tlovett1: Yes, this is already happening: https://github.com/10up/ElasticPress/blob/develop/features/woocommerce/woocommerce.php#L557

allan23:
@ivankristianto this change appears to work as expected. However, this does cause a Fatal PHP error if WooCommerce is below version 3.0. 
`Call to undefined method WC_Order::get_id() `

@tlovett1 Should backwards compatibility be added here or should EP just support the latest version of WooCommerce?

tlovett1:
@allan23 definitely need backwards compat. I merged in my own branch to develop which is similar to this one. Can you test? I use `wc_get_order` which should work since 2.2.

PR #809, opened 2017-04-21
media search feature

shpetimshala:
Index attachments and integrate ES on ajax media search on posts

### DIFF
--- a/readme.txt
+++ b/readme.txt
@@ -36,6 +36,16 @@ Please refer to [Github](https://github.com/10up/ElasticPress) for detailed usag
 
 == Changelog ==
 
+= 2.3 =
+
+Version 2.3 introduces the Media feature which empowers users to search ppt, pptx, doc, docx, xls, xlsx, and pdf files.
+
+### Enhancements
+* Media feature
+* Enable multiple feature status messages
+
+### Bug Fixes
+
 = 2.2.1 =
 
 Version 2.2.1 is a bug fix release. Here are a listed of issues that have been resolved:


features/media-search/media-search.php
```diff
@@ -0,0 +1,76 @@
...
+}
+
+/**
+ * Index all post types
+ *
+ * @param   array $post_types Existing post types.
+ * @return  array
+ */
+function ep_ms_post_types( $post_types ) {
+    $post_types[] = 'attachment';
```
tlovett1: We don't need this. Attachments are indexed as soon as we add the inherit post status.

tlovett1:
How does this work with the infinite scroll?

shpetimshala:
Query is handled by wordpress, all I had to do to make this work with ElasticPress was to change query on "ajax_query_attachments_args" filter. 

Video: https://www.youtube.com/watch?v=cSk1Pf0zRfA

tuanmh:
Good start point but this would require a lot more, as @tlovett1 said, infinite or Grid View is main problem here, off the top of my head, WP dictates the search order in client side, at the time we developed this integration, it was a hard-coded post_date so basically it will ignore search score, you would need to workaround this on server side. We developed this integration while a long time ago since EP 1.6, I'm more than happy to contribute back to ElasticPress @tlovett1 or give you a hand on this @shpetimshala.

shpetimshala:
For my case this is life saver :) I'm developing a site with +56million pageviews monthly, all in Wordpress. The only thing that was missing from ElasticPress that was hurting server performance was search in media library which is being used too much from journalists, sure it has room for improvements and I'll work on it later on based on issues and requests

tlovett1:
Long term I love the idea of media search in EP. @shpetimshala have you seen the media feature that was just added (develop branch)? I spent a lot of time trying to get around the hard coded post_date client side thing that @tuanmh mentioned to no avail. For now, I'd set up a separate repository for this EP feature.

shpetimshala:
@tlovett1 yeah I saw Media feature, I was really happy when I saw that feature in first place but after reviewing the code I saw that it's not doing what I needed. What I had to solve was search in media library to get rid of mysql because after DB grow +3GB it performs really slow. The next thing I'm working on is EP for ACF plugin. You rock btw, it's first project that intrigued me to contribute for open source, I really appreciate your work

PR #819, opened 2017-04-28
Feature/single

shpetimshala:
Handle single post / single page with elasticsearch feature


tlovett1:
@shpetimshala we've found that, given an ID, getting a post/page from MySQL is faster than Elasticsearch. What are your thoughts on that?

shpetimshala:
@tlovett1 sure it's faster with ID but slower with slug, but even with ID is slower than elastic on large database so personally I'm having better performance with elastic than mysql

tlovett1:
Got it. I'd suggest making this a separate plugin. I don't think this makes sense in EP core for now. Thanks!

Issue #820, opened 2017-04-28
Custom WP_Query force ep_integrate to true for WooCommerce product query


ivankristianto:
This is my WP_Query:
```
$query = new WP_Query(
		array(
			'ep_integrate'   => false,
			'post_type'      => 'product',
			'post_status'	=> 'publish',
			'order'	=> 'DESC',
			'orderby' => array( 'meta._regular_price.double' => 'desc', 'title' => 'asc' )
		)
	);
```

Event though I specifically put `'ep_integrate'   => false`, it will be forced to true.
Here is the code that cause it: https://github.com/10up/ElasticPress/blob/develop/features/woocommerce/woocommerce.php#L362

pcfreak30:
I had to add a pre post hook to override this and set ep_integrate to false after this code.

PR #823, opened 2017-04-30
honor ep_integrate when set in WP_Query

ivankristianto:
Fixed issue #820


tlovett1:
This PR should just check `ep_integrate` at the top of the function. If set to false, bail.

ivankristianto:
@tlovett1 PR updated, please CR.

tlovett1:
Awesome!

PR #826, opened 2017-05-03
(Closes #825) Incorrect posts_found variable when Elasticsearch query is not successful

### LINKED ISSUE #825: Incorrect posts_found variable when Elasticsearch query is not successful
I've observed an issue that when Elasticsearch server is inaccessible (or just index) that when ElasticPress allows WP_Query to perform DB query then found_posts is not the number of total found posts. This value is set to first result ID.

lukaspawlik:
Do not modify original SELECT FOUND_ROWS(); MySQL query when Elasticsearch query failed.


lukaspawlik:
Original issue: https://github.com/10up/ElasticPress/issues/825

Issue #811, opened 2017-04-21
WC Draft Product Previews Messed Up


tlovett1:
With WC active and Protected Content inactive, draft product previews are showing the incorrect product.

PR #829, opened 2017-05-08
If a WC draft is being previewed, do not use Elasticsearch.

allan23:
Mentioned in #811, ElasticPress returns the wrong product when being previewed. This PR simply checks to see if the `preview` query string is set and then prevents the search to be run on Elasticsearch.

### DIFF
--- a/features/woocommerce/woocommerce.php
+++ b/features/woocommerce/woocommerce.php
@@ -206,6 +206,13 @@ function ep_wc_translate_args( $query ) {
 	}
 
 	/**
+	 * If this is just a preview, let's not use Elasticsearch.
+	 */
+	if ( isset( $_GET['preview'] ) ) {
+		return;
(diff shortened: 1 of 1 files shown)


features/woocommerce/woocommerce.php
```diff
@@ -206,6 +206,13 @@ function ep_wc_translate_args( $query ) {
 	}
 
 	/**
+	 * If this is just a preview, let's not use Elasticsearch.
+	 */
+	if ( isset( $_GET['preview'] ) ) {
```
ivankristianto: I would use `$query->get( 'preview', false )` instead accessing query_string.

allan23: Good call.

allan23:
@ivankristianto this has been updated.

Issue #813, opened 2017-04-25
Create settings for Decay value for search feature


ivankristianto:
The search feature come with the function_score query: https://www.elastic.co/guide/en/elasticsearch/guide/2.x/function-score-query.html

This is where the code is:
https://github.com/10up/ElasticPress/blob/develop/features/search/search.php#L168-L170

We need to have a settings to set this decay value.
cc: @tlovett1 and @tott

lukaspawlik:
@allan23 @tlovett1 @ivankristianto 

I've built something similar for our customer who needs to play with values to find "the best ones".

Here is the screenshot how configuration looks like. We don't show the offset value as customer is interested only in number of days and "decaying aggression". We also support disabling decaying feature at all.

<img width="1477" alt="screen shot 2017-05-03 at 14 31 41" src="https://cloud.githubusercontent.com/assets/7304218/25660610/55d28c8e-300d-11e7-8781-6af601334704.png">

ivankristianto:
@lukaspawlik it's awesome. yes couple users has been asking about this decay function.
But I wonder why not put that setting into Search module instead? because it is more relevant there.

lukaspawlik:
@ivankristianto An update: I've just added Offset configuration too as customer asked me to do this.  I decided to build a separate new ElasticPress feature for that instead of hooking into `ep_feature_box_settings` action for Search module as JavaScript responsible for sending configuration doesn't pick any non radio field into account when building settings object sent via AJAX (I ported some of this JS into my module logic with the fix to pick all fields instead of radios).

```JavaScript
                var settings = {};

		var $settings = $feature.find('.setting-field');

		$settings.each(function() {
			var type = $( this ).attr( 'type' );
			var name = $( this ).attr( 'data-field-name' );
			var value = $( this ).attr( 'value' );

			if ( 'radio' === type ) {
				if ( $( this ).attr( 'checked' ) ) {
					settings[ name ] = value;
				}
			}
		});
```

used in 

```JavaScript
$.ajax( {
			method: 'post',
			url: ajaxurl,
			data: {
				action: 'ep_save_feature',
				feature: feature,
				nonce: epDash.nonce,
				settings: settings
			}
		} )
```

I am happy to share my code however it won't fit plugin well as I manipulate the entire query structure (depending on the fact if decaying is enabled or not) via `ep_formatted_args` filter. I'll try to prepare PR next week (or in next two weeks) for that when I have a moment.

tlovett1:
[...]

PR #832, opened 2017-05-12
Decaying configuration

lukaspawlik:
Proposed solution for https://github.com/10up/ElasticPress/issues/813

I am attaching screenshot to show how this looks like (I'd appreciate any help with better wording or adding any information what those settings mean).

screen shot 2017-05-12 at 19 58 04

@tlovett1 @ivankristianto I'd like also to suggest adding extra setting for `'boost_mode' => 'sum'` value. From my experiences `sum` mode doesn't have big impact of articles decaying and it's not useful for publishers where most recent content should be promoted a bit more aggressive. From my understanding final score is being calculated as a sum or multiplication of two scores: regular document score and document score after applying decaying function and for example:

regular score for document is 139.46645 and second score after running decay function is 0.8473983 (for 15 days old article). As you may observe decaying function using `sum` `boost_mode` has very little impact on overall score. If this example is applied to 7 months old article (with a bit higher regular document score ~141) then decay function score is ~0.0289 and sum of those two gives us higher score for 7 months old article than for 15 days old (their regular scores are almost equal). When applied `multiply` `boost_mode` the difference between these articles and their scores is significant and most recent article is really promoted over outdated one.

I was working on ElasticPress integration with two publishers and both integrations ended with decay `boost_mode` set to `multiply`.

What are your thoughts @tlovett1 and @ivankristianto ?


lukaspawlik:
Attaching some screenshot with explained scoring for my example:

screen shot 2017-05-12 at 20 12 32

allan23:
Hey @lukaspawlik this is useful, but it may be too much for the basic user to comprehend. It would be best to have a simple on/off toggle for this instead of these advanced settings.

So perhaps just a simple settings box just for decay enabled/disabled.

lukaspawlik:
Hey @allan23 - I fully agree with you however the original expectation in https://github.com/10up/ElasticPress/issues/813 was to be able to configure decay value. From my experiences this feature / configuration change will be almost useless without modifying other parameters as well. How about adding "Advanced settings" link that would show more configurable details?

I am also fine with removing these 3 config settings and leaving just on/off switch - just wanted to share my experiences working with publishers group.

allan23:
@lukaspawlik  ideally, the simpler the better. I believe we can just have an on/off and have the values filterable.

@tlovett1 any thoughts?

tlovett1:
I'd much prefer just an on/off switch for now.

lukaspawlik:
👍

lukaspawlik:
@allan23 @tlovett1 I've applied requested changes. Only basic switch on/off left.

tlovett1:
This works well @lukaspawlik. Nice work! Two things:

1. Let's change the form label from "Decaying enabled" to "Weight results by date".
2. Let's move the tests file into `tests/features`

allan23:
Hey @lukaspawlik, 
When do you think you can get to the last two items?

lukaspawlik:
Hi @allan23 ,

I've reserved some time for tomorrow (Jul 18th).

lukaspawlik:
@allan23 it's done

tests/features/test-decaying-configuration.php
```diff
@@ -0,0 +1,143 @@
...
+		$this->assertTrue( isset(
+			$this->fired_actions['ep_formatted_args']['query'],
+			$this->fired_actions['ep_formatted_args']['query']['function_score'],
+			$this->fired_actions['ep_formatted_args']['query']['function_score']['exp'],
+			$this->fired_actions['ep_formatted_args']['query']['function_score']['exp']['post_date_gmt'],
+			$this->fired_actions['ep_formatted_args']['query']['function_score']['exp']['post_date_gmt']['scale'],
+			$this->fired_actions['ep_formatted_args']['query']['function_score']['exp']['post_date_gmt']['decay'],
+			$this->fired_actions['ep_formatted_args']['query']['function_score']['exp']['post_date_gmt']['offset']
+		) );
+		$this->assertEquals( '14d', $this->fired_actions['ep_formatted_args']['query']['function_score']['exp']['post_date_gmt']['scale'] );
```
tlovett1: Let's remove testing the actual values in case they change.

tests/features/test-decaying-configuration.php
```diff
@@ -0,0 +1,143 @@
+<?php
+
+class EPTestDecayingConfiguration extends EP_Test_Base {
```
tlovett1: Can we just move this to the search tests?

tlovett1:
Overall, this is good. We'll add to 2.4. There are a couple minor tweaks I requested. Thanks!

lukaspawlik:
@tlovett1 I've merged _develop_ and resolved all conflicts and I also adjusted tests as you requested. It's ready for your review.

PR #840, opened 2017-05-19
Fix/product visibility

### LINKED ISSUE #827: Wrong result count with Woocommerce Hide out of stock items from the catalog option
After updating to Woocommerce 3, the result count and pagination of products is incorrect with the "Hide out of stock items from the catalog" option selected ( in product inventory settings ).

I tested this on a clean install with wpdocker and the woocommerce test data.

23 total products, "Hide out of stock items from the catalog" option, set 1 to out of stock

Store page results have 23 products, 3 pages in pagination.

First page shows 10 products
Second page shows 9 products, this is the page where the out of stock product would have displayed.
Third page shows 3 products

Disabling ElasticPress shows results as expected, 22 total, page 1 is 10, page 2 is 10, page 3 is 2.

ivankristianto:
Fixed #827 

Changelogs:
* Add `term_taxonomy_id` to the mapping
* add woocommerce `product_visibility` taxonomy to supported_taxonomies

### DIFF
--- a/classes/class-ep-api.php
+++ b/classes/class-ep-api.php
@@ -698,10 +698,11 @@ class EP_API {
 			foreach ( $object_terms as $term ) {
 				if( ! isset( $terms_dic[ $term->term_id ] ) ) {
 					$terms_dic[ $term->term_id ] = array(
-						'term_id'  => $term->term_id,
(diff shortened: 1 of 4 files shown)


allan23:
This PR resolves the count issue.

PR #850, opened 2017-05-28
add woocommerce related product widget

ivankristianto:
Changelogs:
1. Add new widget: `EP_Related_Products_Widget` 
2. Add new function: `ep_wc_get_related_products`

Fixed #756

### DIFF
--- a/features/woocommerce/widget.php
+++ b/features/woocommerce/widget.php
@@ -1,0 +1,124 @@
+<?php
+
+/**
+ * Related products widget
+ *
+ * @since  2.4
+ */
+class EP_Related_Products_Widget extends WP_Widget {
+
+	/**
+	 * Initialize the widget
+	 */
+	public function __construct() {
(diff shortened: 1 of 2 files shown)


tlovett1:
I'm experiencing some strange behavior:

I have three products with similar names. When looking at any of the products, they all show a related products which makes sense. When I add a product tag to any of them, none show in the widget.

What is the expected functionality for this? How does the algorithm work?

ivankristianto:
The way it works is using ES `more_like_this` query: 
https://www.elastic.co/guide/en/elasticsearch/reference/5.5/query-dsl-mlt-query.html

So here is the query in Related product widget:
https://github.com/10up/ElasticPress/blob/develop/features/related-posts/related-posts.php#L10

Where the `more_like` query is introduce in:
https://github.com/10up/ElasticPress/blob/develop/features/related-posts/related-posts.php#L10
Which I think we should move this to the `ep_formatted_args` to make it usable in WP_Query without activate Related posts feature.

PR #865, opened 2017-06-19
Remove post_meta key from ES post object

tlovett1:
This was left around for backwards compat. It caused issues for people since it doubled the size of meta stored in ES.

@ivankristianto can you test?


ivankristianto:
@tlovett1 i have look and test it.
We are not using `post_meta` anywhere else anymore.
And `post_meta` also has been unset in woocommerce feature: https://github.com/10up/ElasticPress/blob/develop/features/woocommerce/woocommerce.php#L516

can we also delete this `post_meta` in another place such as:
https://github.com/10up/ElasticPress/blob/develop/classes/class-ep-wp-query-integration.php#L302

PR #871, opened 2017-06-23
fix for single product template

isvaljek:
Better check for single product template.


tlovett1:
This won't work. This would return if the current query is not a single but the main query is.

PR #874, opened 2017-06-26
properly unset author_name from the search fields

### LINKED ISSUE #873: author_name not an indexed field
I'm using the debug bar to view queries sent and results received in the wordpress search page. It is requesting the following fields:
```
                                "fields": [
                                    "post_title",
                                    "post_content",
                                    "post_excerpt",
                                    "author_name",
                                    "terms.post_tag.name",
                                    "terms.category.name",
                                    "post_author.login",
                                ],
```
Looking at my index in Kibana, I show the following available fields

```
post_author.display_name
post_author.id
post_author.login
post_author.raw
```
I know that author_name is a valid WP_Query parameter, corresponding to user_nicename. However, it appears that this is not an indexed field? Maybe some incorrect field mapping/indexing here?

ivankristianto:
fixed #873


tlovett1:
Seems like we are doing the same thing a few lines earlier with meta?

ivankristianto:
Yes I just notice that. I updated the code

tlovett1:
This needs to be thought out and tested more @ivankristianto.

ivankristianto:
@tlovett1 I revert code and check it.
For the `meta` and `taxonomies`, it is a key-value element.

PR #878, opened 2017-06-29
Cache Plugins Request for 5 Minutes

### LINKED ISSUE #863: ElasticPress queries /_nodes/plugins on every pageload
Using Query Monitor to watch HTTP requests made by ElasticPress.

I'm seeing an HTTP request to our ES instance on every pageload, regardless of whether that page is using ES functionality or not in both front-end and admin pages. The HTTP request goes to /_nodes/plugins on the ES remote instance.

Seems like a lot of unneeded HTTP requests. Shouldn't this fire only when ES capabilities are actually needed?

tlovett1:
This fixes #863.

Save request results in a transient. This will unfortunately cause the user to have to wait 5 minutes to be shown a dashboard notification if their ES goes down or plugins change.

### DIFF
--- a/classes/class-ep-api.php
+++ b/classes/class-ep-api.php
@@ -2176,54 +2176,91 @@ class EP_API {
 	public function get_elasticsearch_info( $force = false ) {
 
 		if ( $force || null === $this->elasticsearch_version || null === $this->elasticsearch_plugins ) {
-			$path = '_nodes/plugins';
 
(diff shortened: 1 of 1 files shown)


tlovett1:
@ivankristianto @tott can you review/test?

allan23:
@tlovett1 this is working as expected.

(CC: @tott )

PR #887, opened 2017-07-11
add sanity check for taxonomy in prepare_terms

### LINKED ISSUE #881: Trying to get property of non-object in  /elasticpress/classes/class-ep-api.php on line 690
Yes

Bug

When there is an error with a taxonomy and WP returns taxonomy with value of`false`, a `Trying to get property of non-object` error is thrown. At https://github.com/librarystack/ElasticPress/blob/develop/classes/class-ep-api.php#L690, EP tries to access object property `$taxonomy->name`. We need a sanity check prior to this to prevent the error . 
Addding `if(! $taxonomy) continue;` on previous line fixes issue. 

Current WordPress version:
4.8

Current ElasticPress version:
2.3.2

1.5

AWS

ivankristianto:
Fixed #881

### DIFF
--- a/classes/class-ep-api.php
+++ b/classes/class-ep-api.php
@@ -687,6 +687,10 @@ class EP_API {
 		$allow_hierarchy = apply_filters( 'ep_sync_terms_allow_hierarchy', false );
 
 		foreach ( $selected_taxonomies as $taxonomy ) {
+			if( ! is_a( $taxonomy, 'WP_Taxonomy' ) ){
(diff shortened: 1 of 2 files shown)


tlovett1:
This looks solid.

tlovett1:
Hey @ivankristianto I had to remove the is_a WP_Taxonomy check is WP_Taxonomy does not exist prior to 4.7. Is there another solution here?

PR #888, opened 2017-07-11
moved setup_feature on init

### LINKED ISSUE #854: Unable to connect after upgrade
Upgraded elasticpress from 2.2.1 to 2.3 with EP_HOST set in wp-config, after the update it refuses to connect. Never ending "Fix it".

Have the http_request_args to enable self-signed SSL certs, It doesn't seem to matter if this filter is active or not.

Ideas?

ivankristianto:
Fixed #854 

I did run few tests on feature activations, run unit testing and try some search and WP_Query. Seems okay.
But still need some more testing before we merge this


eugene-manuilov:
Going to merge this pool request since there is nothing to concern about.

PR #891, opened 2017-07-13
Index Taxonomies that have publicly_queryable as true

allan23:
This PR will allow taxonomies that are not public to still be queryable.


tlovett1:
This is good to go.

jtsternberg:
> This is good to go.

Does that mean this will be in the next .org release? Asking as I need this functionality. If it will end up in the next release, I will just edit the corresponding line in the current plugin.

tlovett1:
Yep, this will be in next .org version.

PR #894, opened 2017-07-13
Normalizer for sortable fields

allan23:
By adding a normalizer in the 5.0 mappings for sortable fields, the case-sensitivity issue is resolved. This results in five out of the six failing unit tests to pass. 
(The testSearchPostDateOrderbyQuery test is only returning two results instead of three. I'm not sure if this is because the post_title is `ordertes` instead of `ordertest`.)

@tlovett1 and @ivankristianto can you take a look at this to verify if this resolves the issue or if this has the potential to cause different issues to occur.

Edit: The normalizer was introduced in 5.2, so I have added a 5.2 mapping for backwards compatibility.

### DIFF
--- a/includes/mappings/5-2.php
+++ b/includes/mappings/5-2.php
@@ -1,8 +1,8 @@
 <?php
 /**
- * Elasticsearch mapping for ES 5.0
+ * Elasticsearch mapping for ES 5.2
  *
- * @since  2.1.2
+ * @since  2.4
  * @package elasticpress
  */
 
@@ -53,6 +53,12 @@ return array(
 					'type' => 'edgeNGram',
 				),
 			),
+			'normalizer' => array(
+				'lowerasciinormalizer' => array(
+					'type'   => 'custom',
+					'filter' => array( 'lowercase', 'asciifolding' ),
+				),
+			),
 		),
 	),
 	'mappings' => array(
@@ -90,6 +96,7 @@ return array(
 										'sortable' => array(
 											'type' => 'keyword',
 											'ignore_above' => 10922,
+											'normalizer' => 'lowerasciinormalizer',
 										),
 										'raw' => array(
 											'type' => 'keyword',
@@ -141,6 +148,7 @@ return array(
 										),
 										'sortable' => array(
 											'type' => 'keyword',
+											'normalizer' => 'lowerasciinormalizer',
 										),
 									),
 								),
@@ -191,6 +199,7 @@ return array(
 								),
 								'sortable' => array(
 									'type' => 'keyword',
+									'normalizer' => 'lowerasciinormalizer',
 								),
 							),
 						),
@@ -235,6 +244,7 @@ return array(
 						'sortable' => array(
 							'type' => 'keyword',
 							'ignore_above' => 10922,
+							'normalizer' => 'lowerasciinormalizer',
 						),
 					),
 				),
--- a/includes/mappings/5-2.php
+++ b/includes/mappings/5-2.php
@@ -211,6 +211,7 @@ return array(
 								),
 								'sortable' => array(
 									'type' => 'keyword',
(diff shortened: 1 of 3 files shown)


ivankristianto:
@allan23 
I tested it with ES 5.1.1 we have 5 test failures:
```
There were 6 failures:

1) EPTestSingleSite::testSearchPostTitleOrderbyQuery
Failed asserting that two strings are equal.
--- Expected
+++ Actual
@@ @@
-'Ordertest 222'
+'ordertest 111'

/srv/www/10up/htdocs/wp-content/plugins/elasticpress/tests/test-single-site.php:1380

2) EPTestSingleSite::testSearchPostMetaStringOrderbyQueryAsc
Failed asserting that two strings are equal.
--- Expected
+++ Actual
@@ @@
-'ordertest 111'
+'Ordertest 222'

/srv/www/10up/htdocs/wp-content/plugins/elasticpress/tests/test-single-site.php:1407

3) EPTestSingleSite::testSearchPostMetaStringOrderbyQueryAscArray
Failed asserting that two strings are equal.
--- Expected
+++ Actual
@@ @@
-'ordertest 111'
+'Ordertest 222'

/srv/www/10up/htdocs/wp-content/plugins/elasticpress/tests/test-single-site.php:1435

4) EPTestSingleSite::testAuthorDisplayNameOrderbyQueryAsc
Failed asserting that two strings are equal.
--- Expected
+++ Actual
@@ @@
-'findme test 3'
+'findme test 1'

/srv/www/10up/htdocs/wp-content/plugins/elasticpress/tests/test-single-site.php:1529

5) EPTestSingleSite::testSearchTaxNameOrderbyQueryAsc
Failed asserting that two strings are equal.
--- Expected
+++ Actual
@@ @@
-'Ordertest 222'
+'ordertest 333'

/srv/www/10up/htdocs/wp-content/plugins/elasticpress/tests/test-single-site.php:1594

6) EPTestSingleSite::testSearchPostDateOrderbyQuery
Failed asserting that 2 matches expected 3.

/srv/www/10up/htdocs/wp-content/plugins/elasticpress/tests/test-single-site.php:1684

FAILURES!
Tests: 185, Assertions: 611, Failures: 6.
```

and with ES 5.2.2 still have 1 failures:
```
There was 1 failure:

1) EPTestSingleSite::testSearchPostDateOrderbyQuery
Failed asserting that 2 matches expected 3.

/srv/www/10up/htdocs/wp-content/plugins/elasticpress/tests/test-single-site.php:1684

FAILURES!
Tests: 185, Assertions: 620, Failures: 1.
```

And your code here: https://github.com/10up/ElasticPress/pull/894/files#diff-9703716231f66c3ddbc0030ef4dc582cR469
explicitly check for version 5.2, what if later we support for 5.3 & 5.4, the mapping will fallback to 5-0.php.
I suggest it fallback to 5.2.php instead

allan23:
@ivankristianto did I read the PHP docs wrong for version_control?

`By default, version_compare() returns -1 if the first version is lower than the second, 0 if they are equal, and 1 if the second is lower. `

`elseif ( version_compare( $es_version, '5.2' ) >= 0 ) {`

ivankristianto:
@allan23 I was wrong.
Perhaps we could use `switch` there instead of `if..else`? so it's easier to read?

includes/mappings/5-2.php
```diff
@@ -0,0 +1,353 @@
...
+								),
+							),
+						),
+						'login' => array(
+							'type' => 'text',
+							'fields' => array(
+								'raw' => array(
+									'type' => 'keyword',
+								),
+								'sortable' => array(
```
tlovett1: No normalizer?

allan23: Added.

allan23:
Normalizer added to that last sortable

tlovett1:
@allan23 tested this and it works. Nice job!

PR #897, opened 2017-07-18
Fix(spelling): The word Occuring is not spelled right.

IWriteThings:
I update the word occuring to occurring. This is a typo that should be fixed.


tlovett1:
Thanks!

PR #899, opened 2017-07-21
Fix for missing attachments in AJAX indexing from Admin settings screen

cameronterry:
This pull request contains a fix for when you use the Play / Pause / Resume / Refresh buttons on the ElasticPress admin screen which missed Attachments post type (which the rest of plugin seems to handle). 

When testing locally on my WordPress install, I noticed that Attachments were being excluded as the post status "inherit" was missing. The fix included is perhaps overly simple, but hopefully helpful in some form.

* Adds "inherit" post status to the indexable post status.
* But only if the post type Attachments indexable post type is included.
* Makes the AJAX indexing process work consistently with the rest of the plugin (by ensuring Attachments are indexed), which adds new "attachments" to Elasticsearch when added to the Media Library.
* Revisions are excluded (as it is stored as a separate post type).

### DIFF
--- a/classes/class-ep-dashboard.php
+++ b/classes/class-ep-dashboard.php
@@ -535,10 +535,21 @@ class EP_Dashboard {
 		
 		do_action( 'ep_pre_dashboard_index', $index_meta, $status );
 
+		$indexable_post_types = ep_get_indexable_post_types();
+		$indexable_post_status = ep_get_indexable_post_status();
+		/**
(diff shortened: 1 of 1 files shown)


tlovett1:
Hey @cameronterry, we only index `inherit` post type when the Documents feature is active:

https://github.com/10up/ElasticPress/blob/develop/features/documents/documents.php#L172

If you want to change this behavior I'd recommend just doing so with the filter provided.

PR #903, opened 2017-07-27
Add support for post__in orderby clause

### LINKED ISSUE #330: Adding orderby post__in support

sc0ttkclark:
Includes tests, waiting to confirm the tests pass. Will make adjustments if tests fail for whatever reason.

Fixes #330

### DIFF
--- a/README.md
+++ b/README.md
@@ -348,7 +348,7 @@ ElasticPress integrates with `WP_Query` if the `ep_integrate` parameter is passe
     
 * ```orderby``` (*string*)
 
-    Order results by field name instead of relevance. Supports: ```title```, ```modified```, `meta_value`, `meta_value_num`, ```type```, ```name```, ```date```, and ```relevance```; anything else will be interpretted as a document path i.e. `meta.my_key.long` or `meta.my_key.raw`. You can sort by multiple fields as well i.e. `title meta.my_key.raw`
+    Order results by field name instead of relevance. Supports: ```title```, ```modified```, `meta_value`, `meta_value_num`, ```type```, ```name```, ```date```, ```post__in```, and ```relevance```; anything else will be interpretted as a document path i.e. `meta.my_key.long` or `meta.my_key.raw`. You can sort by multiple fields as well i.e. `title meta.my_key.raw`
 
 * ```order``` (*string*)
 
--- a/classes/class-ep-api.php
+++ b/classes/class-ep-api.php
@@ -1959,7 +1959,7 @@ class EP_API {
 			}
 
 			if ( ! empty( $orderby_clause ) && 'rand' !== $orderby_clause ) {
-				if ( 'relevance' === $orderby_clause ) {
+				if ( 'relevance' === $orderby_clause || 'post__in' === $orderby_clause ) {
 					$sort[] = array(
 						'_score' => array(
 							'order' => $order,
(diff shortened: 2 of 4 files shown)


sc0ttkclark:
Related PR attempts: #373, #523

sc0ttkclark:
The fails in the tests don't appear to be related to the post__in orderby tests, so that's great :)

Just needs code review, updating `@since` in phpdoc comments, and then final scheduling for release.

tlovett1:
@sc0ttkclark does this work with pagination?

sc0ttkclark:
Yes, since `post__in` just restricts the data to a specific range of post IDs that match, any `posts_per_page` / `paged` will just work off of those. The ordering logic included in this PR just ensures the order is maintained for any posts returned by ES, based on the order of the full `post__in` array.

tlovett1:
Hmmm, I don't think this will work. Look at this:
```php
// Post ids: 1, 2, 3, 4

$posts = new WP_Query( [
  'post__in' => [ 1, 2, 3, 4 ],
  'posts_per_page' => 2,
  'ep_integrate' => true,
  'orderby' => 'post__in',
  'order' => 'asc',
] );

ES returns posts 4, 3 for page 1.

EP orders them 3, 4 (correct would be 1, 2)
```

This would only work if ES respected the post in order which it does not.

sc0ttkclark:
Updated branch with latest develop

sc0ttkclark:
@tlovett1 Yes, ES does not have the ability (outside of using a custom sort script, which may be disabled for many Elasticsearch hosts/configs) to sort the results by a specific pre-defined order other than by a field itself.

So yes, paged queries will not sort the *entire* set as the MySQL equivalent does, but it does sort the subset that's returned. With that as a known limitation, we could add the caveat as a note in the `orderby` section of the readme for `post__in`.

**What this PR does:**

* Sorts the returned posts by the order of the post IDs passed in `post__in`

**What this PR cannot do:**

* Sort the posts being queried by ES by the order of the post IDs passed in `post__in`, due to limitations of many hosts not allowing sort scripts (and other scripts for that matter)

**Next steps:**

* @tlovett1 to provide feedback on whether to limit the sorting functionality to only queries that have paged = 1
* @tlovett1 to provide approval to include caveat text in the `orderby` readme for `post__in`

I do not think the inability to sort the posts on ES should hold this feature back since most usages of `post__in` I've ever needed/used or seen by other developers have not been paged queries. The functionality in this PR provides much needed functionality that does 100% of what is currently possible with the constraints of ES hosts.

sc0ttkclark:
Limited support is better than none? Open to ideas for solutions that can be supported on many ES hosts. This could just be part one, switching to more comprehensive support long-term when ES gains the ability to sort by supplied arg values or when we find a better approach.

tlovett1:
I don't think adding a solution that partially works for one scenario is the right long term decision for the project. Again, happy to hear others opinions. Also happy to merge a reasonable complete solution.

Thanks for the PR and thoughtful discussion!

sc0ttkclark:
I'll keep thinking about this and will offer up a Gist for getting this functionality to work in the meantime for those of us using post__in without pagination.

Here's another complicated workaround:

- ES post__in query kicks off X number of secondary ES queries
- Each secondary ES query limits by a single post ID (same query args, except post_id = $each_post_id_in_array)
- If a result is returned, add it to the list, increment counter and only return max of what posts_per_page is set to
- If a result is not returned, continue the ES queries until hitting posts_per_page OR no more post__in IDs are left

Yep, I think there's officially no *good* way to do what orderby => post__in is doing in MySQL, in a way that ES can understand -- without a custom ES sort script involved, which aren't allowed on many ES hosts.

Issue #909, opened 2017-08-04
Removes Ability to View Orders In WooCommerce Orders Page


BurlesonBrad:
Have you searched for similar issues before submitting this one?
yep
Is this a bug, question or feature request?
bug
Describe the issue you encountered:
Activate the WC mod in ```/wp-admin/admin.php?page=elasticpress``` & the # of pages goes to *9*

*DE*activate the WC module in ```/wp-admin/admin.php?page=elasticpress``` & all *(167)* pages of orders comes back
![dn-with-es-order-page](https://user-images.githubusercontent.com/6208757/28944013-86028494-7866-11e7-8990-2c309a72646f.png)

Current WordPress version: Current ElasticPress version: Current Elasticsearch version: Where do you host your Elasticsearch server: Other plugins installed (WooCommerce, Simple Redirect Manager, etc..):
https://gist.github.com/BurlesonBrad/eee52e295b840dab912647c2dff8bd59
Steps to reproduce:

1. get a big WooCommerce site with years of orders
2. hope that ElasticPress will help with the speed and indexing of the massive amounts of tables
3. open WooCommerce orders page `/wp-admin/edit.php?post_type=shop_order` 👍 

Screenshots, if needed:
Original number of pages and orders displayed = https://prnt.sc/g41ywo
Activate, Install, Configure, Index ElasticPress = https://prnt.sc/g42676

ivankristianto:
@BurlesonBrad did you make sure that You did fully indexing after activate elasticpress?

BurlesonBrad:
well.... therein lies another conundrum.
@Pagely swears on their life that nothing is of their control, yet at the same time, each time the attempt is made to index https://prnt.sc/g4d1vt I'm left with the message of https://prnt.sc/g4cbut

https://prntscr.com/g4dw1j <-- Debugbar + EP debug bar both on, in, and firing w/out showing errors.

@ivankristianto , I want 'ya to shoot straight with me here so that I can get back over to diagnosing other error messages, product entry conundrums, customer user flow, ads, plugin issues, and things like that.... 😉 
### is Pagely blowing sunshine up my keister here?
_(for the record: I'm not saying that they are, just trying to get a straight answer from them regarding ports 9300 & 9201, and "understanding if timeout issues occur" has been quite a support ticket ping ping game)_ 

Oh, wow. Ok. As I was typing this, I got another reply on the ticket:
[...]

PR #912, opened 2017-08-09
WooCommerce: remove product variations

kallehauge:
_I have completely forgotten to suggest this code. It's something that I've had in a general search plugin I share from project to project:_

We've often experienced clients who change a product from a "Variable product" to a "Simple product" or similar without deleting the variations, first.
The issue is so normal that WooCommerce themselves have a tool in WC Core to remove "variable product orphans".

But why is this relevant for ElasticPress? Well, we would sometimes get clients telling us that the "search doesn't work anymore" and it's due to a fatal PHP error. The crash was typically from a product variant trying to fetch its permalink but could not find it's parent.
In the latest version of WooCommerce, this would end out in a "Maximum function nestling level..." error.

### DIFF
--- a/features/woocommerce/woocommerce.php
+++ b/features/woocommerce/woocommerce.php
@@ -585,6 +585,51 @@ function ep_wc_search_order( $wp ){
 }
 
 /**
+ * Clear WooCommerce product variations orphans before indexing (WP CLI).
+ *
+ * We do this since it is possible for product variations without parents to
+ * crash the EP index sync. The most "popular" error is when fetching the
+ * permalink for the variation.
+ *
+ * @since TBD
+ *
+ * @param array $args
+ * @param array $assoc_args
+ */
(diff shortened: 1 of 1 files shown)


ivankristianto:
Thanks @kallehauge this is an awesome PR.
I did try this to fix issue in #909, I still need to run the index twice, where the second index works as expected.
But the first attempt, it still query the orphaned product variation.

kallehauge:
Thank you very much @ivankristianto ☺️ I will try and take a look at it in the coming week 👍

kallehauge:
@ivankristianto we might not have to do this anymore since the "get permalink infinite loop" is getting fixed in the next release of [WooCommerce ](https://github.com/woocommerce/woocommerce/pull/16427/).

ivankristianto:
So when the parent post is not exist. the permalink would back to default? 
But that PR certainly fix the infinite loop issue.

I agree to close this PR, since we don't want to alter any data in database at this plugin scope.

PR #760, opened 2017-03-09
Add WP_Query fields support

kallehauge:
Support the two `WP_Query` fields cases that are supported in core.

You can read more about my rant of why/how/thoughts in the open issue: #724 

_**NB**: I have not written any tests and I am not sure that I have covered all cases in the different modules (ex `WooCommerce`) - but I have manually tested it on regular searches on a vanilla Woo install with `storefront` theme and dummy data._

/ Kallehauge

PR #913, opened 2017-08-11
Proposal for using posts_pre_query action for overriding the $posts

### LINKED ISSUE #904: Filter the posts using posts_pre_query
I'm suggesting that in addition to `the_posts` filter, that you also add the filter to `posts_pre_query`:

```php
add_filter( 'the_posts', array( $this, 'filter_the_posts' ), 10, 2 );
add_filter( 'posts_pre_query', array( $this, 'filter_the_posts' ), 10, 2 );
```

Then in `EP_WP_Query_Integration::filter_the_posts()` it runs an additional conditional:

```php
	public function filter_the_posts( $posts, $query ) {
		if ( ! ep_elasticpress_enabled( $query ) || apply_filters( 'ep_skip_query_integration', false, $query ) || ! isset( $this->posts_by_query[spl_object_hash( $query )] ) ) {
			return $posts;
		} elseif ( current_filter( 'the_posts' ) && ! did_action( 'posts_pre_query' ) ) {
			// Only filter on the_posts if the posts_pre_query was never run (pre WP 4.6)
			return $posts;
		}

		$new_posts = $this->posts_by_query[spl_object_hash( $query )];

		return $new_posts;
	}
```

The work being done inside #760 enables this, so my suggestion is to take that and combine it with this tweak. This will allow all queries to bypass querying the DB in `WP_Query::get_posts()` when EP has already retrieved the posts.

sc0ttkclark:
Overriding the $posts during this action prevents the real DB query from getting kicked off and allows ElasticPress to be compatible with `fields` support which was added via #760

Fixes #904

Needs unit tests to confirm everything is working as expected still, will wait to see how the results of that turn out.

### DIFF
--- a/classes/class-ep-wp-query-integration.php
+++ b/classes/class-ep-wp-query-integration.php
@@ -47,10 +47,10 @@ class EP_WP_Query_Integration {
 		// Nukes the FOUND_ROWS() database query
 		add_filter( 'found_posts_query', array( $this, 'filter_found_posts_query' ), 5, 2 );
 
-		// Support "fields".
-		add_filter( 'posts_pre_query', array( $this, 'posts_fields' ), 10, 2 );
+		// Query and filter in EP_Posts to WP_Query (WP 4.6+)
+		add_filter( 'posts_pre_query', array( $this, 'filter_the_posts' ), 10, 2 );
 
-		// Query and filter in EP_Posts to WP_Query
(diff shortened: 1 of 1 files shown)


sc0ttkclark:
Need to check into why it's doing that, it's not doing it locally prior to #760 so I'll find out and update when this is ready for review again.

PR #917, opened 2017-08-14
WooCommerce order products search

kallehauge:
Products on orders are, by default, searchable by name on the orders overview page in the admin interface. This PR is an easy way to support the same functionality out of the box and 90% mimics the way WooCommerce currently does (from my understanding).


tlovett1:
Looks great!

PR #926, opened 2017-08-24
Support for ES 5.5.x and strict content type checking

### LINKED ISSUE #925: Support for strict content type checking in ES version 5.x and 6.x
You can enable strict content type checking in 5.x and it's always forced in ES 6.x. It's a fairly simple thing to support, it just requires `Content-Type` to be set on the requests.

In `EP_API::query()` calls, where `$request_args` is set a new line should be added:

```
		$request_args = array(
			'body'    => json_encode( apply_filters( 'ep_search_args', $args, $scope, $query_args ) ),
			'headers' => array(
				'Content-Type' => 'application/json',
			),
			'method'  => 'POST',
		);
```

Without that, I have to add the content type header via the `ep_search_request_args` filter to get requests to work. Everything else works great on ES version 5.5.1 which I've been testing on.

Leaving this here without a PR mainly because I'm not sure how you want to handle testing this for yourselves or if there should be unit testing involved on the newer version.

sc0ttkclark:
Fixes #925

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,6 +17,7 @@ php:
 env:
   - WP_VERSION=latest WP_MULTISITE=1 ES_VERSION="5.0.2"
   - WP_VERSION=latest WP_MULTISITE=1 ES_VERSION="5.2.2"
+  - WP_VERSION=latest WP_MULTISITE=1 ES_VERSION="5.5.1"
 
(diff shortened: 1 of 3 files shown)


tlovett1:
Thanks, Scott!

PR #930, opened 2017-09-06
Filter which post types will be queried for suggestions

moraleida:
The `epas` object holds a `postType` key with the `ep_term_suggest_post_type` filter which is currently unused. All queries search for all post types without distinction.

This introduces a `post_filter` object to the query being run whenever the `postType` key is not the default `all`. The filter can now accept simple strings (such as "post"), an array or a comma-separated list of post types.


moraleida:
Added a new filter to this, in which the user is able to pass which fields will be used for the multi_match query. This makes it possible that the theme define an indexed custom field to be searched.

Issue #931, opened 2017-09-13
Late escaping, per WordPress.com VIP review


jasonbahl:
We're planning on using the plugin on some WordPress.com VIP hosted sites, and there are some blockers from their code review that are preventing us from getting the plugin approved for usage with VIP. 

The blockers are listed below: 

```
includes/dashboard-page.php line 76: MISSING ESCAPING: $left
includes/dashboard-page.php line 79: MISSING ESCAPING: $right
includes/settings-page.php line 42: esc_html_e() \ esc_attr_e() should be used instead of _e() for translation escaping.
includes/intro-page.php line 25: esc_html_e() \ esc_attr_e() should be used instead of _e() for translation escaping.
includes/intro-page.php line 27: MISSING ESCAPING: EP_URL
classes/class-ep-dashboard.php line 288: esc_html__() \ esc_attr() should be used instead of __() for translation escaping.
classes/class-ep-dashboard.php line 295: esc_html__() \ esc_attr() should be used instead of __() for translation escaping.
classes/class-ep-dashboard.php line 302: esc_html__() \ esc_attr() should be used instead of __() for translation escaping.
classes/class-ep-dashboard.php line 315: esc_html__() \ esc_attr() should be used instead of __() for translation escaping.
classes/class-ep-dashboard.php line 328: esc_html__() \ esc_attr() should be used instead of __() for translation escaping.
classes/class-ep-dashboard.php line 341: esc_html__() \ esc_attr() should be used instead of __() for translation escaping.
classes/class-ep-dashboard.php line 356: esc_html__() \ esc_attr() should be used instead of __() for translation escaping.
classes/class-ep-dashboard.php line 364: esc_html__() \ esc_attr() should be used instead of __() for translation escaping.
classes/class-ep-dashboard.php line 371: esc_html__() \ esc_attr() should be used instead of __() for translation escaping.
classes/class-ep-dashboard.php line 378: esc_html__() \ esc_attr() should be used instead of __() for translation escaping.
classes/class-ep-dashboard.php line 504: Detected usage of a non-sanitized input variable: $_POST
classes/class-ep-dashboard.php line 683: Detected usage of a non-sanitized input variable: $_POST
features/protected-content/protected-content.php line 112: esc_html_e() \ esc_attr_e() should be used instead of _e() for translation escaping.
```

I have a PR coming shortly to address these.

PR #932, opened 2017-09-13
#931 - Late escaping per WordPress.com VIP review

jasonbahl:
Addresses #931


tlovett1:
Looks good. Thanks @jasonbahl !

PR #935, opened 2017-09-22
Harden the "Is single" check in WooCommerce feature

kallehauge:
It is currently possible to trigger the WooCommerce feature's arguments translation on single post/product queries which can cause trouble for some projects. One of the problems I've experienced is that the page should be triggered as a 404 page, since the requested product doesn't exist, but instead returns a HTTP 200 with results (I'm guess that the result varies depending on the individual theme - but I have experienced that I get a collection of products).

Example of Woo URL that can cause this issue: https://example.com/?post_type=product&p=12341234
The URL example satisfies the WooCommerce feature's single condition by not containing a `product` query_var.

### DIFF
--- a/features/woocommerce/woocommerce.php
+++ b/features/woocommerce/woocommerce.php
@@ -201,7 +201,7 @@ function ep_wc_translate_args( $query ) {
 	/**
 	 * Do nothing for single product queries
 	 */
-	if ( ! empty( $product_name ) ) {
+	if ( ! empty( $product_name ) || $query->is_single() ) {
 		return;
 	}
 


kallehauge:
@tlovett1 @ivankristianto I would appreciate if this could be accepted as a 2.4 milestone due to its simple fix vs high impact

ivankristianto:
+1 for this fix.

@kallehauge just out of curiosity, why you had to use that permalink? Woocommerce default permalink is ?product=product_slug

kallehauge:
@ivankristianto it's been a while so I cannot remember the direct cause. It might have been some logic from [FG Magento to WooCommerce](https://da.wordpress.org/plugins/fg-magento-to-woocommerce/) - it took a bit of time for me to make this PR after I actually found the scenario.

kallehauge:
@ivankristianto is there anything I need to do to get this merged?

kallehauge:
ping @ivankristianto :)

eugene-manuilov:
Hi @kallehauge! It looks good! Thank you!

PR #939, opened 2017-09-28
Check taxonomy object has the property fixed #881

### LINKED ISSUE #881: Trying to get property of non-object in  /elasticpress/classes/class-ep-api.php on line 690

ivankristianto:
This is safer way to check the property we want from the $taxonomy object.

### DIFF
--- a/classes/class-ep-api.php
+++ b/classes/class-ep-api.php
@@ -692,6 +692,11 @@ class EP_API {
 		$allow_hierarchy = apply_filters( 'ep_sync_terms_allow_hierarchy', false );
 
 		foreach ( $selected_taxonomies as $taxonomy ) {
+			// We check if the $taxonomy object as name property. Backward compatibility since WP_Taxonomy introduced in WP 4.7
+			if ( ! property_exists( $taxonomy, 'name' ) ) {
+				continue;
+			}
+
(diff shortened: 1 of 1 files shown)


tlovett1:
@ivankristianto did you test in pre-4.7?

ivankristianto:
@tlovett1 I did.
Test against WP 4.6 and 4.5 and looks good.
Also run unit testing in 4.6, it throw 1 error but not specific for this issue.

```
There was 1 failure:

1) EPTestSingleSite::testMetaQueryLike
Failed asserting that 1 matches expected 3.

/srv/www/mylocal/htdocs/wp-content/plugins/elasticpress/tests/test-single-site.php:2330
```

tlovett1:
Thanks, @ivankristianto !

Issue #883, opened 2017-07-05
sort orders by date


hrose1:
![sort](https://user-images.githubusercontent.com/19900773/27856881-d16bef00-616f-11e7-99a6-f40b3d7b3a68.png)

We are using the current elastic search version and sorting orders by date - e.g. when searching for a clients orders by email - does not work correctly.

ivankristianto:
@hrose1 you need to click the "date"/"Datum" column head

hrose1:
when clicking on the date column head it is ignoring the search input (email)
![bildschirmfoto 2017-07-06 um 06 35 49](https://user-images.githubusercontent.com/19900773/27895784-7e56cde8-6215-11e7-8c87-09136812afba.png)

ivankristianto:
@hrose1 can you install debug-bar-elasticpress and paste me the Elasticsearch Query Body.
Please check here on how to install: https://www.elasticpress.io/blog/2017/05/debugging/

hrose1:
this is the Query Body:
{
    "from": 0,
    "size": 75,
    "sort": [
        {
            "post_date": {
                "order": "asc"
            }
        }
    ],
    "query": {
        "function_score": {
            "query": {
                "bool": {
                    "should": [
                        {
                            "multi_match": {
                                "query": "f.borgel@t-online.de",
                                "type": "phrase",
                                "fields": [
                                    "post_title",
                                    "post_content",
                                    "post_excerpt",
                                    "meta._order_key.value",
                                    "meta._billing_company.value",
                                    "meta._billing_address_1.value",
                                    "meta._billing_address_2.value",
                                    "meta._billing_city.value",
                                    "meta._billing_postcode.value",
                                    "meta._billing_country.value",
                                    "meta._billing_state.value",
                                    "meta._billing_email.value",
                                    "meta._billing_phone.value",
                                    "meta._shipping_address_1.value",
                                    "meta._shipping_address_2.value",
                                    "meta._shipping_city.value",
                                    "meta._shipping_postcode.value",
                                    "meta._shipping_country.value",
                                    "meta._shipping_state.value",
                                    "meta._billing_last_name.value",
                                    "meta._billing_first_name.value",
                                    "meta._shipping_first_name.value",
                                    "meta._shipping_last_name.value",
                                    "meta._billing_full_name.value",
                                    "meta._shipping_full_name.value",
                                    "meta._wcpdf_invoice_number.value",
                                    "meta._wcpdf_formatted_invoice_number.value",
[...]

PR #940, opened 2017-09-28
Remove weighting by date in admin

lukaspawlik:
@tlovett1 this is enhancement for https://github.com/10up/ElasticPress/issues/883.

@ivankristianto suggested three enhancements however I am not sure if we need/should implement all of them.

Removing weighting results by recency should be definitely added for admin queries (that's why I put it in _Protected Content_ feature). I don't think we should always sort by __score_ as we will override admin pages sorting feature (when clicked on a sortable column). If there is no sorting set on edit.php page then __score_ sorting is used by default.

Last @ivankristianto suggestion about adding _min_score_ sounds quite promising however we would need to find some correct number and this may be very personal preference. I wouldn't add _min_score_ unless we have some idea about its value that could be used by default. 

Re core of https://github.com/10up/ElasticPress/issues/883 - I don't confirm that reported problems with sorting and filtering by email are reproducible.

### DIFF
--- a/features/protected-content/protected-content.php
+++ b/features/protected-content/protected-content.php
@@ -40,7 +40,7 @@ function ep_pc_post_types( $post_types ) {
 
 /**
  * Integrate EP into proper queries
- * 
+ *
  * @param  WP_Query $query
  * @since  2.1
  */
@@ -89,11 +89,18 @@ function ep_pc_integrate( $query ) {
 			$query->set( 'ep_integrate', true );
 		}
 	}
+
+	/**
+	 * Remove articles weighting by date in admin.
+	 *
+	 * @since 2.4
+	 */
+	remove_filter( 'ep_formatted_args', 'ep_weight_recent', 10 );
 }
 
 /**
  * Output feature box summary
- * 
+ *
  * @since 2.1
  */
 function ep_pc_feature_box_summary() {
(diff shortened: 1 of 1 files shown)


ivankristianto:
@lukaspawlik 
Order search results return fuzzy search. So when we sort it by date, it will sort by date, not by score anymore. 
I would suggest to have multiple sort order, by score and by date.

lukaspawlik:
Hi @ivankristianto,

Thanks for your reply.

I wonder if there is a point in adding extra sort order by _score_ here. Assuming that we still want have date order as more significant factor then _score_ would have an impact on results sort order for orders created at the same time (exactly the same time including seconds). If we change sort order to _score_ and _date_ then practically _date_ sort won't be taken into account as it would have an impact only on results with the same score which is less likely than having two orders with the same date.

It's not worth adding secondary _score_ sort order in my opinion.

tlovett1:
We should definitely not sort by _score is a sort method is explicitly set. If no sort method is, I don't understand why you wouldn't factor date into relevancy scores - shouldn't date weighting in theory result in more relevant results?

lukaspawlik:
@tlovett1 I think that date weighting doesn't always increase relevancy as it may promote result with lower initial relevancy over result with higher relevancy just because it's more recent. I think yhis is a great solution rather for news sites. When searching by for example last name: _Gomez_ it may score higher _Komez_ over _Gomez_ based only on their date (it's real example from my local). This tells me that date weighting not always helps relevancy (I'd expect to see _Gomez_ first). Additionally in our case if users want to sort filtered results by date they still can click on sort date icon to sort filtered results.

Issue #943, opened 2017-10-03
PHP Notice:  Trying to get property of non-object in /var/www/mysite.com/wp-content/plugins/elasticpress/features/woocommerce/woocommerce.php on line 274


ahmed-sigmalux:
Have you searched for similar issues before submitting this one?

Yes, but the issue was closed without resolution.

Is this a bug, question or feature request?

Bug.

Describe the issue you encountered:

Debug log shows the following errors pretty frequently:

`PHP Notice:  Trying to get property of non-object in /var/www/mysite.com/wp-content/plugins/elasticpress/features/woocommerce/woocommerce.php on line 274`

As far as I can tell, WordPress, WooCommerce, and ElasticPress work properly, i.e., searches return the expected results.

Current WordPress version:

4.8.2

Current ElasticPress version:

2.3.2

Current Elasticsearch version:

5.6.2

Where do you host your Elasticsearch server:

Self-hosted.

Other plugins installed (WooCommerce, Simple Redirect Manager, etc..):

WooCommerce, Yoast.

Steps to reproduce:

1. Install WordPress, WooCommerce, and ElasticPress
2. Enable and index WooCommerce products
3. Check debug.log

lukaspawlik:
Hi @ahmed-sigmalux,

Thanks for reporting this. I'll check this and create PR if needed.

lukaspawlik:
@ahmed-sigmalux,

PR created and waiting for review.

ahmed-sigmalux:
Thanks for the prompt reply and fix!

lukaspawlik:
This has been merged to develop branch and will be released in 2.4

Closing.

PR #945, opened 2017-10-06
Add sanity checks before accessing object properties

lukaspawlik:
@tlovett1,

This is a fix for #943 - I couldn't reproduce this on my local however it's possible that functions `get_term_by` and `get_term` may not return valid data as expected. Adding extra sanity checks before accessing their results won't do any harm.


lukaspawlik:
@tlovett1 Ready for your review.

PR #946, opened 2017-10-06
Fix sync error when WooCommerce 2.x installed

lukaspawlik:
@tlovett1 This fixes as issue introduced in https://github.com/10up/ElasticPress/pull/917. https://github.com/10up/ElasticPress/pull/917 doesn't work with WooCommerce 2.x that doesn't use  `WC_Order_Item_Product` as `$product_item` but raw array.


tlovett1:
Nice work!

kallehauge:
Great catch @lukaspawlik 👍 

@tlovett1 this is a bit off-topic for the PR (sorry): but I've submitted a bit of code by now and I would like to be more efficient to get my work into the project, faster, since it is something we're using for several clients. What is the supported version philosophy? (Ex: for how long will Woo 2.x be supported? And what WP version?)

Issue #938, opened 2017-09-27
Yoast SEO conflict when activated no WooCommerce order results shown.


RonalddeRooij:
Have you searched for similar issues before submitting this one?
Yes without finding a similar issue.

Is this a bug, question or feature request?
Bug

Describe the issue you encountered:
If plug-in Yoast SEO is active, WooCommerce Order search doesn't show any results. Turning the plug-in off, it works perfectly fine. Installed Debug with Debug ElasticPress:

If Yoast SEO is not active:
Total ElasticPress Queries: 1 

If Yoast SEO is active:
Total ElasticPress Queries: 2

First Query Args has the search keyword however the second query is missing the search keyword. 

Current WordPress version: 4.8.2
Current ElasticPress version: 2.3.2
Current Elasticsearch version: 5.6.1
Current Yoast SEO: 5.5

Where do you host your Elasticsearch server:
Sohosted

Other plugins installed (WooCommerce, Simple Redirect Manager, etc..):
Many plug-ins installed turned them all off and activated them one by one the issue appears if Yoast SEO is active.

Steps to reproduce:

1. Yoast SEO plug-in active
2. Search WooCommerce orders (shows result number, without the results itself)
3. Turn of Yoast SEO plug-in, everything works fine

Screenshots, if needed:
![extra-information](https://user-images.githubusercontent.com/16287765/30910261-467c711a-a384-11e7-8da5-b3c0fa21abf6.jpg)

lukaspawlik:
Thanks @RonalddeRooij for reporting this. I'll check it and I'll get back to you as soon as possible.

RonalddeRooij:
Hi, thanks for picking this up. Do you have any update on this issue?

lukaspawlik:
Hello,

I don't have any update on this ticket yet. I should have something near the end of the week.

lukaspawlik:
Hello @RonalddeRooij,

I was able to inspect this and identify issue. I've prepared fix for this in pull request https://github.com/10up/ElasticPress/pull/947.

Basically this issue occurs only on WooCommerce 2.x. Version 3.x is not affected. I suggest updating WooCommerce to version 3.x or waiting until #947 is merged and new ElasticPress version released.

RonalddeRooij:
@lukaspawlik Thank you very much for investigating the issue. I will update WooCommerce from 2.x to version 3.x to fix my issue.

PR #947, opened 2017-10-06
Disallow duplicated Elasticsearch requests on WooCommerce Orders page

lukaspawlik:
@tlovett1 This is fix for #938 when duplicated Elasticsearch queries are made on WooCommerce (2.x only) orders page. Yoast is calling global `$wp_query->get_posts()` that tries to run ES query again. First thing that it is abundant as there is no point in making the same query again. The second issue is that first `$wp_query->get_posts()` call sets up `$_GLOBAL['post']` object to first post from ES response and this forces all calls to `get_post()` to return this post always (even if `get_post()` is called with numeric parameter).  Duplicated `$wp_query->get_posts()` makes `ep_wc_search_order` (action hook for `parse_query` action) to incorrectly execute following logic:

```php
$search_key_safe = str_replace( array( 'Order #', '#' ), '', wc_clean( $_GET['s'] ) );

	$order = wc_get_order( absint( $search_key_safe ) );

	//If the order doesn't exist, fallback to other fields
	if ( ! $order ) {
		unset( $wp->query_vars['post__in'] );
		$wp->query_vars['s'] = $search_key_safe;
	} else {
		//we found the order. don't query ES
		unset( $wp->query_vars['s'] );
		$wp->query_vars['post__in'] = array( absint( $search_key_safe ) );
	}
```
`absint( $search_key_safe )` returns `0` (for all non numeric search terms) and `wc_get_order()` calling `get_post()` (with an argument of value `0`) pulls always first post from previous ES request. Having `$order` non empty code goes to `else` statement and unsets search term from ES query and adds `post__in` query argument with a value of `0` that makes the second call returning empty results and overwrites main `$wp_query` results with empty set. My fix also makes sure that if `absint( $search_key_safe )` is `0` then `$order` is always false as there can't be a post with ID `0`.

### DIFF
--- a/features/woocommerce/woocommerce.php
+++ b/features/woocommerce/woocommerce.php
@@ -163,6 +163,37 @@ function ep_wc_whitelist_taxonomies( $taxonomies, $post ) {
 }
 
 /**
+ * Disallow duplicated ES queries on Orders page.
+ *
+ * @since 2.4
+ *
+ * @param array    $value Original filter values.
+ * @param WP_Query $query WP_Query
+ *
+ * @return array
+ */
+function ep_wc_disallow_duplicated_query( $value, $query ) {
+	global $pagenow;
+
+	/**
+	 * Make sure we're on edit.php in admin dashboard.
+	 */
+	if ( 'edit.php' !== $pagenow || ! is_admin() || 'shop_order' !== $query->get( 'post_type' ) ) {
+		return $value;
+	}
+
+	/**
+	 * Check if EP API request was already done. If request was sent return its results.
+	 */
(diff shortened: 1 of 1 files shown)


lkraav:
This issue is real on our WC 2.6 shop. Disabling Yoast SEO makes Orders search work, but obv. that's not a solution.

@tlovett1 is this PR too complex to integrate?

lkraav:
PS I just managed to get Orders search to show results by applying a5de1f5 alone.

tlovett1:
Hey guys, we will get this one prepared/merged for 2.5.

PR #949, opened 2017-10-10
Feature/recursionguard

daveross:
ElasticPress calls the `the_content` filter when indexing content. But, theme & plugin out there code calls `the_content` independently when they render content in some contexts, spiraling into recursion until PHP exhausts its resources.

This was tested independently, but not in the context of a client that was seeing this issue. The engineers on that project haven't had time. Still, I wanted to get this PR out there.

### DIFF
--- a/classes/class-ep-api.php
+++ b/classes/class-ep-api.php
@@ -612,7 +612,7 @@ class EP_API {
 			'post_date_gmt'     => $post_date_gmt,
 			'post_title'        => $this->prepare_text_content( get_the_title( $post_id ) ),
 			'post_excerpt'      => $this->prepare_text_content( $post->post_excerpt ),
-			'post_content'      => $this->prepare_text_content( self::apply_filters_recursion_guarded( 'the_content', $post->post_content ) ),
+			'post_content'      => $this->prepare_text_content( self::apply_filters_recursion_guarded( 2, 'the_content', $post->post_content ) ),
 			'post_status'       => $post->post_status,
 			'post_name'         => $post->post_name,
 			'post_modified'     => $post_modified,
--- a/classes/class-ep-api.php
+++ b/classes/class-ep-api.php
@@ -532,9 +532,9 @@ class EP_API {
 	 */
 	private function apply_filters_recursion_guarded( $max_levels = 1, $tag, $value ) {
 
-		$rg = new EP_RecursionGuard( $max_levels, 'the_content' );
+		$rg = new EP_RecursionGuard( $max_levels, $tag );
 		try {
-			$value = apply_filters( 'the_content', $value );
+			$value = apply_filters( $tag, $value );
 		} catch ( EP_RecursionTooDeepException $e ) {
 			$value = $e->value();
 		} finally {
--- a/classes/class-ep-api.php
+++ b/classes/class-ep-api.php
@@ -522,6 +522,31 @@ class EP_API {
 	}
 
 	/**
+	 * Call the functions added to a filter hook but with protection against deep recursion.
+	 *
+	 * @see           apply_filters() This function works effectively the same, but with extra prevention against recursion.
+	 * @param  int    $max_levels The maximum level of recursion permitted.
+	 * @param  string $tag The name of the filter hook.
+	 * @param  mixed  $value The value on which the filters hooked to `$tag` are applied on.
+	 * @return mixed  The filtered value after all hooked functions are applied to it.
+	 */
+	private function apply_filters_recursion_guarded( $max_levels = 1, $tag, $value ) {
+
+		$rg = new EP_RecursionGuard( $max_levels, 'the_content' );
+		try {
+			$value = apply_filters( 'the_content', $value );
+		} catch ( EP_RecursionTooDeepException $e ) {
+			$value = $e->value();
+		} finally {
+			$rg->cleanup();
+			unset( $rg );
+		}
+
+		return $value;
+
+	}
+
+	/**
 	 * Prepare a post for syncing
 	 *
 	 * @param int $post_id
@@ -587,7 +612,7 @@ class EP_API {
 			'post_date_gmt'     => $post_date_gmt,
 			'post_title'        => $this->prepare_text_content( get_the_title( $post_id ) ),
 			'post_excerpt'      => $this->prepare_text_content( $post->post_excerpt ),
-			'post_content'      => $this->prepare_text_content( apply_filters( 'the_content', $post->post_content ) ),
+			'post_content'      => $this->prepare_text_content( self::apply_filters_recursion_guarded( 'the_content', $post->post_content ) ),
 			'post_status'       => $post->post_status,
 			'post_name'         => $post->post_name,
 			'post_modified'     => $post_modified,
--- a/classes/class-ep-recursionguard.php
+++ b/classes/class-ep-recursionguard.php
@@ -1,0 +1,83 @@
+<?php
+
+/**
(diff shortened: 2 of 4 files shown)


classes/class-ep-api.php
```diff
@@ -522,6 +522,31 @@ public function put_mapping() {
...
 	/**
+	 * Call the functions added to a filter hook but with protection against deep recursion.
+	 *
+	 * @see           apply_filters() This function works effectively the same, but with extra prevention against recursion.
+	 * @param  int    $max_levels The maximum level of recursion permitted.
+	 * @param  string $tag The name of the filter hook.
+	 * @param  mixed  $value The value on which the filters hooked to `$tag` are applied on.
+	 * @return mixed  The filtered value after all hooked functions are applied to it.
+	 */
+	private function apply_filters_recursion_guarded( $max_levels = 1, $tag, $value ) {
```
lukaspawlik: @daveross Can we rearrange function arguments to move all _optional_ arguments at the end? This way they have to be always specified.

classes/class-ep-api.php
```diff
@@ -522,6 +522,31 @@ public function put_mapping() {
...
+
+		$rg = new EP_RecursionGuard( $max_levels, $tag );
+		try {
+			$value = apply_filters( $tag, $value );
+		} catch ( EP_RecursionTooDeepException $e ) {
+			$value = $e->value();
+		} finally {
```
lukaspawlik: @daveross _finally_ was introduced in PHP 5.4 and we should be PHP 5.2 compatible the same way as WordPress is. Can you please remove _finally_ block?

classes/class-ep-api.php
```diff
@@ -587,7 +612,7 @@ public function prepare_post( $post_id ) {
 			'post_date_gmt'     => $post_date_gmt,
 			'post_title'        => $this->prepare_text_content( get_the_title( $post_id ) ),
 			'post_excerpt'      => $this->prepare_text_content( $post->post_excerpt ),
-			'post_content'      => $this->prepare_text_content( apply_filters( 'the_content', $post->post_content ) ),
+			'post_content'      => $this->prepare_text_content( self::apply_filters_recursion_guarded( 2, 'the_content', $post->post_content ) ),
```
lukaspawlik: @daveross Shouldn't we use here `$this` instead of `$self`?

classes/class-ep-recursionguard.php
```diff
@@ -0,0 +1,83 @@
...
+	/**
+	 * Constructor. Sets up a guarded version of the filter.
+	 *
+	 * @param integer $max_levels The maximum level of recursion permitted.
+	 * @param string  $tag The name of the filter hook.
+	 */
+	public function __construct( $max_levels = 1, $tag ) {
+		$this->max_levels = $max_levels;
+		$this->tag = $tag;
+		add_filter( $this->tag, $this->get_fn_ref(), 1 );
```
lukaspawlik: @daveross Is our intention to set priority to *1* here or maybe it's number of accepted arguments here? If we set *1* priority here we need to use it in `cleanup()` method too otherwise removing filter won't work (and we'll have several _the_content_ EP_RecursionGuard handlers (and they will start throwing exceptions).

classes/class-ep-recursionguard.php
```diff
@@ -0,0 +1,83 @@
...
+
+	/**
+	 * Returns a reference to this instance's filter handler function.
+	 * Pass the result of this function to add_filter() so it can
+	 * later be removed by remove_filter().
+	 *
+	 * @return Callable
+	 */
+	public function get_fn_ref() {
+		return [ $this, 'handle' ];
```
lukaspawlik: @daveross Can we change this to `array( $this, 'handle' )` as array short syntax was introduced in PHP 5.4 and we want to stay PHP 5.2 compatible (as WordPress is).

classes/class-ep-recursionguard.php
```diff
@@ -0,0 +1,83 @@
...
+	 */
+	private $tag;
+
+	/**
+	 * Constructor. Sets up a guarded version of the filter.
+	 *
+	 * @param integer $max_levels The maximum level of recursion permitted.
+	 * @param string  $tag The name of the filter hook.
+	 */
+	public function __construct( $max_levels = 1, $tag ) {
```
lukaspawlik: @daveross Can we move optional arguments to the end, after mandatory arguments?

classes/class-ep-recursionguard.php
```diff
@@ -0,0 +1,83 @@
...
+		return $value;
+	}
+
+	/**
+	 * Clean up after use. Removes the filter handler.
+	 *
+	 * @return void
+	 */
+	public function cleanup() {
+		remove_filter( $this->tag, $this->get_fn_ref() );
```
lukaspawlik: @daveross I mentioned earlier about setting *1* priority. If our intention there is to use higher priority we need it here too as `remove_filter` won't work.

lukaspawlik:
@tlovett1 I think I found at least one live scenario that runs into _the_content_ recursion issue. This loop ends at some point but it definitely fires _the_content_ filter hooks from one of _the_content_ hooks.

Sample scenario is liveblog. When article is marked as a liveblog it stores their messages/entries in article's comments. When visitors go to article then liveblog starts rendering its content. Liveblog hooks into _the_content_ filter and then it pulls all comments, with type _liveblog_, to render them. Liveblog entry can contain tweet or any other embed and this is rendered via `$wp_embed->autoembed( $content );` call. This call at some point tries to cache oembed HTML in article meta field (wp-includes/class-wp-embed.php:259 in WordPress 4.8.2). ElasticPress plugin adds its own hooks as below:

```php
add_action( 'updated_postmeta', array( $this, 'action_queue_meta_sync' ), 10, 4 );
add_action( 'added_post_meta', array( $this, 'action_queue_meta_sync' ), 10, 4 );
```

Given above when oembed HTML is cached in liveblog article meta then ElasticPress triggers `\EP_Sync_Manager::action_queue_meta_sync` that calls `\ep_prepare_post (\EP_API::prepare_post)` and this function fires `apply_filters( 'the_content', $post->post_content )` that enters to liveblog _the_content_ filter hook that actually is currently running. This loop ends as oembed HTML is not cached during every request however it's possible that any other oembed handler/shortcode can be updating post meta fields always and this would run us into infinite loop.

I see this recursion issue as tricky issue as on one hand we could not run `apply_filters( 'the_content', $post->post_content );` when `ep_prepare_post` is called from `\EP_Sync_Manager::action_queue_meta_sync` as filtered content doesn't seem to be used anywhere.

We do in `\EP_Sync_Manager::action_queue_meta_sync` as below:

```php
$prepared_post = ep_prepare_post( $object_id );

				// Make sure meta key that was changed is actually relevant
				if ( ! isset( $prepared_post['meta'][$meta_key] ) ) {
					return;
				}

				$this->sync_post_queue[$object_id] = true;
```

So basically we run `ep_prepare_post` and check if meta that triggered this hook should be synced with Elasticsearch or not. It happens in `shutdown` action. This action runs `ep_prepare_post` again in `\EP_Sync_Manager::sync_post` and then `post_content` is actually used for sync process.

Easier solution for this recursion issue, caused by triggering it by post meta fields updates, would be adding second parameter to `ep_prepare_post( $post_id, $filter_content = true );` that would be responsible for rendering content or not. Proof of concept available here: https://github.com/10up/ElasticPress/pull/957

@tlovett1 What do you think about this?

tlovett1:
Closing this one. Feel free to reopen.

PR #984, opened 2017-12-19
Hide EP_HOST value on settings page

tomdxw:
Principle of least privilege: WordPress admin users don't need to know
this.

---

In our case, we host WP sites. The clients who use them shouldn't know the username, password, and hostname of the Elasticsearch instance.

Great plugin by the way. Thanks.

### DIFF
--- a/includes/settings-page.php
+++ b/includes/settings-page.php
@@ -35,7 +35,9 @@ if ( defined( 'EP_IS_NETWORK' ) && EP_IS_NETWORK ) {
 				<tr>
 					<th scope="row"><label for="ep_host"><?php esc_html_e( 'Elasticsearch Host', 'elasticpress' ); ?></label></th>
 					<td>
-						<input <?php if ( defined( 'EP_HOST' ) && EP_HOST ) : ?>disabled<?php endif; ?> placeholder="http://" type="text" value="<?php echo esc_url( ep_get_host() ); ?>" name="ep_host" id="ep_host">
+						<?php if ( apply_filters( 'ep_admin_show_host', true ) ) : ?>
+							<input <?php if ( defined( 'EP_HOST' ) && EP_HOST ) : ?>disabled<?php endif; ?> placeholder="http://" type="text" value="<?php echo esc_url( ep_get_host() ); ?>" name="ep_host" id="ep_host">
+						<?php endif ?>
 						<?php if ( defined( 'EP_HOST' ) && EP_HOST ) : ?>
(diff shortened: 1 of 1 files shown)


eugene-manuilov:
Hi @tomdxw! Thanks for your contribution!

Your change looks good but if we add it, it will hide EP_HOST value for everyone, that is not good. I think it would be better to update [`get_host` function](https://github.com/10up/ElasticPress/blob/develop/classes/class-ep-config.php#L88) and apply filters on the `$host` variable before returning it. In this case, you will be able to use a custom hook to hide it on your sites and other people will still see it. Does it make sense to you? If so, do you mind to update your PR?

tlovett1:
I'd probably just hide the input box.

tomdxw:
>  it will hide EP_HOST value for everyone, that is not good.

It will only hide it when the value is set via `define('EP_HOST', ...);`.

I'm sorry, I don't understand. Why would anybody need to see the value? They can't change it if it's been set as a constant.

> I'd probably just hide the input box.

You mean hide it when `(defined('EP_HOST') && EP_HOST)` is true? Or use a filter or a constant to hide it?

eugene-manuilov:
> I'm sorry, I don't understand. Why would anybody need to see the value? They can't change it if it's been set as a constant.

I realize the value may not be useful to you, but for many it's the easiest way to check the host settings without digging into wp-config.php. It's also been part of the plugin for a long time. This isn't something we can just hide at this point. We recommend hiding with CSS if needed.

tomdxw:
> I realize the value may not be useful to you, but for many it's the easiest way to check the host settings without digging into wp-config.php. It's also been part of the plugin for a long time. This isn't something we can just hide at this point.

Okay, fair enough.

> We recommend hiding with CSS if needed.

Hiding with CSS doesn't fix the problem. If we hide it with CSS then they can see it by using "view source".

What about the following? Would this be acceptable?

```
<?php if (!(defined('EP_HIDE_HOST') && EP_HIDE_HOST)) : ?>
    <input <?php if ( defined( 'EP_HOST' ) && EP_HOST ) : ?>disabled<?php endif; ?> placeholder="http://" type="text" value="<?php echo esc_url( ep_get_host() ); ?>" name="ep_host" id="ep_host">
<?php endif ?>
```

eugene-manuilov:
I think filter works better here. Do you mind to update your PR to use a custom filter instead of `defined( 'EP_HIDE_HOST' ) ... `?

tomdxw:
Sure. I've updated the PR now. Let me know if this is okay.

tlovett1:
Works for me! Thanks.

PR #991, opened 2018-01-04
Fixed widgets setup issue

eugene-manuilov:
Feature setup has to be executed as soon as possible to be able to hook actions and filters which are called on `init` action.


tlovett1:
@eugene-manuilov can you reopen against master and remove the readme update? Thanks!

PR #994, opened 2018-01-04
Updated javascript file of the autosuggest feature to use event.target

eugene-manuilov:


tlovett1:
Eugene, can you reopen against develop and remove the readme change?

Issue #987, opened 2017-12-22
FIXED : Sync unindex medias 


Antoinebr:
### Have you searched for similar issues before submitting this one?

Yes 

### Is this a bug, question or feature request?

bug / question

### Describe the issue you encountered:

When I upload an image this image is indexed but if I launch a sync the image is no indexed anymore. 

### Current WordPress version:

4.9.1

### Current ElasticPress version:

2.4

### Current Elasticsearch version:
5.6.5

### Where do you host your Elasticsearch server:

locally ( docker )

### Other plugins installed (WooCommerce, Simple Redirect Manager, etc..):

No

### Steps to reproduce:

1. Upload an image
2. Check the ES index
3. Run a sync from the plugin
4. the image is gone from the index

An other example : 

```php 

add_filter('ep_indexable_post_types', function(){

	return "attachment"; // 0 items synced out of 10
       	return "page"; // 1 items synced out of 1

});

```

Antoinebr:
Fixed ! 

This happened because the plugin is looking for ```publish``` as post status and medias are always ```inherit```

So I fixed the issue with 

```php 

add_filter('ep_indexable_post_status', function(){

	return array( 'publish', 'inherit' );

} );

```

eugene-manuilov:
Thanks, @Antoinebr! I have created a pull request for it.

tlovett1:
Shouldn't this be handled here @eugene-manuilov ? https://github.com/10up/ElasticPress/blob/develop/features/documents/documents.php#L170

eugene-manuilov:
@tlovett1, the **Documents** feature is intended to search things in textual documents, isn't it? Unfortunately, it means that it skips regular attachments which may contain useful and searchable information. 

My fix makes sure that we include **inherit** post status if we have **attachment** post type in the indexable types list. It allows to properly index attachments even if **Documents** feature is disabled.

eugene-manuilov:
Re-created pull request against develop branch. #1007

tlovett1:
I think this makes sense, but the section of the document feature that adds `inherit` to syncable post statuses should be removed. I'll handle that.

tlovett1:
I don't think this makes sense. If someone wants to enable attachment indexing, they just have to filter two filters.

PR #1007, opened 2018-01-24
Added inherit status to support attachments indexing.

eugene-manuilov:


tlovett1:
@eugene-manuilov this just results in all attachments being indexed no matter what. I think #987 needs a new solution.

Ritesh-patel:
@tlovett1 do we need to work on this #987 is closed?

PR #1039, opened 2018-03-15
Fix sync, if change _edit_lock meta field (just open and close post)

Eventusrus:


classes/class-ep-sync-manager.php
```diff
@@ -102,6 +102,10 @@ public function action_queue_meta_sync( $meta_id, $object_id, $meta_key, $meta_v
 		if ( ! empty( $importer ) ) {
 			return;
 		}
+
+		if ($meta_key === '_edit_lock') {
```
csloisel: This should have padded spaces and use [yoda syntaz](https://make.wordpress.org/core/handbook/best-practices/coding-standards/php/#yoda-conditions) to meet WP coding standards:

`if ( '_edit_lock' === $meta_key ) {`

csloisel:
After digging into this, an explicit check for this key shouldn't be necessary. I'm having trouble finding a scenario where this meta update would trigger a sync. The plugin already contains logic to prevent protected keys from triggering a sync: https://github.com/10up/ElasticPress/blob/2.4.2/classes/class-ep-api.php#L802

The WP core method `is_protected_meta()` should return true and prevent this meta key from being set in the prepared post object and in turn prevent a sync. This can be overridden using the `'ep_prepare_meta_allowed_protected_keys'` filter in the plugin, or by using the `'is_protected_meta'` filter in core, but it is disallowed by default.

tlovett1:
I guess if you did have protected meta enabled, you still wouldn't want this to trigger a sync.

tlovett1:
@Eventusrus can you provide anymore info?

Eventusrus:
Sorry, much work, less time :)

@tlovett1 it's true. 
@csloisel , yes meta is not sync, but everyone is implemented prepare_post and loaded full post object. I think, this situation is bad for performance.

allan23:

PR #1044, opened 2018-03-25
Implement --post-ids CLI option to index only specific posts

dotancohen:
This pull request implements a `--post-ids` CLI option to index only specific posts, as described in 
issue 886.

Example usage:
`$ wp elasticpress index --post-ids=52,51`


csloisel:
I've tested and confirmed that this works as expected.

tlovett1:
Awesome. We will queue this up for the next major release.

PR #1056, opened 2018-04-12
Fix #1030: ep_prepare_post getting called too many times on post update

### LINKED ISSUE #1030: prepare_post() in class-ep-api.php called 49 times when creating a shop_order post via WC Rest API
Have you searched for similar issues before submitting this one? Yes

Is this a bug, question or feature request? bug, probably will end up being a feature

Describe the issue you encountered: We are using 'ep_post_sync_args' filter to add line_items and coupon data to the orders (type: shop_order). While a doing full we noticed that 'ep_post_sync_args' was called 49 times when creating a order via WC Rest using the following request,
/wc/v2/orders
request body:{"customer_id":"2","status":"pending"}
We tried to use the 'ep_pre_index_post' filter but found out that this filter does not get called in when doing a full sync via the Woocommerce / ElasticPress dashboard.

I believe the solution would be to add the 'ep_pre_index_post' or a similar filter to line: 564 in class-ep-dashboard.php after the line, $post_args = ep_prepare_post( get_the_ID() );

Current WordPress version: 4.9.4

Current ElasticPress version: 2.4.2

Current Elasticsearch version: 3.3.2

Where do you host your Elasticsearch server: searchly.com

Other plugins installed (WooCommerce, Simple Redirect Manager, etc..): Woocommerce Stripe

Steps to reproduce:

1. Add a echo or error_log message to the function applying the filter, ep_post_sync_args
2. Send the Woocommerce Rest request via, /wc/v2/orders
request body:{"customer_id":"2","status":"pending"}
**Note: I suspect this does not have to be a Woocommerce request**
3. Count the number of times the error_log message appears in the log file.

Screenshots, if needed:

Ritesh-patel:


allan23:
@oscarssanchez can you review as this is a PR for #1030.

oscarssanchez:
Hi @allan23 , 

This is not working in my testing, as i get prepare_post() called multiple times when updating a shop_order meta from woocommerce, though it was called about 3 times for me, not 49 as described originally in #1030 with and without this PR. 

I'm wondering if this was still in progress as I don't see reviewers assigned previously or a description for this PR.

allan23:
Closing down this PR to review original issue.

PR #1062, opened 2018-04-23
Ensure that element exists before accessing it.

dotancohen:
This patch ensures that the aggregation exists before attempting to access it. In my case, we call the aggregations something other than `terms` so the code below was throwing errors.

Note that a better fix would be to iterate over all aggregations and store them separately, however that would require also changing the places which read from `$GLOBALS['ep_facet_aggs']`.


tlovett1:
Probably should be:

```php
if ( ! empty( $response['aggregations'] ) && ! empty( $response['aggregations']['terms'] ) ) {
```

Thoughts?

dotancohen:
There is no need to check `$response['aggregations'] )` first. `empty()` is a language construct, not a function, and is safe to use on undefined elements. Otherwise even `empty( $response['aggregations']['terms'] )` would not be safe!

PR #1074, opened 2018-05-01
ep_is_indexing_wpcli is returning true when WP-CLI index is running

ravichandrach1:
Fix `is_indexing_wpcli` method to check for `ep_wpcli_sync` instead of `ep_index_meta`


tlovett1:
@allan23 can you test?

PR #1084, opened 2018-05-12
return the right value for found_posts

pierrebuet:
When LIMITS are set in the query, we use the filter found_posts_query to return an empty string. $wpdb->get_var('') returns then the last query, which can be anything and put it in found_posts. Later in the set_found_posts function, when calculating the pagination, the function fails if found_posts as a rsult of the $wpdb->get_var('') was not numeric.

Using the filter found_posts, if the query was using limits, we can replace the found_posts value with the right one, as the number of posts in the cached query of Elasticpress


pierrebuet:
@allan23 : I'm really sorry, but I have unfortunately no idea what a unit test is and how to do it. I can only tell you since LIMITS in the original query are the root cause of the issue, that I was testing on my woocommerce system different query types and they all worked with my commit. Sorry for this.

PR #1094, opened 2018-06-11
Exclude password protected posts in search results. 

### LINKED ISSUE #1083: Hide password protected posts from search results
Have you searched for similar issues before submitting this one? Yes.

Is this a bug, question or feature request? Bug or feature request, depending on how you look at it.

WordPress's default search hides password protected posts from logged out users.

ElasticPress doesn't hide password protected posts from logged out users, and doesn't record the password protected status in Elasticsearch so there's no easy way to hide the password protected posts.

Current WordPress version:

4.9.5

Current ElasticPress version:

2.5.2

5.6.5

n/a

Other plugins installed (WooCommerce, Simple Redirect Manager, etc..):

None.

Steps to reproduce:

1. Create a password protected post
2. Search for it as a logged out user and see that it's not in the results
3. Activate ElasticPress and sync
4. Search for the post as a logged out user and see that it's in the results

allan23:
This PR adds post meta to posts that are password protected. It then alters the public query to ElasticSearch by checking against this meta value to prevent password protected posts from appear in search results.

This approach seemed to be the least intrusive as I did not want to alter mappings or prevent the post from being indexed completely, which would not allow it to show in the WP admin if protected content is activated.

This PR fixes #1083

### DIFF
--- a/classes/class-ep-api.php
+++ b/classes/class-ep-api.php
@@ -794,6 +794,11 @@ class EP_API {
 	public function prepare_meta( $post ) {
 		$meta = (array) get_post_meta( $post->ID );
 
+		// If post is password protected, save post meta saying so.
+		if ( ! empty( $post->post_password ) ) {
+			$meta['password_protected'] = true;
+		}
+
 		if ( empty( $meta ) ) {
 			return array();
 		}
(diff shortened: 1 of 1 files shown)


tlovett1:
Looks solid.

PR #1096, opened 2018-06-12
Revert "Exclude password protected posts in search results. "

tlovett1:
Reverts 10up/ElasticPress#1094


lekkerduidelijk:
Can I ask why this was reverted? We're looking into using ElasticPress and this is functionality we rely on. Thanks!

Related #1083 #1094

lekkerduidelijk:
@tlovett1 Did you see my comment?

PR #1117, opened 2018-08-03
Configure Index Prefix on Settings Page

allan23:
Allow index prefix to be configured on settings page instead of just within wp-config.php.

### DIFF
--- a/classes/class-ep-config.php
+++ b/classes/class-ep-config.php
@@ -148,7 +148,7 @@ class EP_Config {
 		} elseif ( ep_is_epio() ) {
 			$credentials = ep_sanitize_credentials( get_option( 'ep_credentials', false ) );
 		} else {
-			$credentials = [];
+			$credentials = [ 'username' => '', 'token' => '' ];
 		}
 
 		if ( ! is_array( $credentials ) ) {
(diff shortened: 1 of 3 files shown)


allan23:
Using JavaScript, verify that the endpoint is ElasticPress.io. If so, then show the prefix field.

allan23:
@tott and @oscarssanchez, can you please review?

classes/class-ep-config.php
```diff
@@ -88,6 +113,51 @@ public function get_host() {
...
+
+		if ( defined( 'EP_IS_NETWORK' ) && EP_IS_NETWORK && ep_is_epio() ) {
+			$credentials = ep_sanitize_credentials( get_site_option( 'ep_credentials', false ) );
+		} elseif ( ep_is_epio() ) {
+			$credentials = ep_sanitize_credentials( get_option( 'ep_credentials', false ) );
+		} else {
+			$credentials = [];
+		}
+
+		if ( ! is_array( $credentials ) ) {
```
brandwaffle: @allan23 if line 151 is executed then this section will never set the blank array keys. Is this intended behavior?

allan23: It is not. Great catch!

allan23:
@brandwaffle PR updated

PR #1124, opened 2018-08-21
Post status per type

dotancohen:
Allow setting the post_status per run, to enable indexing different post types with different status.

Right now my projects are doing workarounds to enable indexing 'pending' items of a specific custom post type only. This feature will make it much easier.


allan23:

Issue #1133, opened 2018-09-25
Post type being indexed when it shouldn't be


jonathanstegall:
Have you searched for similar issues before submitting this one?

Yes

Is this a bug, question or feature request?

Possibly a bug

Describe the issue you encountered:

I have a filter like this:

```
add_filter( 'ep_indexable_post_types', 'indexable_post_types' );
function indexable_post_types( $post_types ) {
	$non_indexable_types = array(
		'partner_offer' => 'partner_offer',
		'wp_log'        => 'wp_log',
		'message'       => 'message',
	);
	$post_types          = array_diff_assoc( $post_types, $non_indexable_types );
	error_log( 'types is ' . print_r( $post_types, true ) );
	return $post_types;
}
```

The log result here is all of our post types except the three in the array above.

However, I'm noticing that when I search, there are `wp_log` post types in my search results. I believe these are all newer posts - that is, our index seems to have skipped the pre-existing log posts, but as new ones are created it seems to be indexing them.

Current WordPress version:

4.9.8

Current ElasticPress version:

2.6.1

Current Elasticsearch version:

Locally I'm running 6.2.4, and remotely I'm running 6.3.1. This behavior happens in both environments.

Where do you host your Elasticsearch server:

We have a single Linux server.

Other plugins installed (WooCommerce, Simple Redirect Manager, etc..):

We do have a significant number of other plugins. The wp_log post type is created by the [WP Logging](https://pippinsplugins.com/wp-logging/) library though.

Steps to reproduce:

1. Enable Search and Protected Content in ElasticPress settings.
2. Filter the indexable post types to remove a type.
3. Create new records of that type.
4. These records show up in search results.

Screenshots, if needed:

brandwaffle:
@jonathanstegall thanks for reporting this!

I did a little digging and it looks like the Protected Content module does a get_post_types() call during a filter that it runs https://github.com/10up/ElasticPress/blob/a24d36ad4ea02ca971e5192b7160b806a7947411/features/protected-content/protected-content.php#L32. If your filter runs before this one, that means that Protected Content is getting the list of all public post types (including wp_log) and essentially "resetting" the indexable post types regardless of what earlier filters are doing.

If this is the case, it's definitely a bug. Can you try firing your filter later than the Protected Content one (set a priority of 11 or higher)? That should temporarily resolve the issue, and we'll get started on a bugfix.

jonathanstegall:
@brandwaffle thanks for your help!

[...]

PR #1135, opened 2018-09-28
Filter prefiltered post_types when protected content feature enabled.

oscarssanchez:
Hi @allan23 , @brandwaffle ,

Could you please use this PR which addresses issue #1133 ?

This commit sets a parameter when calling `get_post_types()` to get all non public post_types, and merges them with the previously filtered post_types in `ep_indexable_post_types ` , in case post_types were filtered before` ep_pc_post_types()`. Probably `array_unique()` is not needed anymore, but happy to read your feedback.

Thanks!

### DIFF
--- a/features/protected-content/protected-content.php
+++ b/features/protected-content/protected-content.php
@@ -32,10 +32,12 @@ function ep_pc_setup() {
 function ep_pc_post_types( $post_types ) {
 	$pc_post_types = get_post_types( array( 'public' => false ) );
 
-	// We don't want to deal with nav menus
-	unset( $pc_post_types['nav_menu_item'] );
+	// We don't want to deal with nav menus if there are
+	if ( $pc_post_types['nav_menu_item'] ) {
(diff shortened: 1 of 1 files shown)


oscarssanchez:
Hi @brandwaffle ,

I don't think we are overwriting what other filters did previously to $post_types.

In $pc_post_types = get_post_types( array( 'public' => false ) ); we make sure to only get the post types that are not publicly queryable, then remove $pc_post_types['nav_menu_item']

The $post_types that have been passed before do not get affected by this.

Finally, we merge both $pc_post_types and $post_types into one array and return it. 

I tested this by filtering out a post type with ep_indexable_post_types and worked in my testing.

I don't think we can use $post_types as it is passed by get_indexable_post_types() since the post types returned are only public:  $post_types = get_post_types( array( 'public' => true ) ); and there is no nav_menu_items in that array since nav_menu_items is not publicly queryable.

I also added the conditional you mentioned in a new commit and removed array_unique.

Thanks and apologies if I was not too clear

PR #1145, opened 2018-10-11
Only query for post IDs when indexing posts

elliott-stocks:
We currently query for all fields, but I don't think this is necessary since we call `get_post( $post_id )` within the `queue_post()` function.

Changing the query to only query for IDs significantly improves the indexing time.

Testing locally with 5,833 documents the index time was an average of 90 seconds. Applying this patch lowered it to around 24 seconds.

Working on a bigger ElasticSearch instance with 819, 362 documents we can now index in about 2.5 hours, previously it took almost triple that. This was using AWS `t2.small.elasticsearch` instance.

### DIFF
--- a/bin/wp-cli.php
+++ b/bin/wp-cli.php
@@ -509,6 +509,7 @@ class ElasticPress_CLI_Command extends WP_CLI_Command {
 				'ignore_sticky_posts'    => true,
 				'orderby'                => 'ID',
 				'order'                  => 'DESC',
+				'fields'                 => 'ids',
 			) );
 
 			if ( $post_in ) {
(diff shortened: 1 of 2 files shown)


brandwaffle:
@elliott-stocks this sounds awesome! @allan23 @oscarssanchez can one of you review?

oscarssanchez:
Hey @brandwaffle , 

This looks good on my part, and posts got indexed correctly on my testing.

My stats before:
Total number of posts indexed: 1247
Total time elapsed: 53.787

And after:
Total number of posts indexed: 1247
Total time elapsed: 49.934

allan23:
@elliott-stocks, this is great. Thank you very much for your contribution to ElasticPress.

PR #1146, opened 2018-10-11
Add compatibility for new MLT structure

oscarssanchez:
Hi @brandwaffle , @allan23 ,

When you have some time, could you please check this PR which fixes the related post feature? This  seems to be broken since ElasticSearch 6.0 and above.

Currently, a 400 error happens and you get this response when using the ElasticPress debug bar :

```
{
    "error": {
        "root_cause": [
            {
                "type": "parsing_exception",
                "reason": "[mlt] query does not support [ids]",
                "line": 1,
                "col": 93
            }
        ],
        "type": "parsing_exception",
        "reason": "[mlt] query does not support [ids]",
        "line": 1,
        "col": 93
    },
    "status": 400
}

```
The feature is broken due to a change in the way MLT (more like this) is structured in ElasticSearch: 
https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-mlt-query.html and in my testing, this happens since version 6.0 . I tested 6.4, 6.3, 6.0 and 5.6 which seems to be the latest one before the release of 6.0 : https://www.elastic.co/downloads/past-releases . The problem seems to start since 6.0

The PR keeps the compatibility for users that use versions below 6.0, and also adds compatibility for more complicated queries (see https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-mlt-query.html )

Thanks

### DIFF
--- a/features/related-posts/related-posts.php
+++ b/features/related-posts/related-posts.php
@@ -7,33 +7,30 @@
  */
 
 function ep_related_posts_formatted_args( $formatted_args, $args ) {
-	// Since 6.0 ElasticSearch uses a new mlt search structure
-	$new_mlt = 6.0 < ep_get_elasticsearch_version();
+	if ( ! empty ( $args[ 'more_like' ] ) ) {
+		//lets compare ES version to see if new MLT structure applies
+		$new_mlt = version_compare( ep_get_elasticsearch_version(), 6.0, '>=' );
 
-	switch ( ! empty ( $args[ 'more_like' ] ) ) {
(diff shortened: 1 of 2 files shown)


oscarssanchez:
Corrected the logic as this was not working properly and added a doc bloc to the function.

allan23:
@oscarssanchez I was not able to get related posts to return from elasticsearch. I tested on ES 5.6.0, 6.2.0, and 6.3.0 all failed.

I've created a pull request for you to test some changes.

oscarssanchez:
@allan23 , the feature is working for me on 5.6 and 6.3 and 6.4

Issue #1153, opened 2018-10-24
prepare_post(). absint was applied to comment_status


jcastilloa:
Have you searched for similar issues before submitting this one? **Yes.**

Is this a bug, question or feature request? **Bug**

_Describe the issue you encountered:_

https://github.com/10up/ElasticPress/blob/9826fc6ddee3c612079d8046e48658da2a9c2023/classes/class-ep-api.php

$comment_status = absint( $post->comment_status );

In method prepare_post, absint was applied to $post->comment_status (string: "open" / "closed").

The result is zero to any value of comment_status.

oscarssanchez:
Hi @jcastilloa ,

Thanks for the catch! 

A PR was created to solve this issue :)

PR #1158, opened 2018-11-07
Do not apply absint to comment_status

oscarssanchez:
Hi @allan23 ,

Could you please use this PR which solves #1153 ?

Previously the $comment_status was set to 0 while applying `absint()`. Now, it gets either an open / closed status which is what it should. 

Thanks!

### DIFF
--- a/classes/class-ep-api.php
+++ b/classes/class-ep-api.php
@@ -583,14 +583,14 @@ class EP_API {
 			);
 		}
 
-		$post_date = $post->post_date;
-		$post_date_gmt = $post->post_date_gmt;
-		$post_modified = $post->post_modified;
+		$post_date         = $post->post_date;
+		$post_date_gmt     = $post->post_date_gmt;
+		$post_modified     = $post->post_modified;
 		$post_modified_gmt = $post->post_modified_gmt;
-		$comment_count = absint( $post->comment_count );
-		$comment_status = absint( $post->comment_status );
-		$ping_status = absint( $post->ping_status );
-		$menu_order = absint( $post->menu_order );
+		$comment_count     = absint( $post->comment_count );
+		$comment_status    = $post->comment_status;
+		$ping_status       = absint( $post->ping_status );
+		$menu_order        = absint( $post->menu_order );
 
 		if ( apply_filters( 'ep_ignore_invalid_dates', true, $post_id, $post ) ) {
 			if ( ! strtotime( $post_date ) || $post_date === "0000-00-00 00:00:00" ) {
--- a/classes/class-ep-api.php
+++ b/classes/class-ep-api.php
@@ -589,7 +589,7 @@ class EP_API {
 		$post_modified_gmt = $post->post_modified_gmt;
 		$comment_count     = absint( $post->comment_count );
 		$comment_status    = $post->comment_status;
-		$ping_status       = absint( $post->ping_status );
+		$ping_status       = $post->ping_status;
 		$menu_order        = absint( $post->menu_order );
 
 		if ( apply_filters( 'ep_ignore_invalid_dates', true, $post_id, $post ) ) {


allan23:
@oscarssanchez can we do this for ping_status as well? 

It should be a string per: https://codex.wordpress.org/Class_Reference/WP_Post

oscarssanchez:
Hi @allan23 , 

Sure, just updated with another commit now.

hjanuschka:
actually that commit breaks media updates at my side:


the request:

```
192.168.239.55:9200/krn-prod/post/1804980?refresh=wait_for

{
  "post_id": 1804980,
  "ID": 1804980,
  "post_author": {
    "raw": "h.januschka",
    "login": "h.januschka",
    "display_name": "h.januschka",
    "id": 275
  },
  "post_date": "2018-11-08 10:32:09",
  "post_date_gmt": "2018-11-08 09:32:09",
  "post_title": "Bildschirmfoto 2018-11-07 um 13.16.52",
  "post_excerpt": "",
  "post_content_filtered": "",
  "post_content": "",
  "post_status": "inherit",
  "post_name": "media__bildschirmfoto-2018-11-07-um-13-16-52",
  "post_modified": "2018-11-08 10:32:09",
  "post_modified_gmt": "2018-11-08 09:32:09",
  "post_parent": 0,
  "post_type": "attachment",
  "post_mime_type": "image\/png",
  "permalink": "https:\/\/imgl.krone.at\/2018\/11\/3eda75b6a02e9602be39cad5ddafcf7c1ad5a250.png",
  "terms": {
    "image_format": [
      {
        "term_id": 8,
        "slug": "dynamic",
        "name": "Originalbild (NxN)",
        "parent": 0,
        "term_taxonomy_id": 8
      }
    ]
  },
  "meta": {
    "copyright": [
      {
        "value": "",
        "raw": "",
        "boolean": false,
        "date": "1971-01-01",
        "datetime": "1971-01-01 00:00:01",
        "time": "00:00:01"
      }
    ]
  },
  "date_terms": {
    "year": 2018,
    "month": 11,
    "week": 45,
    "dayofyear": 311,
    "day": 8,
    "dayofweek": 4,
    "dayofweek_iso": 4,
    "hour": 10,
    "minute": 32,
    "second": 9,
    "m": 201811
  },
  "comment_count": 0,
  "comment_status": "open",
  "ping_status": "closed",
  "menu_order": 0,
  "guid": "https:\/\/wp.krone.at\/wp-content\/uploads\/2018\/11\/3eda75b6a02e9602be39cad5ddafcf7c1ad5a250.png"
}

```


ends up with:

```
{
    "error": {
        "root_cause": [
            {
                "type": "mapper_parsing_exception",
                "reason": "failed to parse [comment_status]"
            }
        ],
        "type": "mapper_parsing_exception",
        "reason": "failed to parse [comment_status]",
        "caused_by": {
            "type": "number_format_exception",
            "reason": "For input string: \"open\""
        }
    },
    "status": 400
}
```


is there some mapping missing?

hjanuschka:
as workaround, without reveting plugin or changing plugin itself:


```php
//workaround: https://github.com/10up/ElasticPress/pull/1158
add_filter('ep_post_sync_args', [$this, 'ep_post_sync_args'], 10, 2);


public function ep_post_sync_args($args, $post_id)
{
    $args['comment_status'] = absint($args['comment_status']);
    $args['ping_status'] = absint($args['ping_status']);
    return $args;
}

```

oscarssanchez:
Hi @hjanuschka ,

Thanks for spotting this, this issue might be indeed a problem with mappings as `comment_status and` `ping_status` are defined as a long type instead of `text` on indexes before this commit. 

I'm not sure if mappings change after a reindex but I'll run a quick index to see if this is happening or we have to set the data-type explicitly.

Thanks!

hjanuschka:
i just want to avoid to run a reindex, its round  about  1TB index size; and  it only works via wp cli in paged mode, if it does not help. and during the run, my editors cannot work - as they would not find many things

hjanuschka:
yeah, once i know that updating mapping, is a working solution, i am going to schedule a nightly maintaince and doing it from 00:00-06:00 in the morning or so. thx for your support and keep up the good work, without the plugin, we would actually have to remove wordpress at all, since its unusable with the amount of posts we have and the default search (no matter how much cpu we add to the db machines).

we even patched/filtered alot to even make media search ajax and stuff like this to also use EP

Issue #1068, opened 2018-04-26
facet search and range slider support?


kwiatdiamonds:
Have you searched for similar issues before submitting this one?
Yes
Is this a bug, question or feature request?
Question
Describe the issue you encountered:
I see that facet search is part of the latest update. I was wondering if it's possible to create a slider ui for a range search, as opposed to a select / checkbox output?

Current WordPress version:

Current ElasticPress version:

Current Elasticsearch version:
Version 2.5

Where do you host your Elasticsearch server:
pagely

Other plugins installed (WooCommerce, Simple Redirect Manager, etc..):

Steps to reproduce:

1. 
2. 
3. 

Screenshots, if needed:

allan23:
@kwiatdiamonds, at the moment the functionality only has checkboxes. However, we will likely add filters to the Facets to allow developers to customize it.

elvismdev:
@allan23 

> @kwiatdiamonds, at the moment the functionality only has checkboxes. However, we will likely add filters to the Facets to allow developers to customize it.

@allan23 Are any of this filters already available and customizations examples somewhere? I'm in need also to have a slider and date facets powered by ElasticPress.

PR #1160, opened 2018-11-07
Allow facet widget customization.

oscarssanchez:
Hi @allan23 ,

Could you use this PR which addresses #1068 ?

This PR uses `include_once()` to include the html output that was previously being displayed as part of `public function widget()`, and by adding a new file `facet-frontend-template.php` to the includes folder.

This enables users to replace the default template with one of their choice as noted in #1068 

Thanks!


oscarssanchez:
Hi @allan23 ,

I removed the facet template and stored the html in a variable as we discussed. Let me know if I catched this right.

Thank you!

features/facets/class-ep-facet-widget.php
```diff
@@ -229,6 +231,8 @@ public function widget( $args, $instance ) {
 			</div>
 		</div>
 		<?php
+		$facet_html = ob_get_clean();
+		echo apply_filters( 'ep_facet_search_widget', $facet_html );
```
allan23: Let's pass along some data (selected filters, current facets, etc.). 

Ideally, we want a developer to either be able to modify the existing HTML or replace it completely with their own.

includes/facet-frontend-template.php
```diff
@@ -0,0 +1,124 @@
+<?php
```
allan23: Is this file still needed?

oscarssanchez:
Hi @allan23 ,

Thanks! I addressed this in a new commit. 

I'm passing  $selected_filters, $terms_by_slug, $outputted_terms as I see those are the main ones that are used to build the widget html.

includes/facet-frontend-template.php
oscarssanchez: This file was deleted. Thanks :)

features/facets/class-ep-facet-widget.php
oscarssanchez: I passed $selected_filters, $terms_by_slug, $outputted_terms to the filter, so they can build upon those :)

Issue #898, opened 2017-07-20
Add support for Sticky posts


sc0ttkclark:
**Have you searched for similar issues before submitting this one?**
Yes

**Is this a bug, question or feature request?**
Feature request

**Describe the issue you encountered:**
When using WP_Query calls with ignore_sticky_posts as true/false, EP does nothing with Elasticsearch arguments sent in the request to float sticky posts up to the top when ignore_sticky_posts=false, so all ES queries are treated ultimately as ignore_sticky_posts=true which may be opposite of the intended WP_Query call.

**Current WordPress version:**
4.7

**Current ElasticPress version:**
2.3.2

**Current Elasticsearch version:**
5.3

**Where do you host your Elasticsearch server:**
Local server

**Other plugins installed (WooCommerce, Simple Redirect Manager, etc..):**
Not related

**Steps to reproduce:**

1. Do a WP_Query call for `'post_type' => 'post'` and `'ignore_sticky_posts' => false` (explicitly setting to false to avoid any suspicion of any defaults).
2. Get back posts, sticky posts are not listed first

**Screenshots, if needed:**
N/A

tlovett1:
Definitely not supporting this right now. We'll try to add to 2.4. Thanks!

PR #1164, opened 2018-11-13
WIP Sticky posts feature.

oscarssanchez:
In regards to https://github.com/10up/ElasticPress/issues/898 

Initial approach to implement sticky post feature:

- When adding initial mappings, we are adding a is_sticky field in ElasticSearch
- When indexing, we add the is_sticky value from wp_options to ES
- We set the _score of the is_sticky: true value boost so it always returns on top.

### DIFF
--- a/classes/class-ep-api.php
+++ b/classes/class-ep-api.php
@@ -639,7 +639,6 @@ class EP_API {
 			'ping_status'           => $ping_status,
 			'menu_order'            => $menu_order,
 			'guid'                  => $post->guid,
-			'is_sticky'             => is_sticky( $post_id )
 		);
 
 		/**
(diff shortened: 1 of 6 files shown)


oscarssanchez:
Hi @allan23 , 

Thanks for your feedback, I'll address this on upcoming commits. In the meantime, I wanted to commit some changes I did to support/fix the following:

Previously, sticky posts were displaying only when orderby was set by rand. As this is not the default on WordPress, I moved this outside and make sure to display sticky posts only on the front page, so this does not conflict when performing normal search queries.  

Also, the  `is_sticky` property was not getting updated when ES ran a update_sync. This was due to `stick_post()` running after `wp_insert_post` hence the status remained the same until we ran a manual resync.

oscarssanchez:
- Added conditionals and mappings for 5.0 and pre 5.0

tlovett1:
I kind of wonder how useful storing sticky posts in EP is. Right now there is just an option that references post IDs. Querying that option and each post based on ID is extremely fast already.

oscarssanchez:
Hi @tlovett1 ,

Thanks, I see your point! This is definitely less code to execute.

oscarssanchez:
@allan23  I included unit tests on the most recent commit, and also changed the conditional to include stickies on top from `is_front_page()` to `is_home()`

PR #1166, opened 2018-11-14
Intro page redesign

oscarssanchez:
Hi @tott ,

Could you please review this PR? 

This one redesigns the intro page that is displayed the first time a user clicks on the elasticpress plugin icon in the sidebar of the wp-admin area.

Before: 

epbefore

Now: 

epafter

Thanks!


includes/intro-page.php
```diff
@@ -20,14 +20,41 @@
...
+			<p class="ep-copy-text"><?php esc_html_e( 'You\'ve taken your first step into a faster and more flexible search and query engine for WordPress', 'elasticpress' ); ?></p>
+		</div>
+		<div class="intro-box">
+			<div class="circle white-circle">
+				<?php esc_html_e( 'Step', 'elasticpress' ); ?><p>2</p>
+			</div>
+			<h2><?php esc_html_e( 'Set up hosting', 'elasticpress' ); ?></h2>
+			<p class="ep-copy-text">
+				<?php echo sprintf( __( 'The next step is to make sure you have a working Elasticsearch server. We recommend creating an %s account or if you want you can set up your own hosting', 'elasticpress' ),
+					'<a href="https://www.elasticpress.io/">ElasticPress.io</a>'
```
tott: I would link this to blank target so we're not taking people out of the workflow

ivanlopez: @tott it might be worth adding a UTM code so you can track traffic from this click.

tlovett1:
What happens if `EP_HOST` is defined already?

includes/intro-page.php
oscarssanchez: Hi @ivanlopez, Can we add the code even if the URL is set to blank ?

oscarssanchez:
Hi @tott ,

I just added the feedback into an interactive process to configure ElasticPress. Could you pleas review? 

We are starting always on Step 2: 

step2

After setting up a host we end on Step 3 if the user visits again the elasticpress page:

step3

After a Sync, we end up on these step:

setupc

Mobile version:
setupcm

I'm wondering if this notice is still pertinent to show or should we redirect users to the intro page again after they set up their host successfully:

notice2

oscarssanchez:
Nevermind I think I got this wrong, I'll redo this today and update

oscarssanchez:
Updated this PR @tott , @brandwaffle, Can you give me your thoughts please?

allan23:
@oscarssanchez  is this already in 3.0?

CC: @tlovett1

oscarssanchez:
Looks like it is @allan23 https://github.com/10up/ElasticPress/pull/1253

Issue #1127, opened 2018-08-23
Using compare LIKE


altendorfme:
Have you searched for similar issues before submitting this one?
**Yes**

Is this a bug, question or feature request?
**Question**

Describe the issue you encountered:
If you use '_compare' => 'LIKE'_ returns:

**Query Response Code: HTTP 400**
```
{
    "error": {
        "root_cause": [
            {
                "type": "parsing_exception",
                "reason": "no [query] registered for [query]",
                "line": 1,
                "col": 172
            }
        ],
        "type": "parsing_exception",
        "reason": "no [query] registered for [query]",
        "line": 1,
        "col": 172
    },
    "status": 400
}
```

WP_Query:
```
		$agenda_semana = new WP_Query( array('ep_integrate'   => true,
				'post_type'      => 'post',
				'post_status'    => 'publish',
				'posts_per_page' => - 1,
				'fields' => 'ids',
				'meta_query'     => array(
						'relation' => 'AND',
						array(
								array(
										'relation' => 'AND',
										array(
												'key'     => 'post_local_data_repetidor',
												'value'   => $semana,
												'compare' => 'LIKE'
										),
										array(
												'key' => 'post_local_data_tipo',
												'value' => 'fixo'
										)

								)
						),
						array(
								$meta_query
						),
				),
                'category__not_in' => array(109078)
		) );
```

Current WordPress version:
**4.9.8**

Current ElasticPress version:
**2.5.2**

Current Elasticsearch version:
**6.2.4**

Where do you host your Elasticsearch server:
**Elastic.co**

altendorfme:
Host: https://XYZ.searchly.com
Time Taken: 129 ms
URL: https://XYZ.searchly.com/XYZ-1/post/_search
Method: POST
Query Args: 
```
array(67) {
  ["ep_integrate"]=>
  bool(true)
  ["post_type"]=>
  string(4) "post"
  ["post_status"]=>
  string(7) "publish"
  ["posts_per_page"]=>
  int(-1)
  ["fields"]=>
  string(3) "ids"
  ["meta_query"]=>
  array(3) {
    ["relation"]=>
    string(3) "AND"
    [0]=>
    array(1) {
      [0]=>
      array(3) {
        ["relation"]=>
        string(3) "AND"
        [0]=>
        array(3) {
          ["key"]=>
          string(25) "post_local_data_repetidor"
          ["value"]=>
          string(9) "wednesday"
          ["compare"]=>
[...]

PR #1167, opened 2018-11-15
Use wildcard query with compare like.

oscarssanchez:
Hi @allan23 , @tott ,

Could you please review this PR which addresses #1127 ?

The compare like query was previously defined as: 

						case 'like':
							if ( isset( $single_meta_query['value'] ) ) {
								$terms_obj = array(
									'query' => array(
										'match' => array(
											$meta_key_path => $single_meta_query['value'],
										)
									),
								);
							}
							break;

However we can't make a query within a filter query context. We can however use a type of a query inside post_filter to filter out results. 

Since compare-> like is defined like this: 

https://github.com/WordPress/WordPress/blob/df0958697a5d897a15ea78d37a001e2f3752875f/wp-includes/class-wp-meta-query.php#L595

I think we could mimic the same behavior by using a wildcard query : 

https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-wildcard-query.html

I've tested this succesfully returning posts that match the provided meta value.

Steps to reproduce testing:

1.- Create 3 posts
2.- Add new meta to the 3 of them like this: 

`add_post_meta( $post_id, 'test', 'single_pr');
add_post_meta( $post_id, 'test', 'single_tr'); add_post_meta( $post_id, 'test', 'singlewow');`

3.-  Make a custom query to search for posts that match 'single' in some part: 

		$test_query = new WP_Query( array('ep_integrate'   => true,
				'post_type'      => 'post',
				'post_status'    => 'publish',
				'posts_per_page' => - 1,
				'fields' => 'ids',
				'meta_query'     => array(
						'relation' => 'AND',
						array(
								array(
										'relation' => 'AND',
										array(
												'key'     => 'test',
												'value'   => single,
												'compare' => 'LIKE'
										),
										array(
												'key' => 'test',
												'value' => 'redundant'
										)

								)
						),
						array(
								$meta_query
						),
				),
                'category__not_in' => array(109078)
		) );

This returns the following response from elasticsearch:

```
{
    "took": 30,
    "timed_out": false,
    "_shards": {
        "total": 5,
        "successful": 5,
        "skipped": 0,
        "failed": 0
    },
    "hits": {
        "total": 3,
        "max_score": null,
        "hits": [
            {
                "_index": "estest1devlocalhost-1",
                "_type": "post",
                "_id": "10",
                "_score": null,
                "_source": {
                    "post_id": 10
                },
                "sort": [
                    1542305933000
                ]
            },
            {
                "_index": "estest1devlocalhost-1",
                "_type": "post",
                "_id": "9",
                "_score": null,
                "_source": {
                    "post_id": 9
                },
                "sort": [
                    1542305927000
                ]
            },
            {
                "_index": "estest1devlocalhost-1",
                "_type": "post",
                "_id": "8",
                "_score": null,
                "_source": {
                    "post_id": 8
                },
                "sort": [
                    1542305917000
                ]
            }
        ]
    }
}

```
Thanks!

### DIFF
--- a/classes/class-ep-api.php
+++ b/classes/class-ep-api.php
@@ -1928,10 +1928,8 @@ class EP_API {
 						case 'like':
 							if ( isset( $single_meta_query['value'] ) ) {
 								$terms_obj = array(
-									'query' => array(
-										'match' => array(
-											$meta_key_path => $single_meta_query['value'],
-										)
+									'wildcard' => array(
+										$meta_key_path => $single_meta_query['value'] . '*',
 									),
 								);
 							}


oscarssanchez:
Additionally, this was tested here: https://github.com/10up/ElasticPress/issues/1127#issuecomment-429864743

allan23:
@oscarssanchez this PR is causing a unit test to fail for a LIKE query. Can you please re-review?

oscarssanchez:
Hi @allan23 ,

You are correct. The Unit test is not passing. 

I noticed two things:

1.- The wildcard query is actually looking for anything that matches a pattern + something more ( that's why my tests were successful as they were looking  "single" on these values: singlewow, single_pr and single_tr, it makes sense why they work as they have the word "single" plus something more. 

It looks like the documentation might be wrong as it points : 

> _It uses the standard shell wildcards: ? matches any character, and * matches zero or more characters._

https://www.elastic.co/guide/en/elasticsearch/guide/current/_wildcard_and_regexp_queries.html

To correct this issue, the regexp query works just fine, however, I would like to have your thoughts if this could be a good fit to replace the wildcard query. 

2.- When trying to reproduce the unit test on WordPress, I found the response from ES is HTTP 400, however the results showing just fine because of defaulting to MySQL. This makes me think the Unit Test are passing with the previous code because the search is getting done (at least this one) via MySQL. 

Thanks!

PR #1173, opened 2018-11-27
Fix postType being set to 'publish' instead of postStatus.

ray-lee:
The `buildSearchQuery` function in autosuggest.js contains this:
https://github.com/10up/ElasticPress/blob/093b9ebbf8518420391706f47d7461a3e5cac136/features/autosuggest/assets/js/src/autosuggest.js#L92-L94

I assume this was a copy/paste error; "publish" should be the post status, not the post type.


oscarssanchez:
Thanks @ray-lee !
I think this is good

oscarssanchez:
@allan23 can you check too?

PR #1194, opened 2019-01-08
send index names as part of _recovery or _stats request

dustinrue:
When calling `wp elasticpress status` or `wp elasticpress stats` the call is to the root of ES server. These calls can fail in shared environments depending on how they are configured. This PR modifies the requests so that they include all indices as part of the request.

Is not tested on large multisite installs.


TheLastCicada:
There's security implications of the `status` command at the moment exposing too much information, so I suggest this issue be addressed as a high priority problem.

PR #1195, opened 2019-01-08
Show term counts in facet widget. (#1180)

### LINKED ISSUE #1180: Show counts in faceted search widget
Yes

Feature request

In the faceted search widget, we'd like to display the number of posts that are associated with each term.

If #1160 is merged, we could do this using the ep_facet_search_widget filter, but we would need to duplicate a lot of code, and this seems like it would be a common enough requirement that it should be supported out of the box. I'm happy to implement this and make a pull request.

Current WordPress version:
4.9.8

Current ElasticPress version:
2.6.1

6.2

ray-lee:
Fixes #1180. The facet widget now shows the count for each term:

facet_term_count

The counts can be hidden with CSS, using the `term-count` class.

In order to get the sass to compile, I had to fix the version of grunt-sass at version 2 in package.json. Version 3 of grunt-sass isn't compatible with the current gruntfile, so the version range can't be `*`.


allan23:
Hi @ray-lee!
Thank you for the pull request.

Since this adds the counts to the facets, which is an excellent feature, I'm concerned about users who upgrade that are used to facets not showing counts that will then start seeing them upon upgrading. I would recommend that you add a filter that enables facet counts instead of having them enabled by default.

allan23:

PR #1197, opened 2019-01-08
ep_indexable_blogs filter 

allan23:
Added in case developers want to override the results of the get_sites() call for public sites.


oscarssanchez:
Hi @allan23 ,

This works for me.

Issue #1199, opened 2019-01-09
`wp elasticpress status` reveals too much information


TheLastCicada:
Have you searched for similar issues before submitting this one?  yes

Is this a bug, question or feature request?  feature request

Describe the issue you encountered:
The command `wp elasticpress status` reveals information about all indexes on the Elasticsearch server, not just the index related to this WordPress install.  On a shared Elasticsearch setup (which is not uncommon), this reveals information about all other customers or users of this Elasticsearch instance, which, in some cases, is information that should not be public and could be considered a security issue.  I suggest the ElasticPress plugin switch to getting the status of only the index for the site in question by specifying the index in the API request (https://www.elastic.co/guide/en/elasticsearch/reference/current/indices-recovery.html).  Basically, `http://<es>:9200/index-name/_recovery` rather than `http://<es>:9200/_recovery`.  

I am aware that this does nothing to secure the actual elasticsearch endpoint and that this information would still be available by querying the Elasticsearch server directly, but our plugin should not provide a shortcut to disclosing this information.  There is currently a pull request that already works to address this, but likely needs some additional attention to be production ready https://github.com/10up/ElasticPress/pull/1194.  

Current WordPress version:  all

Current ElasticPress version:  all

Current Elasticsearch version:  all

Where do you host your Elasticsearch server:  

Other plugins installed (WooCommerce, Simple Redirect Manager, etc..):

Steps to reproduce:

1. `wp elasticpress status` on a shared cluster
2.  view all of the indexes that aren't yours

Screenshots, if needed:

allan23:
@TheLastCicada we are discussing approaches.

PR #1200, opened 2019-01-11
Only display indices index of current site in `wp elasticpress status

turtlepod:
Note: not sure if this is the correct method to do this.

Issue #1199


allan23:
Hi @turtlepod, 

We have decided to use a different approach to this issue. However, we greatly appreciate your willingness to contribute to the plugin. Please keep it up! :-)

-Allan

PR #1204, opened 2019-01-15
Hide prefix trailing dash on settings page

### LINKED ISSUE #1202: Hide trailing dash from Subscription ID input field.
Yes

Feature request

When using ElasticPress.io the Subscription ID field shows a trailing dash after saving.

ElasticPress.io

Steps to reproduce:
1. Go to plugin settings and enter an elasticpress.io endpoint and credentials
2. Save
3. Look at subscription ID field

Task:
Remove the trailing dash from the text field in the subscription ID and only add it in the backend programmatically where required to avoid confusion.

oscarssanchez:
Hi @tott and @allan23 ,

Could you please review this PR which fixes #1202  ?

This PR hides the trailing dash for the EP_PREFIX ( Subscription ID ) on the settings page.

Thank you.

### DIFF
--- a/includes/settings-page.php
+++ b/includes/settings-page.php
@@ -56,7 +56,7 @@ if ( defined( 'EP_IS_NETWORK' ) && EP_IS_NETWORK ) {
 					</th>
 					<td>
 						<?php if ( apply_filters( 'ep_admin_show_index_prefix', true ) ) : ?>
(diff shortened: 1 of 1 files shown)


oscarssanchez:
Hi @tott ,

Thanks for the feedback, using `rtrim()` instead is indeed a better solution. 

This is only affecting the display of `EP_INDEX_PREFIX`, the trailing dash is always appended if there is not a dash already at the end of whatever the user submits for Subscription ID, as you can see here: 

https://github.com/10up/ElasticPress/blob/develop/classes/class-ep-config.php#L263

I've tested this works with elasticress.io credentials without problems.

Issue #1179, opened 2018-12-03
Determine "Sweet-spot" for Bulk Post Indexing


allan23:
On some occasions, indexing fails for a number of reasons. We need to determine if the post count of 350 is a good number.

Perhaps allowing the end-user to set the number of posts in the admin.

Issue #1016, opened 2018-02-01
New hook 'ep_prepare_meta_allowed_public_keys' would be very helpful


waynedavidharris:
Hi,
When setting public post meta keys to index you have a hook to only exclude keys 'ep_prepare_meta_excluded_public_keys'. I have posts with over 40 postmeta fields and only wish to index a few. A hook allowing which keys to include would be very helpful and to be honest feels less counter-intuitive.
I added my own but obviously would make sense to have it in the core.
Its a simple change that complements the existing exclude hook.
Note i also added a third arg array_keys($meta) which i believe is also v helpful
Thanks for the plugin.

class EP_API

***************************************************************
 $allowed_public_keys = apply_filters( 'ep_prepare_meta_allowed_public_keys', false, $post, array_keys($meta));
***************************************************************

foreach ( $meta as $key => $value ) {

			$allow_index = false;

			if ( is_protected_meta( $key ) ) {

				if ( true === $allowed_protected_keys || in_array( $key, $allowed_protected_keys ) ) {
					$allow_index = true;
				}
			} else {

                //*****************************************************************************
                if ( true === $allowed_public_keys || in_array( $key, $allowed_public_keys ) ) {
                    $allow_index = true;
                }
                //*****************************************************************************

				if ( true !== $excluded_public_keys && ! in_array( $key, $excluded_public_keys )  ) {
					$allow_index = true;
				}
			}

			if ( true === $allow_index || apply_filters( 'ep_prepare_meta_whitelist_key', false, $key, $post ) ) {
				$prepared_meta[ $key ] = maybe_unserialize( $value );
			}
		}

tlovett1:
Feel free to open a PR.

Salamander3:
Can someone please write an example of how to use ep_prepare_meta_whitelist_key filter to indext some meta keys? I can't figure it out by looking at it.

PR #1207, opened 2019-01-18
Allow user to set bulk settings

oscarssanchez:
Hi @allan23 ,

Could you please review this PR which adds feature #1179 ?

This PR adds a field  on the settings page if `ep_host` is set and filter `ep_index_posts_per_page` has no callback function: 

epsettings

I've tested this works on single and multisite.

Please let me know your thoughts. 
Thanks!

### DIFF
--- a/classes/class-ep-config.php
+++ b/classes/class-ep-config.php
@@ -120,6 +120,21 @@ class EP_Config {
 	}
 
 	/**
+	 * Retrieve bulk index settings
+	 *
+	 * @return Int The number of posts per cycle to index. Default 350.
+	 */
+	public function get_bulk_settings() {
+		if( defined( 'EP_IS_NETWORK' ) && EP_IS_NETWORK ) {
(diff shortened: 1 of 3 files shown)


kylef000:
@oscarssanchez It would be helpful to include an increase in the timeout of `bulk_index_posts` in class-ep-api.php. I've found that increasing this value fixed indexing failures for me (although this may mask an underlying timeout issue with the ElasticSearch host).

allan23:
@kylef000, you can actually change the timeout via the `ep_bulk_index_posts_request_args` filter.

kylef000:
@allan23 good to know. Thanks! 

I made some a few changes to add in a filter `ep_prepare_meta_allowed_public_keys` (which I later found was a proposed change here https://github.com/10up/ElasticPress/issues/1016). Some of the posts we were indexing had 50+ meta key value pairs due to ACF usage. Was able to go from indexing 5 posts at a time to 3000 (~1hr indexing time to <45 seconds).

PR #1223, opened 2019-01-29
Use match_phrase as alternative for wildcard

oscarssanchez:
Hi @allan23 ,

This PR replaces the use of wildcard query for a match_phrase. This way, we account for queries that are not a pattern.

Thank you,


jeffpaul:
@oscarssanchez can you check the build failure before Allan dives into his review?

allan23:
@jeffpaul Travis is notorious with having issues. It's on our list to replace it with something that is consistent.

PR #1235, opened 2019-02-07
Feature/search feature narrow search 02 06

mmcachran:


tlovett1:
@brandwaffle curious on your thoughts here. This is a setting for "exact matches" on the post feature.

PR #1247, opened 2019-02-13
Do not allow 0 to be set as bulk settings

oscarssanchez:
Hi @allan23 ,

Can you please review this PR ?

This adds a new function to first convert the bulk settings input into a integer, then check if the value is 0 or empty, as this can sometimes be the result of converting into an integer : http://php.net/manual/en/language.types.string.php#language.types.string.conversion

If it is empty or 0, then set bulk settings to default of 350.

This was being problematic as `register_setting()` was submitting an empty string the first time a user sets his elasticsearch host, hence the option was getting updated and ended up getting converted to 0.

Thanks!

### DIFF
--- a/classes/class-ep-config.php
+++ b/classes/class-ep-config.php
@@ -367,6 +367,21 @@ function ep_sanitize_credentials( $credentials ) {
 }
 
 /**
+ * Sanitize bulk settings.
+ *
+ * @param $bulk_settings
+ * @return int
+ */
+function ep_sanitize_bulk_settings( $bulk_settings ) {
+	$bulk_settings = (int) $bulk_settings;
+	if ( empty( $bulk_settings) || 0 === $bulk_settings || ! isset( $bulk_settings ) ) {
+		return $bulk_settings = 350;
+	} else {
+		return $bulk_settings;
+	}
+}
+
+/**
  * Prepare credentials if they are set for future use.
  */
 function ep_setup_credentials(){
--- a/classes/class-ep-config.php
+++ b/classes/class-ep-config.php
@@ -372,13 +372,9 @@ function ep_sanitize_credentials( $credentials ) {
  * @param $bulk_settings
  * @return int
  */
-function ep_sanitize_bulk_settings( $bulk_settings ) {
+function ep_sanitize_bulk_settings( $bulk_settings = 350 ) {
 	$bulk_settings = (int) $bulk_settings;
-	if ( empty( $bulk_settings) || 0 === $bulk_settings || ! isset( $bulk_settings ) ) {
-		return $bulk_settings = 350;
-	} else {
-		return $bulk_settings;
-	}
+	return ( 0 === $bulk_settings ) ? 350 : $bulk_settings;
 }
 
 /**
--- a/classes/class-ep-config.php
+++ b/classes/class-ep-config.php
@@ -373,7 +373,7 @@ function ep_sanitize_credentials( $credentials ) {
  * @return int
  */
 function ep_sanitize_bulk_settings( $bulk_settings = 350 ) {
-	$bulk_settings = (int) $bulk_settings;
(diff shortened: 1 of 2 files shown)


classes/class-ep-config.php
```diff
@@ -366,6 +366,21 @@ function ep_sanitize_credentials( $credentials ) {
 	];
 }
 
+/**
+ * Sanitize bulk settings.
+ *
+ * @param $bulk_settings
+ * @return int
+ */
+function ep_sanitize_bulk_settings( $bulk_settings ) {
```
allan23: Let's set `$bulk_settings` to 350 as a default here.

classes/class-ep-config.php
```diff
@@ -366,6 +366,21 @@ function ep_sanitize_credentials( $credentials ) {
...
+	$bulk_settings = (int) $bulk_settings;
+	if ( empty( $bulk_settings) || 0 === $bulk_settings || ! isset( $bulk_settings ) ) {
```
allan23: Since we've already typecast `$bulk_settings` as an integer, is there any reason why it would be anything other than an integer?

If a string were to be sent to this sanitizing method, that string would be converted to `0`. Also `isset` and `empty` seem redundant. Maybe just check if it's zero?

classes/class-ep-config.php
```diff
@@ -366,6 +366,21 @@ function ep_sanitize_credentials( $credentials ) {
...
+		return $bulk_settings = 350;
+	} else {
```
allan23: This `else` statement is not needed. You can just return `$bulk_settings`.

allan23: Or just put it all on one line like this:

`return ( 0 === $bulk_settings ) ? 350 : $bulk_settings;`

classes/class-ep-config.php
```diff
@@ -366,6 +366,21 @@ function ep_sanitize_credentials( $credentials ) {
...
+/**
+ * Sanitize bulk settings.
+ *
+ * @param $bulk_settings
+ * @return int
+ */
+function ep_sanitize_bulk_settings( $bulk_settings ) {
+	$bulk_settings = (int) $bulk_settings;
+	if ( empty( $bulk_settings) || 0 === $bulk_settings || ! isset( $bulk_settings ) ) {
+		return $bulk_settings = 350;
```
allan23: There shouldn't be a reason to assign $bulk_settings to 350 on the return.

classes/class-ep-config.php
oscarssanchez: Thanks Allan, I addressed this on the recent commit.

classes/class-ep-config.php
oscarssanchez: I think this makes sense since no boolean values can get submitted. Thank you Allan!

classes/class-ep-config.php
oscarssanchez: This looks like a cleaner way to write it. I made the change in the recent commit. Thanks Allan

classes/class-ep-config.php
oscarssanchez: Thanks Allan, I addressed this in the recent commit

classes/class-ep-config.php
```diff
@@ -366,6 +366,17 @@ function ep_sanitize_credentials( $credentials ) {
...
 }
 
+/**
+ * Sanitize bulk settings.
+ *
+ * @param $bulk_settings
+ * @return int
+ */
+function ep_sanitize_bulk_settings( $bulk_settings = 350 ) {
+	$bulk_settings = (int) $bulk_settings;
```
allan23: Let's change this to absint(), just on the off chance that some jokester would put in a negative value.

oscarssanchez: Thanks Allan, nice catch!

Issue #1268, opened 2019-02-28
Improve settings flow for inputting ElasticSearch host info


dkoo:
Is this a bug, question or feature request? **Feature enhancement**

Describe the issue you encountered: **Improve the user flow for inputting ElasticSearch host credentials**

Currently, if you input an ElasticPress.io URL as an ES host, the settings page reveals three additional fields after clicking "save":

https://cl.ly/0e39c95f56fb/Screen%20Recording%202019-02-28%20at%2009.44%20AM.gif

This could be improved by using a JS-based toggle on the page to allow the user to input ElasticPress credentials or a simple URL for third-party host. It would also make it clearer what steps are required for users to complete input in both cases.

PR #1269, opened 2019-02-28
Improved settings flow

dkoo:
Implements a tabbed interface to enter ElasticSearch or ElasticPress.io credentials on the settings page. This lets you toggle the available fields based on whether your host is EP.io or another host. 

https://cl.ly/0e7da052cc3d

Styles are basic WP admin styles. Note that if you have previously entered EP.io credentials, but then save third-party credentials in this form, EP.io-specific fields (subscriber ID, user, password) will be cleared out.


dkoo:
Addresses https://github.com/10up/ElasticPress/issues/1268

dkoo:
Seems like there are some build errors causing the webpack config file to be linted. Will investigate.

tlovett1:
Looks solid.

Issue #1265, opened 2019-02-25
Incompatibility with WordPress 5.0 REST API search route


elvishp2006:
Have you searched for similar issues before submitting this one?
- Yes

Is this a bug, question or feature request?
- Bug

Describe the issue you encountered:
- The new WordPress REST API search route does not work when the ElasticPress is active.

Current WordPress version: **5.1**

Current ElasticPress version: **2.8.1**

Current Elasticsearch version: **5.6.5**

Where do you host your Elasticsearch server: http://elasticsearch:9200

Other plugins installed (WooCommerce, Simple Redirect Manager, etc..):
- None;

Steps to reproduce:

1. Access `http://localhost/wp-json/wp/v2/search/?search=world`
2. Receive the following response: 
```json
{
  "code": "rest_search_handler_error",
  "message": "Internal search handler error.",
  "data": {
    "status": 500
  }
}
```

After some debug on WordPress Core, I noticed response from that [line](https://github.com/WordPress/WordPress/blob/67e811c15e703d2415e7435bf9878282265e4028/wp-includes/rest-api/endpoints/class-wp-rest-search-controller.php#L130) don't have a value on `total` index.

### With ElasticPress:
```php
Array
(
    [ids] => Array
        (
            [0] => 1
            [1] => 2
        )

    [total] => 
)
```

### Without ElasticPress
```php
Array
(
    [ids] => Array
        (
            [0] => 1
        )

    [total] => 1
)
```

valnei:
Same problem here on version 5.0.

allan23:
Hi @elvishp2006, thank you for bringing this to our attention.

PR #1271, opened 2019-03-04
Addition of found_posts filter

allan23:
Per #1265, the REST API fails to return search queries due to found_posts returning a null value.


oscarssanchez:
Hi @allan23 , while this works for me, I could not reproduce #1265 . For me `http://localhost/wp-json/wp/v2/search/?search=world` works with this as well as a previous state, for example: `e3927b497ed1d89e0b6eadbfc229a185eaea1373`. Is there something I'm missing?

oscarssanchez:
I could reproduce with 6.2.4 and works correctly for me now

allan23:
@tlovett1 do you foresee any issues on this?

PR #1276, opened 2019-03-06
Remove Base64 Attachment Data

allan23:


oscarssanchez:
ElasticSearch is showing the PDF content correctly for me and it is not returning the data field anymore.

PR #1277, opened 2019-03-06
Runs processor to remove base64 attachment data field.

allan23:


oscarssanchez:
This is also working for me on 3.0 @allan23

PR #1289, opened 2019-03-18
Fix Taxonomy identification

aluferraz:
When you rewrite the filter url to hide the "pa_" string, ElasticPress breaks the query by searching a taxonomy that doesn't exist.

For example:
http://mywoocommerce.url/shop/?filter_brands=apple 
**Elasticpress searches for a taxonomy named "brands", which doesn't exist. The correct taxonomy is "pa_brands"**

http://mywoocommerce.url/shop/?filter_pa_brands=apple 

**This would produce the right query, but the URL is uglier**

**To fix that, I made the taxonomy identification just like Woocommerce's code (woocommerce/includes/class-wc-query.php / get_layered_nav_chosen_attributes)**

[`class-wc-query.php`](https://github.com/woocommerce/woocommerce/blob/master/includes/class-wc-query.php#L768)


tlovett1:
Hello, please recreate and reopen for version 3.0. Thank you!

Issue #1302, opened 2019-04-01
change host and index prefix on the fly via CLI


mustafauysal:
Overriding host and prefix via CLI command would be useful in some cases.

I'm not sure when #659 will be implemented, but when we able to index different hosts, shifting Elasticsearch version can be smoother.  (~10 hours downtime is not good for working with large datasets)

What are you guys thinking about this? 🤔

oscarssanchez:
Hi @mustafauysal , thanks for the PR!!, I will bring this up in our internal standup.

mustafauysal:
The PR has been merged, I'm closing the issue.

PR #1303, opened 2019-04-01
Feature/override host and prefix via cli

mustafauysal:
related #1302

### DIFF
--- a/includes/classes/Command.php
+++ b/includes/classes/Command.php
@@ -627,7 +627,7 @@ class Command extends WP_CLI_Command {
 		}
 
 		if ( ! empty( $args['include'] ) ) {
-			$include               = explode( ',', str_replace( ' ', '', $assoc_args['include'] ) );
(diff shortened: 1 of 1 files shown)


tlovett1:
@mustafauysal this looks fine to me. Perhaps we should separate this code out so it can be used for other commands as well e.g. put_mapping?

mustafauysal:
@tlovett1 in e1f693647830040c42babb506ba91fe3c2daaefc seperated code with `maybe_*` methods.


another alternative approach, we can move logic into the `get_host()` and `get_index_prefix()` methods. 

Something like: 
```
	if ( defined( 'WP_CLI' ) && is_callable( '\WP_CLI::get_runner' ) ) {
		$args = \WP_CLI::get_runner()->assoc_args;
		if ( isset( $args['ep-host'] ) && $args['ep-host'] ) {
			$host = $args['ep-host'];
		}
	}
```

I don't like the adding extra complexity to util functions, though.

PR #1314, opened 2019-05-03
Changed max int to use core predefined constant.

fabianmarz:
### Problem
- Static value might interfere 32-bit signed integers as the max value is `2,147,483,647`, while for 64-bit it's `9,223,372,036,854,775,807`.

### Goal
- Avoid conflicts between 32- and 64-bit.

### Proposed solution
- Use predefined PHP constant to output the maximum integer value depending on the system.

### Unclear
- Was there an intention by use the static value instead of the constant?


tlovett1:
Thanks @fabianmarz

PR #1318, opened 2019-05-14
ep_find_related is gone

goaround:
After updating to 3.0 the function ep_find_related() seems to be gone. But its still referenced under Features in the README.md

Should also be mentioned in the changelog because otherwise it will throw a Uncaught Error: Call to undefined function ep_find_related()

### Description of the Change

ep_find_related removed from the readme

### Alternate Designs

ep_find_related could be preserved and just call the Feature class.

### Benefits

Prevent confusion

### Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,7 +50,7 @@ ElasticPress understands data in real time, so it can instantly deliver engaging
 
 Available API functions:
 
-* `ep_find_related( $post_id, $return = 5 )`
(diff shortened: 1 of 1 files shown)


tlovett1:
Hey @goaround, we can add this for backwards compat and deprecate it. I'll merge that along with your docs change into 3.0.1.

PR #1335, opened 2019-05-22
ep_retrieve_the_[$type] filter

gassan:
### Description of the Change

I could not found informations about removal of the filter. so I have added it again.

### Benefits

its possible again to populate hit scores to documents

### Possible Drawbacks

None


tlovett1:
Thanks!

PR #1342, opened 2019-05-24
Integrate WooCommerce default filter by price widget with ES range query 

oscarssanchez:
### Description of the Change

This PR modifies EP default query for a range query (https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-range-query.html) when the default WooCommerce filter by price widget has been set and submitted.

### Alternate Designs

I explored the possibility of building an EP widget for this purpose but came into conclusion accommodating for the range query was the only necessary step. This also enables users to create alternative designs for filtering e.g a text input

### Benefits

I've seen filtering has been a pain point for ElasticPress users as well as ElasticPress.io users. This would be a step forward to provide support for advanced filtering with the use of facets. 

### Possible Drawbacks

Might need testing in more scenarios.

### Checklist:


traed:
@oscarssanchez I did some testing against WooCommerce default products. There seem to be some issues though. If I filter on prices between 0 and 15, products that cost 2 disappears. If I filter between 0 and 2, products that cost between 0 and 19 appears. My guess is that it compares strings when it should be comparing numeric values.

oscarssanchez:
Hi @traed ,

Thanks so much for spotting that bug. I can confirm in the range between 11-19 elasticsearch is not returning any results for $2 products. The mappings of meta._price are set to double not to string, so I don't think this is iikely the issue. 

I could not reproduce the bug where if you filter between 0-2 products that cost more appear. On that example, i get it to show only the $2 product one. 

I reached out in elasticsearch forums in regard to see if there's any insight.

oscarssanchez:
@traed turns out i was looking at meta._value.price instead of meta._long.price, after this it looks like this is working correctly. Could you test one more time ? cc @allan23

traed:
@oscarssanchez Seems to be working now :)

oscarssanchez:
adding @tlovett1 for review

includes/classes/Feature/WooCommerce/WooCommerce.php
```diff
@@ -824,4 +825,53 @@ public function requirements_status() {
...
+	 * @return mixed
+	 *
+	 */
+	public function price_filter( $args, $query_args ) {
+		$set_low_price = ! empty( $_GET['min_price'] );
+		$set_high_price = ! empty( $_GET['max_price'] );
+		$query = $args['query'];
+
+		//Account for match all context
+		if ( ( $set_high_price || $set_low_price ) && ! empty( $args['query']['match_all'] ) ) {
```
tlovett1: Will throw a notice/warning if `query` not set.

includes/classes/Feature/WooCommerce/WooCommerce.php
```diff
@@ -824,4 +825,53 @@ public function requirements_status() {
...
+			unset ( $args['query']['match_all'] );
+			$args['query']['range']['meta._price.long']['gte'] = $_GET['min_price'];
```
tlovett1: Will throw a warning/notice if many of the parent array properties are not set.

includes/classes/Feature/WooCommerce/WooCommerce.php
oscarssanchez: Thanks @tlovett1 I will address this on next commit.

includes/classes/Feature/WooCommerce/WooCommerce.php
oscarssanchez: Hi @tlovett1, sorry I did not understand this comment, for what i know `$args['query']['range']['meta._price.long']['gte'] = $_GET['min_price'];` will create all of the parent array properties and `unset ( $args['query']['match_all'] );` is getting checked on L.843 

Could you please outline the steps to reproduce the scenario where this is failing so I get a better idea of what is wrong here?

oscarssanchez:
Thanks for your comments @tlovett1 , I had one question in regard of one of them

oscarssanchez:
Hi @tlovett1 , i updated the PR to check if $args['query'] is empty. 

For the checks requested for https://github.com/10up/ElasticPress/pull/1342/commits/9f913d5bd6792d83a5a265f1fd1681a47e1eb9f5#diff-398dcfe6a8fd4b7046b31cadddc5e6bbR852 and similar other ones, my intention here was to create all the parent arrays. So for example, for `$args['query']['range']['meta._price.long']['lte']`, the parent `$args['query']['range']` will never be set prior to the variable assignment. This is not throwing an error for me and creates the parent and value just fine:

This is $args['query'] before:

es2


And after parent array creation/unsetting match all array:

es1


The same can be achieved by creating the `['range']['meta._price.long']['lte']` array and then merging it with `$args['query']` but I thought this was easier. I can adjust if this approach makes it easier to understand for others.

oscarssanchez:
Fixed the phpcs errors now so CI build passes.

tlovett1:
See #1491

PR #1344, opened 2019-05-24
Product attribute facets for 3.0.

allan23:
### Description of the Change

This PR allows WooCommerce product attributes to be facetable in 3.0. The initial implementation of attribute facets (in 2.8) caused a bug where attributes were disabled on product category pages.

### DIFF
--- a/includes/classes/Feature/Facets/Facets.php
+++ b/includes/classes/Feature/Facets/Facets.php
@@ -202,7 +202,10 @@ checked<?php endif; ?> value="any"><?php echo wp_kses_post( __( 'Show all conten
 			return false;
 		}
 
-		if ( ! ( $query->is_post_type_archive() || $query->is_search() || ( is_home() && empty( $query->get( 'page_id' ) ) ) ) ) {
+		if ( ! ( ( function_exists( 'is_product_category' ) && is_product_category() )
+		         || $query->is_post_type_archive()
+		         || $query->is_search()
+		         || ( is_home() && empty( $query->get( 'page_id' ) ) ) ) ) {
 			return false;
 		}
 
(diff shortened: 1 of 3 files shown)


includes/classes/Feature/Facets/Facets.php
```diff
@@ -224,10 +227,14 @@ public function facet_query( $query ) {
 
 		$taxonomies = get_taxonomies( array( 'public' => true ) );
 
+		// Allow other plugins to modify the available taxonomies.
+		$taxonomies = apply_filters( 'ep_facet_include_taxonomies', $taxonomies, false );
```
fabianmarz: What is the second argument `false` is for? 🤔 Doesn't seem useful to me.

includes/classes/Feature/Facets/Widget.php
```diff
@@ -343,6 +343,7 @@ public function form( $instance ) {
 		$facet = ( ! empty( $instance['facet'] ) ) ? $instance['facet'] : '';
 
 		$taxonomies = get_taxonomies( array( 'public' => true ), 'object' );
```
fabianmarz: @oscarssanchez, @tlovett1 did you test the widget as well? This is not working for me as in line 362 the foreach expects an object to display the widget titles while [`add_product_attributes()`](https://github.com/10up/ElasticPress/pull/1344/files#diff-398dcfe6a8fd4b7046b31cadddc5e6bbR742) returns an array of slugs only. Therefor the foreach loop cannot access properties from the objects and thus the list will render empty option/values.

I see following possibilities:

1. Expect a taxonomy array in general by remove the second argument of `
get_taxonomies` and adjust the loop accordingly
2. Return attribute taxonomy objects instead of names/slugs

I think option 1 would be less effort, and more performant but will also render simple slugs instead of nice name labels which might uncomfortable for some users. What do you think?

finalwebsites: Please go for option 2 thanks :)

includes/classes/Feature/Facets/Facets.php
fabianmarz: Fixed this in #1393

includes/classes/Feature/Facets/Widget.php
fabianmarz: Provided a PR in #1393.

PR #1345, opened 2019-05-24
Resolves PHP notice after upgrade to 3.0.

allan23:
### Description of the Change

When visiting the ElasticPress Settings page, the following PHP notice is generated:

`[24-May-2019 18:14:17 UTC] PHP Notice:  Undefined index: defaults_enabled in /includes/dashboard.php on line 424`

This PR does an additional check to verify that the array index exists.


tlovett1:
@allan23 I'm not getting this notice and I don't understand how you are.

tlovett1:
@allan23 is this still an issue?

PR #1352, opened 2019-05-31
Pass $post_id twice in ep_post_sync_kill for backwards compatibility

### LINKED ISSUE #1351: Fatal error with ep_post_sync_kill filter in 3.0
__Have you searched for similar issues before submitting this one?__
Yes

__Is this a bug, question or feature request?__
~~undocumented backwards compatibility change~~ this is documented in the changelog for v3.
I would say this is a bug as it is a simple (yet serious) error to mitigate.

__Describe the issue you encountered:__
After upgrading to v3, I get a fatal error from a callback for the `ep_post_sync_kill` filter. In v2, it provided 3 arguments to the callback, which my callback is expecting, and in v3 it only passes 2.

v2
https://github.com/10up/ElasticPress/blob/ca929db2869576b5e1975436eda4d21f150e50b3/classes/class-ep-sync-manager.php#L268

v3
https://github.com/10up/ElasticPress/blob/b3d96a72dd54db1def357d6c0f093a2a182b6e06/includes/classes/Indexable/Post/SyncManager.php#L161

__Current WordPress version?__
5.2.1

__Current ElasticPress version?__
3.0.2

### Error

```
Too few arguments to function ...(), 2 passed in /var/www/html/wp-includes/class-wp-hook.php on line 286 and exactly 3 expected
```

aaemnnosttv:
### Description of the Change

v2 passes $post_id as the 3rd argument. This also prevents a fatal for existing filters that expect 3 arguments to be passed.

v2 implementation
https://github.com/10up/ElasticPress/blob/ca929db2869576b5e1975436eda4d21f150e50b3/classes/class-ep-sync-manager.php#L268

### Benefits

Updates to v3 from v2 won't trigger a fatal error for common `ep_post_sync_kill` filter usage.

### Possible Drawbacks

Developer/maintainer may not be aware of the change to this filter.

### Applicable Issues

Resolves #1351

### DIFF
--- a/includes/classes/Indexable/Post/SyncManager.php
+++ b/includes/classes/Indexable/Post/SyncManager.php
@@ -80,7 +80,7 @@ class SyncManager extends SyncManagerAbstract {
 			$indexable_post_types = $indexable->get_indexable_post_types();
 
 			if ( in_array( $post_type, $indexable_post_types, true ) ) {
-				if ( apply_filters( 'ep_post_sync_kill', false, $object_id ) ) {
+				if ( apply_filters( 'ep_post_sync_kill', false, $object_id, $object_id ) ) {
 					return;
 				}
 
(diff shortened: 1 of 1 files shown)


tlovett1:
Great idea, @aaemnnosttv. Thank you.

Issue #1270, opened 2019-03-04
Ability to disable rendering of post/pages when indexing.


evanmcmahonf:
Have you searched for similar issues before submitting this one? Yes.

Is this a bug, question or feature request? Feature Request (Happy to create a PR)

Describe the issue you encountered: 

Our site is heavily dynamic, and most of this is driven by shortcodes.
This means on development environments rendering a post/page can be slow (> 1s).

When indexing, the plugin attempts to render these posts to store them in the `post_content_filtered` property.

Due to the dynamicness of our site, we wish to turn of this rendering, both for performance reasons and and due to the search being done on outdated content.

Current WordPress version: 4.9.9

Current ElasticPress version: 2.8.1

Current Elasticsearch version: 6.3

Where do you host your Elasticsearch server: AWS

Feature:
Add the ability through a filter to be able to turn of rendering of the post content

See: https://github.com/10up/ElasticPress/blob/b3d96a72dd54db1def357d6c0f093a2a182b6e06/includes/classes/Indexable/Post/Post.php#L218

I'd be happy to open a PR if you are happy with the idea.

Thanks a lot!

oscarssanchez:
Thanks @evanmcmahonf for your enthusiasm to make the plugin better, I will bring this to our standup next week. 

Kind regards,

evanmcmahonf:
No worries :)
In the meantime, we've just commented out the `apply_filters` and set it to `''`.

oscarssanchez:
Closing per #1358 :)

PR #1358, opened 2019-06-10
Add/post content filtered indexing

oscarssanchez:
Hi @allan23 ,

Could you please review this PR to provide functionality for #1270 ? This adds a flag before and if set to true will function as it was before, but if set to false, it will pass an empty string as the content.

Thanks


oscarssanchez:
Assigning to @tlovett1

PR #1359, opened 2019-06-11
Refactor admin notice , admin screen "resolver", and install path logic

tlovett1:
This PR does not touch any user-facing APIs. It solves a few problems to ultimately make the plugin more stable/maintainable:

1. Admin notices are buggy. Weird notices are showing in weird places. The logic is a house of cards currently.
2. Determining what "screen" is being viewed (settings/dashboard/install) is repetitive.
3. The install "path" logic (what step are we on?) is too complicated and a very hard to modify.

I've refactored logic into three classes: `Screen`, `AdminNotices`, `Installer`.

`Screen` is a simple API to get the current EP screen (either dashboard, settings, or install).  This class is very useful for notices and the dashboard code.

`AdminNotices` is a much more thoughtful way of determine what notices should show on a given screen. It makes use of the `Screen` API to figure out where the user is.

`Installer` is a simple API to determine which install step the user is on. The steps are the same as before.

The PR includes tests for all the new classes.


allan23:
@tlovett1 let's discuss this next week at the meeting.

oscarssanchez:
Hi @tlovett1 ,

It looks like this is working as it used to. Some feedback specific for elasticpress.io clients which i think is not on the scope of this refactor, maybe this will be better to address in another ticket (?) @brandwaffle  

When you input an elasticpress.io host, it redirects you to the 3rd step, assuming the host is set correctly, however, with elasticpress.io you have to input your token and subscription id to set the host successfully. I think this can confuse clients when using the plugin for the first time: 

Video:
https://youtu.be/05009MxrgTk

Issue #1360, opened 2019-06-11
Multisite: Not Allowed to Access Page After Sync


allan23:
__Is this a bug, question or feature request?__
Bug

__Describe the issue you encountered:__
As an end user on a WordPress 5.2.1 multisite, when I try to run a dashboard sync from the ElasticPress dashboard (network page), I get redirected to a URL that displays: 
`Sorry, you are not allowed to access this page.`

This is due to a bug in ElasticPress where if the EP Intro page is not shown, the user will get redirected to it. In this case, the redirect goes to the single site URL instead of the network/multisite one.

Here is a video showing the issue:
https://youtu.be/FSbuVgutf-4

PR #1361, opened 2019-06-11
Setting correct intro URL depending on single site or multisite

allan23:
Described in #1360.


tlovett1:
This is going to have to be slightly reworked one #1359 is merged.

PR #1362, opened 2019-06-11
Enhancement: Control Index-ability of Individual Sites in Multisite

### LINKED ISSUE #1212: Create a Network Admin Dashboard menu enabling users to choose whether a site should be indexed
Previously in 2.7, we pushed some code that automatically set sites with a non-public value in the wp_sites table to "do not index." While this seemed like a good approach at the time, we've had significant feedback from people who have non-public sites that they do want to index, for a variety of reasons including testing behind a login and indexing an Intranet-style site.

We will be rolling back the setting that causes all non-public sites not to index. In order to address the issue of enabling users to set a single or set of sites within a network install to not index, we need a UI that will accommodate this, and then we can execute on the engineering to build it in.

allan23:
This PR introduces a switch toggle next to each site on the Sites Network Admin page allowing the user to select which sites should get indexed with ElasticPress.

By default, ElasticPress allows all sites on the network to become indexed. This can be problematic for private/internal, deleted, spam, or archived sites that the user may not wish to index. This PR gives the user the ability to turn on or off sites they want indexed and disables indexing on sites marked as deleted, archived, or spam.

Screenshot

### Applicable Issues

This PR is to resolve #1212.


includes/dashboard.php
```diff
@@ -1287,3 +1311,61 @@ function use_language_in_setting( $language = 'english' ) {
...
+
+/**
+ * Add column to sites admin table.
+ *
+ * @param string[] $columns Array of columns.
+ *
+ * @return string[]
+ */
+function filter_blogs_columns( $columns ) {
+	$columns['elasticpress'] = 'ElasticPress Indexing';
```
tlovett1: Missing internationalization.

tlovett1:
Overall looks fine. Need to fix linting errors.

oscarssanchez:
Hi @allan23  This is working correctly for me

tlovett1:
When all sites are turned off, it doesn't seem to index users.

tlovett1:
When all sites are turned off, it doesn't seem to index users.

oscarssanchez:
Also resolved lint issues @tlovett1

PR #1363, opened 2019-06-12
Fixed search results to avoid multiple reflows.

fabianmarz:
### Description of the Change
- Currently the suggested search results get inserted in their search result list one by one which is an expensive operation as it grows.
- The PR changes this behaviour by prepare a concatenated string list and append the elements to the list once.

### Benefits
- Avoid performance bottlenecks

### Verification Process
- To build the files I ran `$ gulp js`

### Checklist:


tlovett1:
Looks good.

PR #1367, opened 2019-06-14
Added filter for bulk index action args.

fabianmarz:
### Description of the Change
- This PR adds the possibility to filter the bulk index action/metadata arguments.

### Benefits
- When working with the join/relation field we need to route the child to the same shard as the parent. When index one by one we can achieve this by adding the `?routing` argument. As this is not working with the Bulk API we need to define the routing in the action/metadata line per document.

### Possible Drawbacks
- Potential performance drawbacks due to new hook registration and another `wp_json_encode` function call

### Verification Process
Tested by register a new filter:
```php
add_filter('ep_bulk_index_action_args', function ($args, $document) {
  if (!empty($document['post_parent'])) {
    $args['index']['routing'] = $document['post_parent'];
  }
  return $args;
}, 10, 2);
```

```
$ wp elasticpress index --post-ids=126284,125924 --setup

GET /myindex/_search
{
  "query": {
    "bool": {
      "must": [
        {
          "has_child": {
            "type": "product_variation",
            "query": {
              "match": {
                "post_excerpt": "natur"
              }
            }
          }
        }
      ]
    }
  }
}
```

### Checklist:

### DIFF
--- a/includes/classes/Indexable.php
+++ b/includes/classes/Indexable.php
@@ -232,10 +232,15 @@ abstract class Indexable {
 		$body = '';
 
 		foreach ( $object_ids as $object_id ) {
-			$body .= '{ "index": { "_id": "' . absint( $object_id ) . '" } }' . "\n";
+			$action_args = array(
+				'index' => array(
+					'_id' => absint( $object_id ),
+				),
+			);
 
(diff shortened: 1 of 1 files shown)


tlovett1:
Looks good.

PR #1369, opened 2019-06-18
Media Library Search Integration

allan23:
### Description of the Change
This PR allows a user to enable Elasticsearch integration in the Media Library from the Protected Content settings.

### DIFF
--- a/includes/classes/Feature/ProtectedContent/ProtectedContent.php
+++ b/includes/classes/Feature/ProtectedContent/ProtectedContent.php
@@ -33,6 +33,7 @@ class ProtectedContent extends Feature {
 		$this->title = esc_html__( 'Protected Content', 'elasticpress' );
 
 		$this->requires_install_reindex = true;
+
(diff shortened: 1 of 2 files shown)


tlovett1:
@allan23 anyway to get a few simple unit tests?

tlovett1:
This won't work. Currently attachments are only indexed if documents is enabled.

oscarssanchez:
I believe this has been implemented already, shall we close this one?

archon810:
As per @oscarssanchez above, is this PR still necessary or media search has already been implemented?

brandwaffle:
@archon810 if you activate Protected Content all media titles will be searched. If you also activate Documents then the content of the attachments (assuming there is any) would also be searchable. This PR is definitely outdated and we'll be looking at closing it in our weekly meeting.

brandwaffle:
Media search is now a part of the core plugin.

PR #1372, opened 2019-06-20
Whitelist data-feature & data-field-name html attributes. Resolves #1370

### LINKED ISSUE #1370: Hitting error when saving changes through dashboard
__Is this a bug, question or feature request?__
Bug

__Describe the issue you encountered:__
Hitting javascript error in console whenever I try to change and save settings through the dashboard. Chrome 74.0.3729.169.

__Current WordPress version?__
v4.8.3

__Current ElasticPress version?__
v3.0.3

__Current Elasticsearch version?__
v6.5.4

__Where do you host your Elasticsearch server?__
bonsai.io

__What ElasticPress features do you have currently active?__
Default installation

__Steps to reproduce:__

1. Install the plugin through Wordpress Plugins. Update settings to include Elasticsearch server. Start indexing. All fine up to this step.
2. Go to Elasticsearch dashboard. Turn off Post Search, click Save, and it won't save successfully. The error is available in Javascript console.

__Screenshots, if needed:__
https://ibb.co/zRXsWVy
https://ibb.co/rwsBrwT

mnazwan:
### Applicable Issues
Resolves #1370

### DIFF
--- a/includes/dashboard.php
+++ b/includes/dashboard.php
@@ -65,18 +65,19 @@ function filter_allowed_html( $allowedtags, $context ) {
 		$ep_tags = $allowedposttags;
 
 		$atts = [
-			'type'        => true,
-			'checked'     => true,
-			'selected'    => true,
-			'disabled'    => true,
-			'value'       => true,
-			'class'       => true,
-			'data-*'      => true,
-			'id'          => true,
-			'style'       => true,
-			'title'       => true,
-			'name'        => true,
-			'placeholder' => '',
+			'type'            => true,
+			'checked'         => true,
+			'selected'        => true,
+			'disabled'        => true,
+			'value'           => true,
+			'class'           => true,
+			'data-*'          => true,
+			'data-field-name' => true,
+			'id'              => true,
+			'style'           => true,
+			'title'           => true,
+			'name'            => true,
+			'placeholder'     => '',
 		];
 
 		$ep_tags['input']    = $atts;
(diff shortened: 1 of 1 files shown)


includes/dashboard.php
```diff
@@ -65,18 +65,19 @@ function filter_allowed_html( $allowedtags, $context ) {
...
-			'title'       => true,
-			'name'        => true,
-			'placeholder' => '',
+			'type'            => true,
+			'checked'         => true,
+			'selected'        => true,
+			'disabled'        => true,
+			'value'           => true,
+			'class'           => true,
+			'data-*'          => true,
```
andersrehn: I get the feeling that `data-*` does not work as expected in at least a wordpress `4.9.10` version. 

Expected behaviour is to allow all data- tags. 
Actual result is that you must for each given data key allow them.

includes/dashboard.php
```diff
@@ -94,6 +95,8 @@ function filter_allowed_html( $allowedtags, $context ) {
 			'target'         => true,
 		];
 
+		$ep_tags['a']['data-feature'] = true;
```
andersrehn: This does indeed solve the issue that the console is broken. However the dashboard does not save any data even if the UI tells you it did. 

If we add `$ep_tags['input']['data-field-name'] = true;` to the line below, the data will be saved as well.

mnazwan: Changes required for ```<input>``` already covered in default whitelisted attributes, Line 75.

includes/dashboard.php
mnazwan: Agree. This is just a quick fix.

tlovett1:
Pushed a commit to develop very similar to this but added a few more attributes. Thanks guys.

Issue #1373, opened 2019-06-20
Display an admin warning if nodes < 2


brandwaffle:
Per the Elasticsearch best practices we should show a notice to users in the Admin Dashboard if they're running an Elasticsearch setup with only 1 node. Using our new notice system, this should be dismissable in the Admin but mandatory on ElasticPress dashboard screens.

In general, this warning should notify if status != green (a single node automatically triggers yellow status).

We also want to link to the new Health dashboard so folks can track down where the issue is occurring.

oscarssanchez:
Closing as this has been implemented already

PR #1374, opened 2019-06-21
Add new submenu page "Index Health"

oscarssanchez:
### Description of the Change
This PR adds a new elasticpress submenu page called _Index health_ which enables users to check basic data of their elasticsearch indices in a meaningful way. Options to check are current indices names and health, total documents on each index, averages on query/indexing times and total of operations. ChartJS was used to make all charts. Finally there's a section to check the useful total data about their cluster, like total memory, total documents and total size.

ep12

Screen Shot 2019-06-21 at 1 07 12 PM

cc @allan23 @brandwaffle @tott @tlovett1 

### Benefits
This enables users to be more alert of their indices.

### Checklist:


dustinrue:
I like the graphs.

Maybe the status messages should read a bit differently though? I'm not sure "Green" means much to people. Maybe something like this:

* Green -> Healthy
* Yellow -> Degraded
* Red -> Unavailable

Healthy is easy enough to understand. 

An index can be "degraded" (yellow) if the underlying cluster is in the process of a rolling restart, upgrade or anything that would cause it to be shuffling shards around. It means that the data is available but performance may not be great right now.  Also, an index will be yellow if it is running on a single server and Elasticsearch is unable to properly distribute replica shards. In this situation where nodes == 1 but replicas > 0 the index status will be yellow. The end user can resolve this state by either indexing with no replicas or by increasing the number of cluster nodes.

Red status means the cluster is responding but something is wrong with their index is not available.

oscarssanchez:
Addressed #1373 on last commit. @brandwaffle . However, based on the above comment from @dustinrue, it seems like yellow status can also mean other stuff is happening besides running on a single node. To check nodes we can use https://www.elastic.co/guide/en/elasticsearch/reference/current/cluster-nodes-stats.html , however, on elasticpress.io this endpoint throws a 403 forbidden. How should we address this ? 

I agree the proposed statuses sound more meaningful @dustinrue, thanks for taking the time to look at it :) . @brandwaffle  should we address and change ?

tlovett1:
Is this done? Needs a few unit tests.

tlovett1:
@oscarssanchez I'm going to recreate this PR without the chartjs source files. Those files bloat the repo and aren't needed.

Issue #971, opened 2017-11-08
Support for 'relation' in tax_query inner arrays


tirriainen:
Have you searched for similar issues before submitting this one?
Yes
Is this a bug, question or feature request?
Feature request

Is there a way to use 'relation' in tax_query (and meta_query) inner arrays? What I'm trying to achieve is something like this:

```
$args['tax_query'] => [
	'relation' => 'and'
	[
		'taxonomy' => TAX1_TAX_NAME,
		'terms'    => 1,					
	],
	[
		'relation' => 'or',
		[
			'taxonomy' => TAX2_TAX_NAME,
			'terms'    => 2,							
		],
		[
			'taxonomy' => TAX3_TAX_NAME,
			'terms'    => 3,							
		]				
	]
];
```

eugene-manuilov:
Hi @tirriainen! Thanks for submitting this. We will work on it.

PR #1378, opened 2019-06-25
Add support for nested taxonomy queries

dkotter:
### Description of the Change

Currently, when building out taxonomy query information to send to elasticsearch, we only account for a single level of nesting. But WordPress supports taxonomy queries that are nested, so that information ends up being lost and the query results don't match what we want.

This PR fixes that by recursively looping through the passed in `tax_query` information. This ensures that we process all levels of nesting that may exist. The majority of the code here is the same as before, as far as building out the query. I've added the recursive looping and changed slightly how we set this tax query information in our main elasticsearch query.

### Benefits

Nested tax queries currently do not work, so adding this feature brings us closer in line to the core functionality WordPress supports

### Possible Drawbacks

Logic to build out the tax query is slightly changed, so there's always a chance of having missed some use case that this might break.

### Verification Process

I ran a WP_Query that contained a nested tax query, first without ElasticPress enabled for that query and then with ElasticPress enabled. I verified the results in both instances matched. I also did the same test a few other times with non-nested tax queries, to ensure those still worked.

For reference, here's one of the nested queries I tested with:
```
$args = array(
	'ep_integrate' => 1,
	'post_type'    => 'post',
	'tax_query'    => array(
		'relation' => 'OR',
		array(
			'taxonomy' => 'category',
			'field'    => 'slug',
			'terms'    => array( 'category-1' ),
		),
		array(
			'relation' => 'AND',
			array(
				'taxonomy' => 'post_tag',
				'field'    => 'slug',
				'terms'    => array( 'tag2' ),
				'operator' => 'NOT IN',
			),
			array(
				'taxonomy' => 'category',
				'field'    => 'slug',
				'terms'    => array( 'category-2' ),
			),
		),
	),
);
```
### Checklist:

### Applicable Issues

#971

### DIFF
--- a/includes/classes/Indexable/Post/Post.php
+++ b/includes/classes/Indexable/Post/Post.php
@@ -546,121 +546,23 @@ class Post extends Indexable {
 		}
 
 		if ( ! empty( $args['tax_query'] ) ) {
-			$tax_filter          = [];
-			$tax_must_not_filter = [];
-
 			// Main tax_query array for ES.
 			$es_tax_query = [];
 
-			foreach ( $args['tax_query'] as $single_tax_query ) {
-				if ( ! empty( $single_tax_query['taxonomy'] ) ) {
-					$terms = isset( $single_tax_query['terms'] ) ? (array) $single_tax_query['terms'] : array();
-					$field = ( ! empty( $single_tax_query['field'] ) ) ? $single_tax_query['field'] : 'term_id';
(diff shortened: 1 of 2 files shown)


thomasdiesenreiter:
He, just tested it, works great! Looking forward to get this into a release, we really need that one =)

tlovett1:
Hey @dkotter, this definitely needs a few unit tests. Mind adding them?

tlovett1:
Nevermind. I added tests.

PR #1383, opened 2019-07-01
make user mapping compatible with earlier versions than 5.0

mustafauysal:
### Description of the Change
Currently, ElasticPress supports Elasticsearch 1.7. However, user mapping is not compatible with the earlier versions due to `text` and `keyword` support. – https://www.elastic.co/blog/strings-are-dead-long-live-strings 

In these changes, backporting user mapping with `string` type.  

### Benefits
Supporting user search to existing Elasticpress setups without worry about the version.

### Possible Drawbacks
Probably, we don't want to maintain the mapping for legacy setups. Instead, we can update the related part of the documentation. (User feature requires WordPress 5.1+ and Elasticsearch 5.0+) 

### Verification Process
run `wp elasticpress index --setup` with the earlier versions of Elasticsearch (5.0 <)

### Checklist:

### DIFF
--- a/includes/classes/Indexable/User/User.php
+++ b/includes/classes/Indexable/User/User.php
@@ -645,7 +645,18 @@ class User extends Indexable {
 	 * @return boolean
 	 */
 	public function put_mapping() {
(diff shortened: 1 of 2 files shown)


tlovett1:
Thanks @mustafauysal !

PR #1398, opened 2019-07-12
Support ES 7.0+

tlovett1:
This PR adds functionality to EP to make it compat with ES 7.0. CC @cmmarslender @brandwaffle


tlovett1:
@cmmarslender should be fixed now

PR #1402, opened 2019-07-13
Only hooks up certain query altering actions/filters if we should be …

### LINKED ISSUE #1332: Installing Elasticpress prevents searching orders with customer name
I noticed that when I have Elasticpress installed there are some issues with orders view in the backend. For example searching the orders by customer name isn't working when Elasticpress is active (protected content featured disabled). If I enable the protected content indexing, then the search with customer name does work but with this setting I'm also seeing problems where filtering orders by status there are orders displayed with other statuses as well.

Using WC 3.6.3, Elasticpress 3.0.1, Elasticsearch 5.6.16

__Features enabled:__ Post search, woocommerce, facets, related, protected content

__Steps to reproduce (order search):__

1. Enable Elasticpress with protected contend disabled
2. Search orders with customer names in backend orders view => search doesn't give any results and search seems to work only with orderID

Main point is enabling Elasticpress plugin shouldn't break existing Woocommerce features.

cmmarslender:
…integrating the current query with Elasticsearch to avoid breaking non-ES queries

Fixes #1332


includes/classes/Feature/WooCommerce/WooCommerce.php
```diff
@@ -237,28 +237,7 @@ public function disallow_duplicated_query( $value, $query ) {
 	 * @since  2.1
 	 */
 	public function translate_args( $query ) {
-
```
tlovett1: Seems like there are other conditions for determining whether the query should be integrated with that are not moved to the new method.

cmmarslender: I'm not seeing them. was a copy/paste from the old method to this one: https://cloudup.com/cBFOf5IfG4d

PR #1438, opened 2019-08-13
fix incorrect author_name usage.

willgladstone:
### Description of the Change
The `author_name` query var is meant to use the [user nicename](https://developer.wordpress.org/reference/classes/wp_query/#author-parameters). 

Currently the Post indexable is setting this value to the display_name. This means we can't search using this field without altering the query via the `ep_formatted_args` hook.

### Alternate Designs
I currently have this code in place to force the code to work as expected, but it will need to loop twice over the args to ensure that the indexes for both `post_filter` and `aggs` are the same. I assume they are, but I would also not be surprised if they weren't.
```
function set_search_args( $formatted_args, $args ) {
	if ( ! empty( $args['s'] ) && ! empty( $args['author_name'] ) ) {
		foreach ( $formatted_args['post_filter']['bool']['must'] as $i => $filter ) {
			if ( isset( $filter['term']['post_author.raw'] ) ) {
				unset( $formatted_args['post_filter']['bool']['must'][ $i ]['term']['post_author.raw'] );
				$formatted_args['post_filter']['bool']['must'][ $i ]['term']['post_author.login.raw'] = $args['author_name'];
			}
		}
		foreach ( $formatted_args['post_filter']['bool']['must'] as $i => $filter ) {
			if ( isset( $filter['term']['post_author.raw'] ) ) {
				unset( $formatted_args['aggs']['terms']['filter']['bool']['must'][ $i ]['term']['post_author.raw'] );
				$formatted_args['aggs']['terms']['filter']['bool']['must'][ $i ]['term']['post_author.login.raw'] = $args['author_name'];
			}
		}
	}
	return $formatted_args;
}
```
### Benefits

It will work as expected.

### Possible Drawbacks

I believe this is relatively non-invasive.

### Verification Process

This is tested on an local docker instance running ES as well as an instance over at elastic.co. I used the debugging log to compare queries and ran local curl calls with altered queries to find the solution. After making changes, the queries are now returning posts of only a specific author slug.

### Checklist:

- [ x ] I have read the [**CONTRIBUTING**](/CONTRIBUTING.md) document.
- [ x ] My code follows the code style of this project.

### Applicable Issues
I have not checked.

### DIFF
--- a/includes/classes/Indexable/Post/Post.php
+++ b/includes/classes/Indexable/Post/Post.php
@@ -677,9 +677,12 @@ class Post extends Indexable {
 
 			$use_filters = true;
 		} elseif ( ! empty( $args['author_name'] ) ) {
+			// Since this was set to use the display name initially, there might be some code that used this feature.
+			// Let's ensure that any query vars coming in using author_name are in fact slugs.
+			$author_login = sanitize_user( $args['author_name'] );
 			$filter['bool']['must'][] = array(
 				'term' => array(
-					'post_author.raw' => $args['author'],
+					'post_author.login.raw' => $author_login,
 				),
 			);
 


tlovett1:
I'm not understanding how this would work. `post_author.raw` stores the same thing as `post_author.login.raw`: 
https://github.com/10up/ElasticPress/blob/develop/includes/mappings/post/7-0.php
https://github.com/10up/ElasticPress/blob/develop/includes/classes/Indexable/Post/Post.php#L171

willgladstone:
Yes correct, it is supposed to be a the login. But it is passing it the author ID. I put the sanitation checks in place, but are you suggesting they aren't necessary and we just need to switch `$arg['author]` to `$arg['author_name']`? Or am I missing something? This is on query, not on index.

willgladstone:
Looking closer now, I guess you saw the change of the variable assignment, which I get was confusing. I did it for clarity purposes as author_name is a variable name that makes little sense to begin with. Using it's actual assignment makes more sense than the redundant post_author.raw which is not clear what it is holding. But that isn't the issue, the issue is assigning it an author ID. I'm happy to put it back to the original raw instead of the login, but it makes no difference in the outcome and it's more readable the way it is I believe.

PR #1443, opened 2019-08-16
Add a Terms Indexable

dkotter:
### Description of the Change

A [new filter](https://core.trac.wordpress.org/ticket/41246) was added to WordPress core (coming in 5.3), that allows term queries to be short-circuited with your own results. This provides the ability for ElasticPress to take over any queries that utilize `WP_Term_Query`.

This PR adds a new Terms Indexable that will index all terms across a site and then takes over term queries if `ep_integrate` is set to true.

### Benefits

Terms will now be stored in their own index and they can be searched across. Any queries that make use of `WP_Term_Query` can gain potential performance benefits by utilizing elasticsearch over a normal database query.

### Possible Drawbacks

Not all queries will be made faster by this integration, so testing will need to be done to determine if it's worth using elasticsearch for term queries.

Anytime objects are updated that have terms associated with them, we re-index those terms, resulting in possibly longer save runs.

### Verification Process

Tested these changes locally on a multisite install, with 4 sites on the network. Each site had a handful of categories and tags set; some nested, some not; some with term meta, most not; most associated with at least one object, some associated with nothing.

- Verified that running a full index will add all term information into a new terms index within elasticsearch.
- Verified that adding a new term or updating an existing term will add/update that term within the correct index
- Verified that running a `WP_Term_Query`, with `ep_integrate` set to true, runs the query against elasticsearch
- Tested all the various [query params](https://developer.wordpress.org/reference/classes/WP_Term_Query/__construct/#parameters) that `WP_Term_Query` supports, to make sure those work when run against elasticsearch

Note further testing is recommended here, to further ensure all the various query params work, especially when used in conjunction with each other. Also more testing around updating terms and adding/updating posts that use terms, to make sure things are updated correctly, particularly the term counts.

### Checklist:


jeffpaul:
@brandwaffle @tlovett1 @tott please let me know once this gets through code review as we'll look to have someone else from the team help with any changes from that review... thanks!

jeffpaul:
@tlovett1 the two build failures appear related to the weighting work.  Are you seeing weighting-related issues in this PR or is this good to be merged in alongside the WP 5.3 release in November?  Essentially, do you need anything else from the team?

tlovett1:
@jeffpaul we are good to go for now. Thank you for you and @dkotter's help!

tlovett1:
@petenelson what's the status of unit tests on this one? Thanks.

petenelson:
With the latest `develop` branch merged in, there were a couple errors in the unit tests when running locally.

```
1) ElasticPressTest\TestTerm::testTermQueryOrderParent
Undefined offset: 0

/var/www/html/wp-content/plugins/elasticpress/tests/php/indexables/TestTerm.php:500

2) ElasticPressTest\TestTerm::testBasicTermQuery
Failed asserting that null is true.

/var/www/html/wp-content/plugins/elasticpress/tests/php/indexables/TestTerm.php:220
```
I'll take a look at these, as well as code coverage for the rest of the term unit tests.

Issue #1471, opened 2019-08-28
Validate mapping currently in index against expected version


tott:
Sometimes the mapping fails or gets outdated. It would be good to detect outdated mappings and provide feedback in the admin interface.

tlovett1:
@tott can you explain to me what this is intended to solve? If someone upgrades Elasticsearch, show them a warning they need to reindex? Seems pretty edge case. It would be nice to have a unit test and include as part of a major release.

tott:
This is in relation to #1466.We have had a good amount of cases where things are failing due to invalid mapping, not only from this issue but also prior to this. This will help identify when the mapping is outdated and a reindex should occur

PR #1472, opened 2019-08-28
Feature implementation for #1471

tott:
This changeset implements a way to determine the mapping version currently on the index.
It compares it against the expected mapping version and displays a
dismissable notice.
The notice dismissal is dependent on the es_version
This changeset also introduces the _meta mapping to store the mapping
version for a more predicatble detection


tott:
This fixes #1471

includes/classes/AdminNotices.php
```diff
@@ -571,6 +573,61 @@ protected function process_host_error_notice() {
...
+			$dismiss = get_site_option( 'ep_hide_maybe_wrong_mapping_notice', false );
+		} else {
+			$dismiss = get_option( 'ep_hide_maybe_wrong_mapping_notice', false );
+		}
+
+		if ( $dismiss == $es_version ) {
+			return false;
+		}
+
+		$mapping_file_wanted = '5-2.php';
```
tott: Note to self - Refactor in next version to abstract this out to dedicated function

tott:
Tests are currently failing due to the way other tests are structured. Will need to change admin notices test cases to properly reflect the expected mapping version and create new test cases for the version notice

includes/classes/AdminNotices.php
```diff
@@ -571,6 +573,72 @@ protected function process_host_error_notice() {
...
+
+		$mapping_file_wanted = Indexables::factory()->get( 'post' )->get_mapping_name();
+
+		$post_index           = Indexables::factory()->get( 'post' )->get_index_name();
+		$mapping_file_current = Elasticsearch::factory()->determine_mapping_version( $post_index );
+		if ( is_wp_error( $mapping_file_current ) ) {
+			return false;
+		}
+
+		if ( ! $mapping_file_current || $mapping_file_wanted != $mapping_file_current ) {
```
tlovett1: Should use !==

includes/classes/AdminNotices.php
```diff
@@ -571,6 +573,72 @@ protected function process_host_error_notice() {
...
+			$html = sprintf( __( 'It seems the mapping data in your index does not match the Elasticsearch version used. We recommend to reindex your content using the sync button on the top of the screen or through wp-cli by adding the <i>--setup</i> flag.', 'elasticpress' ) );
```
tlovett1: Should use `em` instead of `i`

includes/classes/AdminNotices.php
```diff
@@ -571,6 +573,72 @@ protected function process_host_error_notice() {
...
+
+			if ( defined( 'WP_DEBUG' ) && WP_DEBUG ) {
+				$html .= '<span class="notice-error-es-response-code"> ' . sprintf( __( 'Current mapping: %1$s , Expected mapping: %2$s', 'elasticpress' ), esc_html( $mapping_file_current ), esc_html( $mapping_file_wanted ) ) . '</span>';
```
tlovett1: Should use `esc_html__`

includes/classes/AdminNotices.php
```diff
@@ -571,6 +573,72 @@ protected function process_host_error_notice() {
...
+	 * Dismiss: Always dismissable per es_version as custom mapping could exist
+	 * Show: All screens
+	 *
+	 * @since   3.1.5
+	 * @return array|bool
+	 */
+	protected function process_maybe_wrong_mapping_notice() {
+		// we might have this dismissed
+		if ( defined( 'EP_IS_NETWORK' ) && EP_IS_NETWORK ) {
+			$dismiss = get_site_option( 'ep_hide_maybe_wrong_mapping_notice', false );
```
tlovett1: We never check `$dismiss` to see if we should not show. This notice will not be dismissable.

tlovett1 (dismissed):

petenelson:
@tlovett1 @tott I'm going to take a look at any final cleanup needed for this before we finish/merge the code review

oscarssanchez:
Should we add the same for other mappings we have? (terms, user, etc) and, if we do it, we should probably rename the new introduced filters. Also needed to update the since tags.

includes/classes/AdminNotices.php
```diff
@@ -592,6 +594,73 @@ protected function process_host_error_notice() {
...
+		// we need a host
+		$host = Utils\get_host();
+		if ( empty( $host ) ) {
+			return false;
+		}
+
+		// we also need a version
+		$es_version = Elasticsearch::factory()->get_elasticsearch_version( false );
+
+		if ( false === $es_version || $dismiss === $es_version ) {
```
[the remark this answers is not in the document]
felipeelia: @oscarssanchez it seems `$dismiss` is supposed to be a boolean, isn't it? I mean, it seems it really should have its own conditional, like in [here](https://github.com/10up/ElasticPress/blob/f3ef9d080b8e7dc3522014974f796a5ba1cb0af2/includes/classes/AdminNotices.php#L109).

includes/classes/AdminNotices.php
```diff
@@ -592,6 +594,73 @@ protected function process_host_error_notice() {
...
+		$mapping_file_wanted = Indexables::factory()->get( 'post' )->get_mapping_name();
+
+		$post_index           = Indexables::factory()->get( 'post' )->get_index_name();
+		$mapping_file_current = Elasticsearch::factory()->determine_mapping_version( $post_index );
+		if ( is_wp_error( $mapping_file_current ) ) {
+			return false;
+		}
+
+		if ( ! $mapping_file_current || $mapping_file_wanted !== $mapping_file_current ) {
+			$html = sprintf( '%1$s <em>--setup</em> %2$s', esc_html__( 'It seems the mapping data in your index does not match the Elasticsearch version used. We recommend to reindex your content using the sync button on the top of the screen or through wp-cli by adding the', 'elasticpress' ), esc_html__( 'flag.', 'elasticpress' ) );
```
felipeelia, line 648: In terms of i18n, this string isn't great. It would be much better for translators if this were:
```
$html = sprintf(
	/* translators: 1. <em>; 2. </em> */
	esc_html__( 'It seems the mapping data in your index does not match the Elasticsearch version used. We recommend to reindex your content using the sync button on the top of the screen or through wp-cli by adding the %1$s--setup%2$s flag', 'elasticpress' ),
	'<em>',
	'</em>'
);
```

felipeelia (dismissed):

includes/classes/AdminNotices.php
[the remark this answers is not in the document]
felipeelia: Actually, I was wrong. With the change introduced in `dismiss_notice()`, this really makes sense as is. In fact, `$dismiss` will receive the es_version, so it can be displayed again if it was dismissed before but for a different ES version.

felipeelia:
@oscarssanchez, I've fixed WP Acceptance tests in #2281 and pushed a commit addressing a few things here. Do you mind checking everything again?

oscarssanchez (approved):

oscarssanchez:
Looks good @felipeelia

Issue #1426, opened 2019-07-30
WPQuery with tags dont work


altendorfme:
**Wordpress version:** 5.2.2
**ElasticPress version:** 3.1.1 (Reindexed)
**Kinsta Elastic**

category__and, category__in work correctly, but if we try to use tag__and, tag__in or tag_id, it does not return the posts correctly, below is an example of WP_Query, Querys and the Result.

**WPQUERY:**
```
$tag = array(1, 2)
$args  = array(
	'ep_integrate' => true,
	'post_type'      => 'post',
	'post_status'    => 'publish',
	'tag__and'        => $tag,
	'posts_per_page' => 6
);
```

**QUERY ARGS:**
```
array(66) {
  ["ep_integrate"]=>
  bool(true)
  ["post_type"]=>
  string(4) "post"
  ["post_status"]=>
  string(7) "publish"
  ["tag_id"]=>
  int(361474)
  ["posts_per_page"]=>
  int(6)
  ["error"]=>
  string(0) ""
  ["m"]=>
  string(0) ""
  ["p"]=>
  int(0)
  ["post_parent"]=>
  string(0) ""
  ["subpost"]=>
  string(0) ""
  ["subpost_id"]=>
  string(0) ""
  ["attachment"]=>
  string(0) ""
  ["attachment_id"]=>
  int(0)
  ["name"]=>
  string(0) ""
  ["static"]=>
  string(0) ""
  ["pagename"]=>
  string(0) ""
  ["page_id"]=>
  int(0)
  ["second"]=>
  string(0) ""
  ["minute"]=>
  string(0) ""
  ["hour"]=>
  string(0) ""
  ["day"]=>
  int(0)
  ["monthnum"]=>
  int(0)
  ["year"]=>
  int(0)
  ["w"]=>
  int(0)
  ["category_name"]=>
  string(0) ""
  ["tag"]=>
  string(0) ""
  ["cat"]=>
  string(0) ""
  ["author"]=>
  string(0) ""
  ["author_name"]=>
  string(0) ""
  ["feed"]=>
  string(0) ""
  ["tb"]=>
  string(0) ""
  ["paged"]=>
  int(0)
  ["meta_key"]=>
  string(0) ""
  ["meta_value"]=>
  string(0) ""
  ["preview"]=>
  string(0) ""
  ["s"]=>
  string(0) ""
  ["sentence"]=>
  string(0) ""
  ["title"]=>
  string(0) ""
  ["fields"]=>
[...]

PR #1474, opened 2019-08-28
Add tag_in and tag__and support

oscarssanchez:
### Description of the Change

This PR adds compatibility for `tag__and` and `tag_in` as requested on #1426 by querying elasticsearch` terms.post_tag.term_id` field.

In the case of tag_id, make sure it is always an singular value as described on https://developer.wordpress.org/reference/classes/wp_query/#tag-parameters

> tag_id (int) – use tag id.

And for tag__and, providing a single value or array should be acceptable:

> tag__and (array) – use tag ids.

### Checklist:

### DIFF
--- a/includes/classes/Indexable/Post/Post.php
+++ b/includes/classes/Indexable/Post/Post.php
@@ -579,26 +579,42 @@ class Post extends Indexable {
 		// set tax_query if it's implicitly set in the query.
 		// e.g. $args['tag'], $args['category_name'].
 		if ( empty( $args['tax_query'] ) ) {
-			if ( ! empty( $args['category_name'] ) ) {
-				$args['tax_query'][] = array(
-					'taxonomy' => 'category',
-					'terms'    => array( $args['category_name'] ),
-					'field'    => 'slug',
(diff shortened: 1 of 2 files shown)


tlovett1:
@oscarssanchez this needs a unit test.

oscarssanchez:
Added a Unit test @tlovett1 , can you please review again? Thanks!

includes/classes/Indexable/Post/Post.php
```diff
@@ -579,26 +579,42 @@ public function format_args( $args ) {
...
+						'taxonomy' => 'post_tag',
+						'terms'    => array( $args['tag'] ),
+						'field'    => 'slug',
+					);
+					break;
+				case ! empty( $args['tag__and'] ):
+					$args['tax_query'][]   = array(
+						'taxonomy' => 'post_tag',
+						'terms'    => $args['tag__and'],
+						'field'    => 'term_id',
```
tlovett1: This isn't going to work. I think you need to provide `operator` as `AND`.

PR #1476, opened 2019-08-29
Update to new tag filtering for GH Actions

helen:
They said yesterday they'd fix it "this week", guess they got to it quickly so the last tag failed to deploy. See https://github.community/t5/GitHub-API-Development-and/How-to-restrict-execution-of-GitHub-Actions-workflow-on-tags/m-p/29640#M2671


tlovett1:
Thanks!

PR #1494, opened 2019-09-24
Add filter to support indexing revisions

elliott-stocks:
### Description of the Change
Add support for indexing revisions when updating/deleting posts.

### Checklist:

### Changelog Entry

- Add support for indexing revisions via the `ep_index_revisions` filter.


adamsilverstein:
This looks good @elliott-stocks! Do we need to document this filter anywhere?

elliott-stocks:
Thanks for the quick review, @adamsilverstein ! I don't think we document filters (not that I've seen anyway).

adamsilverstein:
> I don't think we document filters (not that I've seen anyway).

Yea, I didn't see any documentation in the code, I thought perhaps it lived elsewhere :) I'll try to open a issue around adding filter docs.

For the `Distributor` plugin we are working towards generating filter docs based on our inline JSDoc notations (which we would need to add here first), see https://github.com/10up/distributor/pull/448

tlovett1:
@brandwaffle did we figure out what we are doing with revisions?

brandwaffle:
@tlovett1 yes, we're going to implement this change to not index revisions by default but provide a filter if needed.

tlovett1:
Fixed in a slightly different way in #1640

PR #1499, opened 2019-09-25
Query for object IDs when indexing

elliott-stocks:
### Description of the Change
This is pretty similar to the previous pull request prior to 3.x #1145. Currently we query for the entire post object which is unnecessary since we call `get_post()` later on anyway (within `prepare_document()`).

On large databases this has a noticeable difference on the indexing time, especially for posts with a lot of content.

### Benefits
Improves performance of indexing posts.

### Possible Drawbacks
The only potiental breaking change that I've noticed is with the `ep_item_sync_kill` filter. The second parameter was previously the entire object, whereas now it's only the object ID.

Open to suggestions on how we handle backwards compatibly for this.

### Verification Process
Performed a full index via WPCLI and within the dashboard and compared the results to the previous index.

### Checklist:

### DIFF
--- a/includes/classes/Command.php
+++ b/includes/classes/Command.php
@@ -703,25 +703,25 @@ class Command extends WP_CLI_Command {
 
 			if ( ! empty( $query['objects'] ) ) {
 
-				foreach ( $query['objects'] as $object ) {
+				foreach ( $query['objects'] as $object_id ) {
 
 					if ( $no_bulk ) {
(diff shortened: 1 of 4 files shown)


petenelson:
@elliott-stocks I'm going to look at adding some unit tests to this PR before we start the code review and merge process

tlovett1:
Closing in favor of #1640

Issue #1515, opened 2019-10-09
Weighting: Illegal offset type in isset or empty


khelil:
**Describe the bug**
After updating to version 3.1.4 (same with 3.2) i'm getting Warning messages on Search page :

Warning: Illegal offset type in isset or empty in ../elasticpress/includes/classes/Feature/Search/Weighting.php on line 370

Warning: Illegal offset type in .../elasticpress/includes/classes/Feature/Search/Weighting.php on line 389

And i'm not getting search results from Elastic, but the classic WP.
Autosuggest is broken also, got a 404 error.

**Steps to Reproduce**
No steps, just broke after 3.1.4 update. Just updated to 3.2 but got same problems

**Environment information**
 - Device: MacBook
 - OS: Mojave
 - Browser and version: Chrome Version 77.0.3865.90
 - Plugins and version: WPML + ACF
 - Theme and version: Sage 8.6 from Roots.io

**Additional context**
Working with a website running locally on Vagrant, indexing et request on Elasticpress.io

tlovett1:
@khelil does the error happen with WPML deactivated?

khelil:
Yep @tlovett1 no more errors when WPML deactivated and search works fine !
Perhaps an issue with untranslated weight fields ?

tott:
I added an initial changeset to address the symptoms. I'm still evaluating if I can identify the cause.

brandwaffle:
@khelil We'd like to include this in our next release (targeted for today) but we'd need to understand if there are any other factors that are causing this issue before we push the code that works around the symptoms. Can you let us know if you've got any custom code per @tott 's question about? Thanks in advance and thanks for submitting this bug report!

khelil:
@brandwaffle @tott sorry guy'z, just saw your message. We have custom code for theme and plugin, we developed both from scratch, but no code is changing the search feature or results. 
In doubt, if you want, send me you emails i can give you access to our repo on github, for you to check.

khelil:
@brandwaffle @tott no more bug here. After updating WPML to Version 4.3.1, no more notices and search results are as expected ! Closing the issue, Thank you 🙏

PR #1530, opened 2019-10-17
Avoid notices when fields/weights are not as expected

tott:
### Changelog Entry
Avoid notices outlined in #1515.


tlovett1:
@tott has this been tested?

tlovett1:
@tott has this been tested?

tott:
I was still hoping for more feedback on the ticket but given the lack of response I will do some testing and advise when this is good to go.

felipeelia:
Given that the notices came from some (malformed) custom code and no one else complained about it in these years, I'm closing this one. If we see the error again, we can always reopen it.

PR #1531, opened 2019-10-18
Add a Comments Indexable

dkotter:
### Description of the Change

Similar to #1443, a new [filter](https://core.trac.wordpress.org/changeset/46086) was added to WordPress core and is coming in 5.3 that allows comment queries to be short-circuited with your own results. This provides ElasticPress the ability to take over any queries that utilize `WP_Comment_Query`.

This PR adds a new Comments Indexable that will index all comments across a site (regardless of comment type or status) and then takes over `WP_Comment_Query` if `ep_integrate` is true or if the `search` argument is used.

### Benefits

Comments will now be stored in elasticsearch and can be searched across as needed. Any queries that make use of `WP_Comment_Query`, either directly or indirectly, can gain potential performance benefits by utilizing an elasticsearch query over a traditional database query.

### Possible Drawbacks

Not all queries will be made faster by this integration, so testing will need to be done to determine if it's worth using elasticsearch for comment queries in all situations.

### Verification Process

Tested these changes locally on a multisite install, with 4 sites on the network. Each site had a handful of comments, some of these had replies, others did not.

- Verified that running a full index will add all comment information into a new comments index within elasticsearch.
- Verified that adding a new comment or updating an existing comment (in the admin) will add/update that comment within the correct index
- Verified that running a `WP_Comment_Query`, with `ep_integrate` set to true, runs the query against elasticsearch
- Tested all the various query params that `WP_Comment_Query` supports, to make sure those work when run against elasticsearch

Note further testing is recommended here, to further ensure all the various query params work, especially when used in conjunction with each other.

### Checklist:


jeffpaul:
@brandwaffle @tlovett1 @tott please let me know once this gets through code review and if there are any changes needed from that review... thanks!

dinhtungdu:
@tlovett1 I fixed some bugs and increased test coverage to 83.14%

shadyvb:
@dinhtungdu Thanks for adding all the tests!

Can't help but notice they're failing though..

includes/classes/Indexable/Comment/Comment.php
```diff
@@ -0,0 +1,1053 @@
...
+				'count' => true,
+			]
+		);
+
+		$defaults = [
+			'number'  => $this->get_bulk_items_per_page(),
+			'offset'  => 0,
+			'orderby' => 'comment_ID',
+			'order'   => 'desc',
+		];
```
shadyvb, line 729: Need to specify `'type' => 'all'` otherwise EP will only index comments with 'comment' type.
```suggestion
	public function query_db( $args ) {
		$all_query = new WP_Comment_Query(
			[
				'count' => true,
				'type'  => 'all'
			]
		);

		$defaults = [
			'type'    => 'all',
			'number'  => $this->get_bulk_items_per_page(),
			'offset'  => 0,
			'orderby' => 'comment_ID',
			'order'   => 'desc',
		];
```

includes/classes/Indexable/Comment/QueryIntegration.php
```diff
@@ -0,0 +1,336 @@
...
+					'comment_parent',
+					'user_id',
+				]
+			);
+
+			foreach ( $comment_return_args as $key ) {
+				if ( isset( $comment_array[ $key ] ) ) {
+					$comment->$key = $comment_array[ $key ];
+				}
+			}
```
shadyvb, line 216: I think it's important to return `\WP_Comment` objects rather than plain ones, as it's quite possible the consuming code would check the object type.
```suggestion
			$comment = new \WP_Comment( (object) $comment_array );

			if ( ! empty( $comment_array['site_id'] ) ) {
				$comment->site_id = $comment_array['site_id'];
			} else {
				$comment->site_id = get_current_blog_id();
			}
```

includes/classes/Indexable/Comment/Comment.php
dinhtungdu: @shadyvb I can't reproduce your issue. What is the ES version used in your testing? Can you give me steps to reproduce it?

shadyvb: @dinhtungdu I think it makes sense to explicitly query for `all` comments, instead of allowing the possibility that this might break if the consuming code assumes `''` means only the `comment` type ( which holds true in my case, and a few other plugins that does not support multiple comment types ).

So, it would cover more use cases and be less likely to fail. After all, that's why `'all'` is an option for the `type` argument, so it can be used in situations like this.

dinhtungdu: @shadyvb According to the source of `WP_Comment_Query`, `''` and `'all'` have the same effect, so I don't think we need to specify `all` in this method. See: https://github.com/WordPress/wordpress-develop/blob/master/src/wp-includes/class-wp-comment-query.php#L742-L745

Can you explain to me why other plugins can affect this method, we're querying WordPress comments only?

shadyvb: What I was trying to say is that `'all'` should work for ALL cases, but `''` would not, eg: it's the matter of being better, not being right or wrong.

I'd be happy with just adding a filter and letting each consumer adapt to their needs.

```suggestion
	public function query_db( $args ) {
		$all_query_args = apply_filters( 'ep_search_all_comments_args', [
			'count' => true,
		] );
		
		$all_query = new WP_Comment_Query( $all_query_args );

		$defaults = apply_filters( 'ep_search_paginated_comments_args', [
			'number'  => $this->get_bulk_items_per_page(),
			'offset'  => 0,
			'orderby' => 'comment_ID',
			'order'   => 'desc',
		] );
```

@dinhtungdu Does that make more sense ?

includes/classes/Indexable/Comment/QueryIntegration.php
```diff
@@ -0,0 +1,336 @@
...
+		if ( false === $ep_query ) {
+			$query->elasticsearch_success = false;
+			return $results;
+		}
+
+		if ( ! empty( $query->query_vars['count'] ) ) {
+			return count( $ep_query['documents'] );
+		}
+
+		$query->found_comments        = $ep_query['found_documents'];
```
shadyvb, line 146: `$query->max_num_pages` should be calculated here as well.
```suggestion
		$query->found_comments        = $ep_query['found_documents'];
		$query->max_num_pages         = $query->query_vars['number'] <= 0 ? 1 : max( 1, ceil( $query->found_comments / absint( $query->query_vars['number'] ) ) );
```

includes/classes/Indexable/Comment/Comment.php
```diff
@@ -0,0 +1,1053 @@
...
+
+		$args = apply_filters( 'ep_comment_query_db_args', wp_parse_args( $args, $defaults ) );
+
+		$query = new WP_Comment_Query( $args );
+
+		array_walk( $query->comments, [ $this, 'remap_comments' ] );
+
+		return [
+			'objects'       => $query->comments,
+			'total_objects' => absint( $all_query ),
```
shadyvb, line 743: This call should use `WP_Comment_Query->get_comments()` method, otherwise `absint` would always return 1 casting the actual class rather than the integer.

```suggestion
			'total_objects' => absint( $all_query->get_comments() ),
```

includes/classes/Indexable/Comment/Comment.php
dinhtungdu: @shadyvb Thanks for the clarification, I added a new filter to all query as you suggested. I merged some of your suggestions as well. Great work! I really appreciate all the work you have done!

shadyvb:
For the reference, there is a forked branch based on this one here with a few more commits at https://github.com/humanmade/ElasticPress/compare/e0a7c...75ef6 . This is running in production now without noticeable issues, so thought it'd be useful to share.

jeffpaul:
FYI that https://core.trac.wordpress.org/changeset/48990 was merged into core

includes/classes/Indexable/Comment/Comment.php
```diff
@@ -0,0 +1,1082 @@
...
+					],
+				];
+			}
+		} else {
+			$sort[] = [
+				$orderby => [
+					'order' => $order,
+				],
+			];
+		}
```
felipeelia, line 1077: @Rahmon it looks like this entire section could be rewritten with a switch adjusting $orderby and then in the end we would have only one
```
			$sort[] = [
				$orderby => [
					'order' => $order,
				],
			];
```
do you mind rewriting it?

includes/classes/Indexable/Comment/Comment.php
```diff
@@ -0,0 +1,1082 @@
...
+	 */
+	public function query_db( $args ) {
+
+		$defaults = [
+			'post_type' => 'any',
+			'number'    => $this->get_bulk_items_per_page(),
+			'offset'    => 0,
+			'orderby'   => 'comment_ID',
+			'order'     => 'desc',
+		];
```
felipeelia, line 725: @Rahmon @brandwaffle wondering if we shouldn't have a `'status' => 'approve'` here.

felipeelia (changes requested):

docs/indexables.md
```diff
@@ -1,6 +1,6 @@
...
 
-As of 3.0, ElasticPress ships with the following built-in Indexables: Posts, Terms, and Users. The Posts Indexable roughly corresponds to the previous WP_Query integration, and the Users Indexable adds support for WP_User_Query in ElasticPress. The Terms Indexable adds support for [WP_Term_Query](https://developer.wordpress.org/reference/classes/wp_term_query/). Future versions of ElasticPress will include additional WordPress APIs (such as WP_Comment_Query), and you can also create your own custom Indexables by extending the Indexable class.
+As of 3.5, ElasticPress ships with four built-in Indexables: Posts, Terms, and Users. The Posts Indexable roughly corresponds to the previous WP_Query integration, the Users Indexable adds support for WP_User_Query in ElasticPress, the Comments Indexable adds support for [WP_Comment_Query](https://developer.wordpress.org/reference/classes/WP_Comment_Query/), and the Terms Indexable adds support for [WP_Term_Query](https://developer.wordpress.org/reference/classes/wp_term_query/). Future versions of ElasticPress will include additional WordPress APIs, and you can also create your own custom Indexables by extending the Indexable class.
```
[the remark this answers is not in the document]
Rahmon: Updated :+1:

includes/classes/Feature/Comments/Comments.php
```diff
@@ -0,0 +1,110 @@
...
+	}
+
+	/**
+	 * Determine feature reqs status
+	 *
+	 * @since  3.6
+	 * @return FeatureRequirementsStatus
+	 */
+	public function requirements_status() {
+		$status = new FeatureRequirementsStatus( 0 );
```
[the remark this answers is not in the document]
Rahmon: Updated to not auto-enable by default.

includes/classes/Indexable/Comment/Comment.php
Rahmon: Good catch. Thank you :pray:

includes/classes/Indexable/Comment/Comment.php
Rahmon: Updated

includes/mappings/comment/7-0.php
```diff
@@ -0,0 +1,263 @@
...
+				],
+			],
+			'comment_date'           => [
+				'type'   => 'date',
+				'format' => 'YYYY-MM-dd HH:mm:ss',
+			],
+			'comment_date_gmt'       => [
+				'type'   => 'date',
+				'format' => 'YYYY-MM-dd HH:mm:ss',
+			],
```
shadyvb, line 200:
```suggestion
			'comment_date'           => [
				'type'   => 'date',
				'format' => 'yyyy-MM-dd HH:mm:ss',
			],
			'comment_date_gmt'       => [
				'type'   => 'date',
				'format' => 'yyyy-MM-dd HH:mm:ss',
			],
```

includes/mappings/comment/initial.php
```diff
@@ -0,0 +1,266 @@
...
+					],
+				],
+				'comment_date'           => [
+					'type'   => 'date',
+					'format' => 'YYYY-MM-dd HH:mm:ss',
+				],
+				'comment_date_gmt'       => [
+					'type'   => 'date',
+					'format' => 'YYYY-MM-dd HH:mm:ss',
+				],
```
shadyvb, line 202:
```suggestion
				'comment_date'           => [
					'type'   => 'date',
					'format' => 'yyyy-MM-dd HH:mm:ss',
				],
				'comment_date_gmt'       => [
					'type'   => 'date',
					'format' => 'yyyy-MM-dd HH:mm:ss',
				],
```

shadyvb (changes requested):
The date format mapping is not correct, which causes the date sorting to malfunction. It should be `yyyy-MM-dd HH:mm:ss` rather than with capital Ys: `YYYY-MM-dd HH:mm:ss`.

felipeelia:
@Rahmon can you please give a look at @shadyvb 's comment? For the Posts indexable, it seems that for 5.0 we do use YYYY but for 7 we use yyyy:
https://github.com/10up/ElasticPress/blob/develop/includes/mappings/post/5-0.php#L270
https://github.com/10up/ElasticPress/blob/develop/includes/mappings/post/7-0.php#L293

Also, looking at mappings, _Posts_ has a 5.2 version that is not present for Comments. Isn't it necessary here? Finally, can we update the since tags in the mapping files?

Thanks!

Rahmon:
@felipeelia I've updated the PR.

Regarding a mapping for 5.2, I think it's not necessary since other Indexables doesn't have either. Looked through the Elasticsearch documentation, I didn't find a reason to create a separate mapping for this version. Maybe I'm missing something :thinking:

felipeelia:
@Rahmon , while testing this, I noticed a couple of things:

With WooCommerce and Protected Content enabled:
1. Although it won't index WC order notes during the bulk index, the order note will get indexed if you create a new one directly into a WC order. We'll need to tweak `ElasticPress\Indexable\Comment\SyncManager::action_sync_on_update` to avoid that.
2. I had some unapproved comments stored, one of them was `I want to learn how to make chinese eggrolls` (I really don't know from where it came). Going to /wp-admin/edit-comments.php?s=chinese, I can see 2 queries being sent to ES, none of them with results BUT I can see the comment is listed. So we'll need to check why we have 2 queries and why it is querying MySQL anyway.

Rahmon:
> @Rahmon , while testing this, I noticed a couple of things:
> 
> With WooCommerce and Protected Content enabled:
> 
> 1. Although it won't index WC order notes during the bulk index, the order note will get indexed if you create a new one directly into a WC order. We'll need to tweak `ElasticPress\Indexable\Comment\SyncManager::action_sync_on_update` to avoid that.
> 2. I had some unapproved comments stored, one of them was `I want to learn how to make chinese eggrolls` (I really don't know from where it came). Going to /wp-admin/edit-comments.php?s=chinese, I can see 2 queries being sent to ES, none of them with results BUT I can see the comment is listed. So we'll need to check why we have 2 queries and why it is querying MySQL anyway.

@felipeelia thanks for the review :pray: 

(1) I've fixed this issue and added some tests.

(2) I've updated to better integrate with Protected Content enabled. However, I noticed that the Comments page does a lot of queries to retrieve the comments.

felipeelia:
@Rahmon assigning back to you to sort the last issue with WC orders and client notes as we've discussed. Thanks!

felipeelia (approved):

PR #1540, opened 2019-11-02
Update 7-0.php

lgladdy:
### Description of the Change

In ES 7 (which I know isn't officially supported), objects fail to index if a meta field can't be parsed as a date by default.

Adding this setting makes it behave like <ES7, ignoring the failed field but indexing the rest of the fields.

### Alternate Designs

You can set the ignore_malformed option per field, but it seems sensible to just bring back ES6 behaviours.

### Benefits

ES7 support.

### Possible Drawbacks

I haven't tested this on <ES6, but I assume it will just skip over any settings it doesn't understand. If not, then this will require a much more significant overhaul to support ES 7 - but I figured submitting a PR rather than an issue was a better way forward :)

### Verification Process

Tested on a production ES7 site.


tlovett1:
Thank you! We will get this in the next version.

tlovett1:
What are the steps for reproducing this bug? What version of ES are you using?

tlovett1:
Nevermind, I see. Thanks!

Issue #58, opened 2014-08-01
Track/report usage analytics


AaronHolbrook:
Record how visitors interact with ES: What they're searching for, what was returned, perhaps what they clicked on as a result.

Submitted by @tddewey. 

Great idea - would love to see this reported in a cool dashboard widget even.

PR #1573, opened 2019-11-26
Chore/update autosuggest analytics

mlaroy:
### Description of the Change

Related to #58 - fires off a GA event when an autosuggest item is clicked, if GA is available.

### Alternate Designs

This is an update from an older branch, that had this new code in a JS file that no longer exists within the Features directory.

### Benefits

Allows for tracking with GA when autosuggest results are clicked.

### Verification Process

Tested in the browser with, and without GA on the page to see if the callback/fallback function (page redirect) works as expected.

### Checklist:

### Applicable Issues
#58 

### Changelog Entry

* Add GA tracking event to autosuggest results click


tlovett1:
This feels like we are crossing into pretty opinionated territory. What if a user has GA but doesn't want the event sent? What if they want to format the event?

I'd recommend just providing a JS hook there for people to add in whatever they.

mlaroy:
@tlovett1 I made this change, adding a custom event hook, so that if you want to use this for GA, you could do:
```
window.addEventListener( 'ep-autosuggest-click', e => {
  gtag( 'event', action, {
      'event_category' : 'EP :: Autosuggest',
      'event_label' : e.detail.url,
      'transport_type' : 'beacon',
      'event_callback': function(){
         // whatever you want to do here
         console.log(e.detail.searchTerm);
      }
} );
});
```

mlaroy:
Update: we maintain the custom event hook, but add the gtag call if the feature is enabled in settings: 

Screen Shot 2019-12-12 at 9 58 56 AM

PR #1575, opened 2019-11-27
Fix/autosuggest facet agg conflict

mphillips:
### Description of the Change

Fix a conflict between Autosuggest and Facets where if the autosuggest WP_Query is executed after the main search query, the aggs returned from the search query (stored in $GLOBALS['ep_facet_aggs']) will be overridden and the checkboxes in the Facets widget will not function. This issue does not exist in version 3.2.1, but I haven't tracked down exactly why.

This fix works, but I think there needs to be a better way to resolve this. For example, only set the $GLOBALS['ep_facet_aggs'] variable based on the results of the main query, but I don't fully understand the ramifications of that.

To duplicate this issue:

1. Activate version 3.2.4 of the plugin and enable autosuggest (make sure an autosuggest endpoint is specified if not using EP.io).
2. Set up a search page with a EP Facet widget on it.
3. Search for content that has terms assigned to it that you can filter by. Make sure you see two Elasticsearch requests using the Debug Bar ElasticPress plugin, or some other method. One request should be your search query, the other should have a EP-Search-Term = ep_autosuggest_placeholder header.
4. Try to click on one of the facets in the widget. You should not be able to. 

### Alternate Designs

N/A

### Benefits

This fixes a bug.

### Possible Drawbacks

None that I'm aware of.

### Verification Process

I verified this change locally by first testing a search using version 3.2.4 using the instructions above. I then tested the same search after the changes in this pull request were applied.

### Checklist:

- [ where is this? ] I have read the [**CONTRIBUTING**](/CONTRIBUTING.md) document.

### Changelog Entry

Fix a conflict between Autosuggest and Facets where if the autosuggest WP_Query is executed after the main search query, the aggs returned from the search query (stored in $GLOBALS['ep_facet_aggs']) will be overridden and the checkboxes in the Facets widget will not function.


tlovett1:
Thanks @mphillips. This was fixed a bit differently in another PR.

PR #1585, opened 2019-12-05
error_fix(SearchOrdering.php) - Invalid Argument

Retro64XYZ:
{"message":"Invalid argument supplied for foreach()","context":{"exception":{"class":"ErrorException","message":"Invalid argument supplied for foreach()","code":0,"file":"/var/www/html/pdinfo/releases/20190829162419/web/app/plugins/elasticpress/includes/classes/Feature/SearchOrdering/SearchOrdering.php:743"}

It is possible for the pointers variable in function handle_post_trash to be empty. This is generally not an issue unless running a very strict policy of not allowing any Errors. Our use case requires that we verify that variables being manipulated are verified as not empty.

Please consider accepting this change.

### Description of the Change

Elasticpress can attempt to manipulate an empty variable that it expected to be an array. This change will prevent this manipulation from happening and prevent an exception.

### Alternate Designs

No alternative was considered because this appears to be the easiest and clearest way of managing this possibility. If the array is empty, exit the function.

### Benefits

 This increases resiliency and helps prevent the system from logging an exception. We follow a strict policy when it comes to exceptions so without this fix our application cannot run.

### Possible Drawbacks

No known drawbacks. 

### Verification Process

What process did you follow to verify that your change has the desired effects?

- How did you verify that all new functionality works as expected?
    I added this change and ran it locally  in our live production environment. It resolved our issues and has demonstrated no odd behavior or produced any problems for us.
- How did you verify that all changed functionality works as expected?
    I used the application and verified I could add and subtract custom search results with no issues, errors, or additional shouts in the logs.
- How did you verify that the change has not introduced any regressions?
    This simple change only checks for the existence of a value and has no introduced regressions.

Describe the actions you performed (e.g., commands you ran, text you typed, buttons you clicked) and describe the results you observed.

### Checklist:

### DIFF
--- a/includes/classes/Feature/Documents/Documents.php
+++ b/includes/classes/Feature/Documents/Documents.php
@@ -195,7 +195,7 @@ class Documents extends Feature {
 				 * @param  {string} $id Pipeline ID
 				 * @return  {string} new ID
 				 */
-				$path  = trailingslashit( $index ) . 'post/' . $post['ID'] . '?pipeline=' . apply_filters( 'ep_documents_pipeline_id', Indexables::factory()->get( 'post' )->get_index_name() . '-attachment' );
+				$path = trailingslashit( $index ) . 'post/' . $post['ID'] . '?pipeline=' . apply_filters( 'ep_documents_pipeline_id', Indexables::factory()->get( 'post' )->get_index_name() . '-attachment' );
 			}
 		}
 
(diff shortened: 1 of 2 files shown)


Retro64XYZ:
The contributing link just takes me to a list of jobs. So I clicked yes, but I don't see a contributing documentation section.

tlovett1:
Thank you!

Issue #1553, opened 2019-11-13
missing token ‘ep-search-term’ in CORS header


jt-alex:
Hi, Sorry a bit of a newbie here. I'm getting this CORS error when trying to use the autosuggest feature of elasticpress

Cross-Origin Request Blocked: The Same Origin Policy disallows reading the remote resource at <EP> (Reason: missing token ‘ep-search-term’ in CORS header ‘Access-Control-Allow-Headers’ from CORS preflight channel).

Is this something I need to configure on my site or in the AWS Elasticsearch?

Hope you can help me with this. 

Thanks!

tott:
This is usually caused by an underlying issue with the request itself. Can you look in the web inspector for the request/response that is sent to the ElasticSearch endpoint when you use autosuggest. If you look at the response you get from the server you might be able to see a more detailed message. You can also install the debug-bar-elasticpress plugin to get some more insights into the queries/responses which can help debug the underlying issue. 
It would be also helpful to ensure you're running the most recent version of ElasticPress and that you're not seeing any javascript errors in your Browser console as you're sending the autosuggest requests.

brandwaffle:
@jt-alex we have identified some JavaScript in the plugin that is causing this header to be emitted incorrectly. We're working on a fix.

PR #1593, opened 2019-12-11
Feature/ajax header epio

mlaroy:
### Description of the Change

In autosuggest, using `Utils\is_epio()` to conditionally add to the `epas` object. The JavaScript then checks for this, and only adds the header if this is present.

### Benefits

As detailed in https://github.com/10up/ElasticPress/issues/1553, this will only allow the 'ep-search-term' header if using the ep.io service, preventing errors for other ES solutions.

### Possible Drawbacks

What if someone with a custom ES host has modified their server to require this header?

### Verification Process

Used the console to determine if the ajaxHeader key gets added to the epas, and the Network tab to determine if the headers object is succesfully added to the ajax config used with the jQuery ajax method,

### Checklist:

### Applicable Issues
fix/SIT-654

#1553


tlovett1:
What's the logic behind only sending this header for EP.io?

mlaroy:
@tlovett1 this was discussed in our meeting on Tuesday, this will prevent CORS issues for plugin users using different ES services where this header is the culprit.

Issue #1579, opened 2019-12-03
Autosuggest triggering in all input[type="search"]


samikeijonen:
**Is your enhancement related to a problem? Please describe.**
I think autosuggest triggering in all `input[type="search"]`. In site there can be other search fields which are not related to ElasticSearch.

**Describe the solution you'd like**
Being able to define which search fields will have autosuggest. This can be class value.

Or set a class which disable autosuggest.

MediaMaquina:
Hi, this option already exist on dashboard -> Elasticpress (unless i didn't understand you?)
![settings](https://user-images.githubusercontent.com/8179611/70236319-01f21700-1765-11ea-9daa-55592d50dddb.png)

samikeijonen:
@MediaMaquina Is this for **additional** fields, not actually replacing the default fields? My hunch is yes but I don't have access to that page in my use case (client site).

mlaroy:
@samikeijonen that's right, this field doesn't replace default fields, but only adds to them. Default fields currently include `.ep-autosuggest`, `input[type="search"]`,  and `.search-field`,

felipeelia:
FWIW, in #2181 (released in EP 3.6.0) this is possible via code using the `ep_autosuggest_default_selectors` filter.

Issue #1394, opened 2019-07-08
Autosuggest: Facet Widget choices


fabianmarz:
__Have you searched for similar issues before submitting this one?__
- ✅ 

__Is this a bug, question or feature request?__
- bug

__Describe the issue you encountered:__
- In https://github.com/10up/ElasticPress/blob/develop/assets/js/autosuggest.js#L248 `input[type="search"]` is used to trigger the autosuggest search query
- Such query is also triggered when using a Facet widget to drill down the facet choices as it's using `input[type="search"]` as well which also inserts an autosuggest list. In my case the list is empty (but didn't figure out why. I think that might be another topic)
- Probably the selector list should be tied to a more specific selector to avoid unnecessary ES queries

__Current WordPress version?__
- 4.9

__Current ElasticPress version?__
- head:develop

__Current Elasticsearch version?__

- 6.7

__Where do you host your Elasticsearch server?__
- Locally

__What ElasticPress features do you have currently active?__
- Autosuggest
- Facets
- Protected content

__Other plugins installed (WooCommerce, Simple Redirect Manager, etc..)?__
- WooCommerce

__Steps to reproduce:__

1. Insert facet widget into page
2. Check XHR when filtering for facet options
3. An `ep-autosuggest` list is inserted as well

__Screenshots, if needed:__
![image](https://user-images.githubusercontent.com/863823/60818590-3b1a6100-a19e-11e9-93ea-d578325292ea.png)

oscarssanchez:
It feels like this is still a problem per https://github.com/10up/ElasticPress/blob/develop/assets/js/autosuggest.js#L359

oscarssanchez:
Weirdly I could not reproduce this, although it would seem as if the problem is still valid. 

Tested with twentynineteen and twentytwentyone. @felipeelia or @Rahmon can you confirm or deny you can reproduce the bug?

felipeelia:
I can't reproduce it. Actually, although we really get all `input[type="search"]` for autosuggest, we do remove the facet ones later since [EP 3.4.3](https://github.com/10up/ElasticPress/releases/tag/3.4.3) with:

```
const epInputs = Array.from(epInputNodes).filter(
	(node) => !node.classList.contains('facet-search'),
);
```

PR #1597, opened 2019-12-12
Feature/disable autosuggest selector

mlaroy:
### Description of the Change

To solve for https://github.com/10up/ElasticPress/issues/1579, this PR adds a field to enter selectors to ignore when activating autosuggest. It adds to the epas object for the JS to consume and filter out.

Screen Shot 2019-12-12 at 9 43 08 AM

Screen Shot 2019-12-12 at 9 46 14 AM

### Benefits

This will allow users to select a field to be NOT an autosuggest field, as the selector `input[type="search"]` is quite generic.

This PR also adjusts how jQuery wraps the search fields, so that having multiple search fields on a page will work properly.

### Verification Process

Tested with multiple search fields, some allowed and some ignored to ensure all were working as intended

### Checklist:

### Added
- Add autosuggest selector field to disable autosuggest for the given selector(s)


fabianmarz:
IMO the feature should be the other way round as it's way too complicated to list all search inputs. Furthermore this isn't really helpful as each search field uses the same query configuration as it's triggered via AJAX. Additionally excluding newly added search fields needs to be kept in mind when adding new plugins whose add new functionality.

To me it makes lot more sense to simply adjust the line as follows:

```diff
- const $epInput = jQuery( `.ep-autosuggest, input[type="search"], .search-field, ${  epas.selector}`  );
+ const $epInput = jQuery( `${  epas.selector}`  );
```

If you want to support all search fields however you can still use `.search-field,  input[type="search"]` as the backend input value. 🤷‍♂

brandwaffle:
@fabianmarz I hear what you're saying but we don't want to 1) break backwards compatibility and 2) the goal of the plugin is to enable a site owner to add search to their theme without additional work when they activate the Search feature.

Are you thinking we would pre-fill the value of the backend field to the current default selector, which would then accomplish the goal of handling the issues I listed above but still enable someone to easily change those values (rather than simply being able to disable some of them)?

fabianmarz:
@brandwaffle, yes, that's valid reasons for sure! 🙂 I think pre-fill the values as suggested by you is a good idea but may still break the backwards compatibility as the settings would have to be saved once for existing installations. To tackle this we could move the default selector set to the `localize_script`/`$epas_options section and don't make them configurable via the backend to have a "fix" set as before. As there is the `ep_autosuggest_options` filter already, devs would still be able to adjust them as the default set might interfere with 3rd party WooCommerce plugins which introduce "search" fields but provide different functionality.

Maybe something like the following:
```php
'selector' => '.ep-autosuggest, input[type="search"], .search-field,' . empty( $settings['autosuggest_selector'] ) ? 'ep-autosuggest' : esc_html( $settings['autosuggest_selector'] ),

// or maybe append them afterwards?
'selector' => empty( $settings['autosuggest_selector'] ) ? 'ep-autosuggest' : esc_html( $settings['autosuggest_selector'] ),
// … 

$epas_options['selector'] .= ', .ep-autosuggest, input[type="search"], .search-field';
```


One concern: 
> the goal of the plugin is to enable a site owner to add search to their theme without additional work when they activate the Search feature.

That's the point with Wordpress. It's popular for its easiness, huge ecosystem and variety of plugins. Most of those plugins (and themes) come with tons of configuration settings which make people doesn't require any technical skill or HTML knowledge to build pages. If such persons install plugins wich introduce fields which use fields which have one of the default selectors, they might break some functionality without noticing. That's why I think the disable approach might not be the best solution.

What do you think?

fabianmarz:
I found the issue which I had with the default selector set: https://github.com/10up/ElasticPress/issues/1394. The facet widget adds another "search" field and broke the functionality. Not sure if this is still relevant. Will have a look and let you know.

JakePT:
#2181 partially addressed this by adding a filter on the default selectors, so they could be excluded by a developer as needed.

PR #1602, opened 2019-12-14
Check for AJAX calls.

mmcachran:
Fix issue where AJAX calls are not taking weighting into account.


tlovett1:
@mmcachran can you explain the use case? Are you trying to use weighting with Protected Content?

mmcachran:
Hi, @tlovett1 - This came out of an issue I was seeing on EAB. Advanced Search is using AJAX to load in more results and since weighting was not being applied it threw off the count. I've added a workaround but it would be nice to have this fixed in the plugin.

https://gitlab.10up.com/eab/eab.com/blob/feature/advance-search-fixes/mu-plugins/eab/includes/EAB/Search/ElasticPress.php#L329

felipeelia:
Closing this one out as it now can be handled with the `ep_is_integrated_request` filter introduced by #2267.

PR #1617, opened 2020-01-08
Support all taxonomies in root WP_Query

### LINKED ISSUE #1576: Custom Taxonomies not included in ElasticPress query
Hi,

Custom Taxonomies are included in index (if that is what you set in Settings), but are not included in queries.

The problem is that you use $query_vars variable (stripped version of full WPQuery object, which is just fine) to build EP query, but in case of custom taxonomies, $query_vars does not include tax_query array, but instead you have "tax_slug" array in $query_vars and WPQuery->tax_query with WPTax object.

To be more specific, you have some sort of fix for empty tax_query array here https://github.com/10up/ElasticPress/blob/develop/includes/classes/Indexable/Post/Post.php#L702

The problem is that you are only handling core taxonomies, category and post_tag, and not custom taxonomies.

The result is a rebuilded tax_query array that does not include custom taxonomies, and you get all the posts returned.

I have reproduces this issue on several diferent environments with the same results.

I made a quick fix using filter ep_formatted_args, but you should rotate through all public taxonomies and use that for rebulding tax_query array (like you've done in the indexing phase).

WP 5.3
EP 3.2.6

tlovett1:
### Description of the Change

Right now we don't do a tax_query for taxonomies called in the root of `WP_Query` e.g.
```php
new WP_Query( [
  'tax_name' => 'slug',
] );
```

Should fix #1576.

### DIFF
--- a/includes/classes/Indexable/Post/Post.php
+++ b/includes/classes/Indexable/Post/Post.php
@@ -697,45 +697,63 @@ class Post extends Indexable {
 		 * @since 0.9.1
 		 */
 
-		// set tax_query if it's implicitly set in the query.
-		// e.g. $args['tag'], $args['category_name'].
-		if ( empty( $args['tax_query'] ) ) {
-			switch ( $args ) {
-				case ! empty( $args['category_name'] ):
-					$args['tax_query'][] = array(
-						'taxonomy' => 'category',
-						'terms'    => array( $args['category_name'] ),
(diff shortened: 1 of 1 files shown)


wss-chadical:
:+1:

petenelson:
@tlovett1 I can tackle some unit tests for this if you haven't written any yet.

PR #1619, opened 2020-01-10
Fixed PHP autosuggest placeholder is not being used.

fabianmarz:
### Description of the Change

- While the autosuggest search string placeholder value can be changed via a [PHP hook](https://github.com/10up/ElasticPress/blob/develop/includes/classes/Feature/Autosuggest/Autosuggest.php#L408), it's not picked up by the autosuggest JS and therefore the query will break or doesn't deliver the correct results.

### Alternate Designs

- Instead of deleting `const placeholder = 'ep_autosuggest_placeholder';` I considered using `const placeholder = queryJSON.placeholder;` but decided to drop the line completely as the placeholder is used in `buildSearchQuery()` only.

### Benefits

- Don't break any the results if other than the default value is used :)

### Possible Drawbacks
- X

### Verification Process
- Inspected the query after the search and replace was done to ensure the query runs with the actual search term.

### Changelog Entry

> Fixed - Autosuggest doesn't work if the placeholder is changed via a `ep_autosuggest_query_placeholder` filter.
 

### Credits

Props @fabianmarz @burhandodhy 

### Checklist:

### Applicable Issues
- Didn't find any.


tlovett1:
@mustafauysal can you review this one?

mustafauysal:
@tlovett1 looks good to me 👍

felipeelia:
This PR will need to be refreshed and have tests added. Setting it to EP 5.3.0.

PR #1621, opened 2020-01-10
Fixes for WP_User_Query 'fields' parameter

### LINKED ISSUE #1609: WP_User_Query returns objects when 'fields' => 'ID' is queried
**Describe the bug**

Adding `ep_integrate` to a user query with `'fields' => 'ID'` returns an array of objects, which is different from the default behavior for returning just the user IDs.

Default behavior:

```php
$user_ids_search = new \WP_User_Query([
    'ep_integrate' => false,
    'fields' => 'ID',
    'search' => 'admin',
]);

$results = $user_ids_search->get_results();

// $results:
//
// array (size=1)
//  0 => string '1'
```

EP behavior:

```php
$user_ids_search = new \WP_User_Query([
    'ep_integrate' => true,
    'fields' => 'ID',
    'search' => 'admin',
]);

$results = $user_ids_search->get_results();

// $results:
//
// array (size=1)
//   0 => 
//    object(stdClass)[2165]
//      public 'ID' => int 1
//      public 'elasticsearch' => boolean true
```

**Steps to Reproduce**
1. Create a `WP_User_Query` with the `'fields' => 'ID'` and `'ep_integrate' => true` parameters.
2. Observe that the results of the query contain an array of `stdClass` objects.

**Expected behavior**
The results of the query contain an array of user IDs, without a wrapper object.

**Environment information**
 - Device: PC
 - OS: Windows 10 Pro
 - Browser and version: N/A
 - Plugins and version: ElasticPress 3.3 / ElasticSearch 6.3.2
 - Theme and version: Custom theme
 - Other installed plugin(s) and version(s): N/A

**Additional context**

The difference isn't hard to work around, but it makes migrating existing user queries tricky. This is especially true for queries in which `'ep_integrate' => true` is enabled automatically, and can break existing code.

The problem seems to come from [`format_hits_as_users` in `User\QueryInteraction`](https://github.com/10up/ElasticPress/blob/develop/includes/classes/Indexable/User/QueryIntegration.php#L114), which wraps the results in objects.

[`Post\QueryIntegration` has some separate handlers based on the field type](https://github.com/10up/ElasticPress/blob/develop/includes/classes/Indexable/Post/QueryIntegration.php#L297), which might be a good alternative to handle this problem for users.

petenelson:
### Description of the Change
- Added `testSanitizeCredentials()` to help me get up to speed on EP unit tests and additional examples to the development documentation.
- Updated `maybe_filter_query()` for WP_User_Query to handle a single field and multiple fields as an array.
- Added unit tests `testSingleUserFieldQuery()` and `testMultipleUserFieldsQuery()`

### Benefits
User queries with 'fields' will now work correctly.

### Possible Drawbacks
None

### Verification Process
Created unit tests for the updated functionality.

### Checklist:

### Applicable Issues
Fixes #1609 

### Changelog Entry
- Fixed bug where passing 'fields' to WP_User_Query was not returning the requested fields

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,5 @@ node_modules
 .history
 docs-built
 npm-log.log
+test-coverage-html/
+.phpunit.result.cache
--- a/docs/development.md
+++ b/docs/development.md
@@ -28,6 +28,18 @@ Now run the tests:
 EP_HOST="http://elasticsearch:9200" phpunit
 ```
 
+To run a specific test, pass the `--filter` parameter:
+
+```
+EP_HOST="http://elasticsearch:9200" phpunit --filter=testSanitizeCredentials
+```
+
+To verify all of the code paths are being tested, you can generate an HTML code coverage report for your tests. Pass `--coverage-html test-coverage-html` as a parameter and a report will be generated in that directory.
+
+```
+EP_HOST="http://elasticsearch:9200" phpunit --filter=testSanitizeCredentials --coverage-html test-coverage-html
+```
+
 To run WP Acceptance, navigate to the root of the plugin and run:
 
 ```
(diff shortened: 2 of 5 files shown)


tlovett1:
Awesome! Thank you.

PR #1624, opened 2020-01-14
Moved EP indexable posts class registration into a function that runs in the plugins_loaded action

### LINKED ISSUE #1613: http_request_args filter is not run in settings screen
**Describe the bug**
The `http_request_args` filter is not run when EP makes a request to elasticsearch in the WP admin, specifically on the settings page of the plugin.

My scenario is that I'm adding headers to requests to my elasticsearch server running in AWS to authorize the requests. 

If I index via the cli, everything works properly and the settings screen works as expected for a time. However, once some caches go away (I assume), the plugin can no longer communicate with elasticsearch via the settings screen. Note that searches on site work properly (i.e. external http requests are processed via the `http_request_args` filter).

**Steps to Reproduce**
A minimal plugin which reproduces the issue:
```php
<?php
/**
 * Plugin Name: Test EP with http_request_args filter
 */
add_filter('http_request_args', function (array $args, string $url) {
    error_log('http_request_args processing request for url: ' . $url);
    return $args;
}, 1, 2);
```

1. install and activate the test plugin
1. intentionally create a problem communicating with elasticsearch (in my test I removed my host's IP from allowed IPs in the AWS access policy).
1. ensure wordpress is in debug mode and tail the error log (or otherwise make it so that you will see the error logs)
1. in EP's settings screen, click the "try again" link

**Expected behavior**
The http request to elasticsearch should be processed by the `http_request_args` filter and log line should be generated: `http_request_args processing request for url: <my elasticsearch url>`. 

Note that other installed plugins making external requests are processed.

**Environment information**
- WordPress 5.2.5
- Ubuntu 16.04.5 LTS

**Additional context**
I checked EP source and it calls `wp_remote_request` under the hood, so the filter should be called. I can only assume it's not be registered when the external request call happens, but I'm no expert in the WP admin page request life-cycle.

petenelson:
### Description of the Change
Moves the `Indexables::factory()->register( new Indexable\Post\Post() );` call into a `plugins_loaded` action.

### Alternate Designs
None.

### Benefits
The indexable post class make a call to Elasticsearch inside its constructor. Since this constructor is being called prior to any other plugins being loaded, any hooks/filters in the other plugins will not run.

### Possible Drawbacks
Possible breaking changes for existing plugins/themes that interact with EP, but the scope of this is unknown.

### Verification Process
Ran all of the existing unit tests locally, all passed.
```
root@350c654785b0:/var/www/html/wp-content/plugins/elasticpress# EP_HOST="http://elasticsearch:9200" ./vendor/bin/phpunit
Installing...
Installing network...
Running as multisite...
WordPress version 5.3.2
Elasticsearch version 7.5.1
Installing WooCommerce version 3.6.5 ...
Not running ajax tests. To execute these, use --group ajax.
Not running ms-files tests. To execute these, use --group ms-files.
Not running external-http tests. To execute these, use --group external-http.
PHPUnit 7.5.12 by Sebastian Bergmann and contributors.

...............................................................  63 / 242 ( 26%)
............................................................... 126 / 242 ( 52%)
............................................................... 189 / 242 ( 78%)
.....................................................           242 / 242 (100%)

Time: 4.59 minutes, Memory: 86.25 MB

OK (242 tests, 797 assertions)
```

### Checklist:

### Applicable Issues
Fixes #1613 

### Changelog Entry
- Moved EP indexable posts class registration into a function that runs in the `plugins_loaded` action

### DIFF
--- a/elasticpress.php
+++ b/elasticpress.php
@@ -83,60 +83,61 @@ if ( $network_activated ) {
 }
 
 /**
- * Sets up all the various EP classes.
+ * Sets up the indexable posts class.
  *
  * @return void
  */
-function ep_setup() {
-	global $wp_version;
-
+function register_indexable_posts() {
 	/**
 	 * Handle indexables
 	 */
 	Indexables::factory()->register( new Indexable\Post\Post() );
 
-	/**
-	 * Handle features
-	 */
-	Features::factory()->register_feature(
-		new Feature\Search\Search()
-	);
+}
+add_action( 'plugins_loaded', __NAMESPACE__ . '\register_indexable_posts' );
 
-	Features::factory()->register_feature(
-		new Feature\ProtectedContent\ProtectedContent()
-	);
+global $wp_version;
 
-	Features::factory()->register_feature(
-		new Feature\Autosuggest\Autosuggest()
-	);
+/**
+ * Handle features
+ */
+Features::factory()->register_feature(
+	new Feature\Search\Search()
+);
 
-	Features::factory()->register_feature(
-		new Feature\RelatedPosts\RelatedPosts()
-	);
+Features::factory()->register_feature(
+	new Feature\ProtectedContent\ProtectedContent()
+);
 
-	Features::factory()->register_feature(
-		new Feature\WooCommerce\WooCommerce()
-	);
+Features::factory()->register_feature(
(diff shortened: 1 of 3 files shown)


twhid:
@tlovett1 @petenelson thanks!

PR #1627, opened 2020-01-16
Use post_author.display_name for weighted author field search

### LINKED ISSUE #1622: Weighting Engine uses wrong name for author field
**Describe the bug**

The weighting engine uses the field `author_name` to map to the weighting of the Author. Unfortunately it's not being translated into an ES field. For example, in the Posts indexable we map the author_name to the `post_author.login.raw` field https://github.com/10up/ElasticPress/blob/941b5c69e1b0622babe3c54852e03a1a91b70d33/includes/classes/Indexable/Post/Post.php#L831-L835

The weighting engine just tries to query for a weighted field named `author_name` which doesn't exist in the index and therefore doesn't match any results.

**Steps to Reproduce**

1. Check the `Author` box in the Weighting Engine.
2. Do a search for an author's name or login. No author-based results appear.

**Expected behavior**

Results for Authors matching the search term should appear.

**Screenshots**

**Environment information**
Just core + EP.

**Additional context**

I don't think we should simply copy the code from the Posts indexable. I don't think people on a typical site where authors are important are looking for `jupdike` so much as  `John Updike`. I would prefer we use `post_author.display_name` for the weighting/search field.

petenelson:
### Description of the Change
Updates the `author_name` field in weighted search to use `post_author.display_name` to better support more natural searches (ex: John Doe vs john-doe).

### Alternate Designs
None

### Benefits
Frontend searches for posts by author name will include more relevant results.

### Possible Drawbacks
Existing plugin or theme code that is expecting the author_name to use the login slug may no longer work.

### Verification Process
Updated `testAuthorNameQuery()` to support searches by display name. Ran all unit tests, all but one passed but it is related to sticky posts.

### Checklist:
- [X] My change requires a change to the documentation.
- [X] I have updated the documentation accordingly.

### Applicable Issues
Fixes #1622 

### Changelog Entry
- Updated `author_name` in WP_Query integration to search the `post_author.display_name`


includes/classes/Indexable/Post/Post.php
```diff
@@ -811,6 +811,12 @@ public function format_args( $args, $wp_query ) {
 			$use_filters = true;
 		}
 
+		// Use the original author name passed into the query. query_vars
+		// contains a slugified author_name value.
+		if ( isset( $wp_query->query['author_name'] ) ) {
+			$args['author_name'] = sanitize_text_field( $wp_query->query['author_name'] );
```
tlovett1: @petenelson I don't follow this part. Mind explaining?

petenelson: Can you expand on what you're not following?

includes/classes/Feature/Search/Weighting.php
```diff
@@ -396,6 +396,13 @@ public function recursively_inject_weights_to_fields( &$fieldset, $weights ) {
 					if ( 0 !== $weight ) {
 						$fieldset['fields'][ $key ] = "{$field}^{$weight}";
 					}
+
+					// Use display_name as the searchable field. author_name was
+					// used in previous versions of the plugin.
+					if ( 'author_name' === $field ) {
```
tlovett1: Shouldn't this be within the weight check on line 396?

tlovett1: BTW I'm seeing this isn't working in the production plugin. author_name weighting is broken.

PR #1641, opened 2020-01-22
Terms indexable updates

petenelson:
* Updated term mapping for ES 7.0
** Removed `standard` as a filter https://www.elastic.co/guide/en/elasticsearch/reference/current/breaking-changes-7.0.html#standard-filter-removed
** Added `index.max_shingle_diff`
** Removed `term` as one of the levels in the mapping array to match the levels of the mapping array for posts.
* Fixed testTermQueryOrderParent()

### DIFF
--- a/includes/classes/Indexable/Term/QueryIntegration.php
+++ b/includes/classes/Indexable/Term/QueryIntegration.php
@@ -220,7 +220,7 @@ class QueryIntegration {
 				}
 			}
 
-			$term->elasticsearch = true; // Super useful for debugging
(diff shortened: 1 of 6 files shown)


tlovett1:
@petenelson for the mapping, the one @dkotter originally created seems like it's for ES 5-6. You edited to work for 7. I actually think we need to leave his original mapping, rename it accordingly, and create another ES 7 with all your changes.

tlovett1:
Awesome work!

Issue #1496, opened 2019-09-24
Indexing with wp cli doesn't pause requests to EP.


eflorea:
**Describe the bug**
On a multi-site install running elasticpress index for a specific site does not pause requests to EP.

**Steps to Reproduce**

1. Network activate EP plugin
2. Run wp cli: elasticpress index --url=subsite
3. Requests are still being made to EP while the indexing is running.

CC: @christianc1

tlovett1:
Are you saying requests are being made on parts of the site besides `subsite`? That's intended behavior.

tlovett1:
Thinking about this a bit more, here are the scenarios:
1. Active on individual site in network. Indexing without `--network-wide`
2. Network active. Indexing without `--network-wide
3. Network active. Indexing with `--network-wide

For 1, we obviously don't want to pause requests across the network.

For 2, I think we should pause requests across the network. We are currently *not* doing that.

For 3, we are correctly pausing requests across the entire network.

If we are in agreement on 2, then your PR needs to be fixed @asharirfan

eflorea:
Yes, the scenario I'm talking about is 2.

oscarssanchez:
Looks like #1647 resolved this. Closing.

PR #1647, opened 2020-01-29
Rethink usage of network-wide CLI argument

tlovett1:
### Description of the Change

There are a few places we are checking `--network-wide` when we should be checking `EP_IS_NETWORK` (see #1496). This refactors that logic. Network wide is only needed in indexing, put mapping, and deleting index(es).

### Testing

To test, all CLI functions should be tested with and without network wide as well as network activated and not network activated.

### Checklist:

### Applicable Issues

#1496


tlovett1:
@eflorea do you mind testing this?

PR #1649, opened 2020-01-30
Fix doc for get-cluster-indexes and get-indexes

barryceelen:
The [cli commands documentation](http://10up.github.io/ElasticPress/tutorial-wp-cli.html) for `get-cluster-indexes` and `get-indexes` is incorrect, it uses underscores in stead of dashes.


tlovett1:
Thanks!

PR #1655, opened 2020-02-06
Enforce conversion of post_status to plain array if not 'any'

chrean:
### Description of the Change

During our tests of ElasticPress we have found out that, in some cases, `$post_status` does not get translated to a plain array if it is an associative array, and that causes weird errors in requests to Elasticsearch. 
If it's a plain array it will just work as intended, instead.
In other parts of the code this conversion already happens, so for safety and consistency we think we should be adding it wherever it applies (unless we are missing something obvious).

### Alternate Designs

N/A

### Benefits

Should be more solid, not returning errors in case a hash is passed instead of an array. :)

### Possible Drawbacks

Not sure there are any.

### Verification Process

Our code does the following:

```php
$post_statuses = Indexables::factory()->get( 'post' )->get_indexable_post_status();

$query_args = [
	'post_type' => $post_type,
	'post_status' => array_values( $post_statuses ), // Associative array
];

$query = return new WP_User_Query( $query_args );

$formatted_args = $indexable->format_args( $query->query_vars, $query );

// Get exact total count since Elasticsearch default stops at 10,000.
$formatted_args['track_total_hits'] = true;

$es_result = $indexable->query_es( $formatted_args, $query->query_vars );
```

Removing the `array_values()` conversion will cause the query to fail.

### Checklist:

### Applicable Issues

N/A

### Changelog Entry

- Enforce conversion of `$post_status` from associative array to plain array if `args['post_status'] !== 'any'`


tlovett1:
I'd really like to rewrite the logic here since it's overly confusing.

I don't think the variable you are using array_values on is always an array per the code when there is less than 2 terms.

chrean:
Makes sense, thanks for the feedback @tlovett1 .
What's your take here, then? Should we double check or did you have a more consistent refactor in mind (I agree about the logic being confusing at times for that function).
In the meanwhile, we are passing the array through `array_values()` on our side.

brandwaffle:
@chrean @tlovett1 I think WP itself does this in the event it receives post_status as a keyed array. I haven't had a chance to revisit that code recently, but if it does happen in core we might as well use the same approach.

tlovett1:
@chrean can you pull out all the unit testing script additions into a new PR?

tlovett1:
@moraleida if you have time, can you test this one out?

moraleida:
@chrean @tlovett1 I looked at the base branch for this and it's looking very different from what I see in the diff. Before I dive too deep into testing, can you confirm the following, please?

The only affected area of the code base should be this if statement:
https://github.com/Automattic/ElasticPress/blob/develop/includes/classes/Indexable/Post/Post.php#L1224-L1242

And within that, the intended change is this:
https://github.com/moraleida/ElasticPress/commit/785e3d128bf3564d53728d94e3aa44621865c4e2

Is this correct?

PR #1659, opened 2020-02-13
check constant definition first

mustafauysal:
### Description of the Change

@tlovett1 - It seems the changes in 79d49e944043aa5647ad68f261509d1732d3a897  (#1647) causing a regression when the plugin is not network-wide activated. 

(Alternatively, we can set `define( 'EP_IS_NETWORK', $network_activated );` by removing if statement for making sure the constant defined)

### Testing

To test, all CLI functions should be tested with and without network-wide as well as network activated and not network activated.

### Checklist:


tlovett1:
Good catch.

PR #1661, opened 2020-02-14
add hierarchical taxonomy support to root level tax_query

mustafauysal:
### Description of the Change

Changes in #1617 breaking WooCommerce category archives. It can potentially break any hierarchical taxonomy archives (IMO). Appending child slugs seemed the solution. 

@tlovett1 opinions on this?

### Verification Process

* Visit a WooCommerce parent product category.
* Make sure parent category has a child category and there are products belongs to child category only
* See the products on the main category

### Checklist:

### Applicable Issues

#1576 could be related

### DIFF
--- a/elasticpress.php
+++ b/elasticpress.php
@@ -221,6 +221,7 @@ function handle_upgrades() {
 			'3.0',
 			'3.1',
 			'3.3',
+			'3.4',
 		)
 	);
 
--- a/includes/classes/Feature/WooCommerce/WooCommerce.php
+++ b/includes/classes/Feature/WooCommerce/WooCommerce.php
@@ -293,31 +293,10 @@ class WooCommerce extends Feature {
 			if ( ! empty( $term ) ) {
 				$integrate = true;
 
-				$terms          = (array) $term;
-				$children_terms = [];
-
-				// to add child terms to the tax query
-				if ( is_taxonomy_hierarchical( $taxonomy ) ) {
-					foreach ( $terms as $term ) {
-						$term_object = get_term_by( 'slug', $term, $taxonomy );
-						if ( $term_object && property_exists( $term_object, 'term_id' ) ) {
-							$children = get_term_children( $term_object->term_id, $taxonomy );
-							if ( $children ) {
-								foreach ( $children as $child ) {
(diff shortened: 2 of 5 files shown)


tlovett1:
@mustafauysal this isn't how WP works by default so WooCommerce must be doing something else to accomplish that behavior. Let's discuss.

tlovett1:
@mustafauysal I think we need to fix this for all of EP and not just WooCommerce. This might be relevant: https://github.com/10up/ElasticPress/blob/develop/includes/classes/Indexable/Post/Post.php#L399

tlovett1:
I actually think all we need to do is set https://github.com/10up/ElasticPress/blob/develop/includes/classes/Indexable/Post/Post.php#L399 to `true` by default

mustafauysal:
@tlovett1 yep, setting `ep_sync_terms_allow_hierarchy` `true` seems work.  We should require reindexing for version 3.4 upgrade to make this work.

tlovett1:
@brandwaffle thoughts on just doing that?

brandwaffle:
@tlovett1 @mustafauysal I'm fine with that approach.

mustafauysal:
@tlovett1 @brandwaffle I have refreshed the PR.

Well, the first approach seemed more alike to WP_Query behavior and the new changes will increase index size and bandwidth usage. _(I'm not sure we should concern about that since term data is relatively small for most cases. We might want to add some upgrade notice to readme file)_


@tlovett1 Do we need `ElasticPressTest\TestPost::testPostTermSyncSingleLevel` and `ElasticPressTest\TestPost::testPostTermSyncSingleLevelQuery`? If so, are you ok with removing the following assertions to fixing test cases?

https://github.com/10up/ElasticPress/blob/ed49184e69e53337bb60d44b6ce131870100c8e0/tests/php/indexables/TestPost.php#L309-L311

https://github.com/10up/ElasticPress/blob/ed49184e69e53337bb60d44b6ce131870100c8e0/tests/php/indexables/TestPost.php#L458-L460

tlovett1:
Yep, let's remove those assertions.

tlovett1:
Merging this now as this is a big fix. Let's add the unit test fixes in another PR.

Issue #1456, opened 2019-08-23
Re-factor to Vanilla JS by removing jQuery dependency 


samikeijonen:
**Is your enhancement related to a problem? Please describe.**
In most cases projects don’t use jQuery in the FE anymore and jQuery is kind of a blocker.

**Describe the solution you'd like**
Re-write JS like `autosuggest.js` in Vanilla JS.

oscarssanchez:
Closing as frontend seems to be Vanilla JS per https://github.com/10up/ElasticPress/issues/2074

PR #1663, opened 2020-02-14
Refactor JavaScript files to remove jQuery

mlaroy:
### Description of the Change

This PR is a large refactor of the JavaScript for the autosuggest feature, with the primary goal of removing the dependency on jQuery. Addresses issues #1456 and #1394

**Update**: per discussion on April 28, this PR has been expanded to include the removal of jQuery from all JS files, as well as updating to the latest eslint config settings from 10up. This includes:
* autosuggest
* facets
* admin dashboard pages:
   - settings
   - ES sync
   - stats
   - sites admin
   - ordering
   - related posts

### Benefits

Aside from the removal of jQuery, this PR fixes an issue with the up/down keys not keeping their selection when navigating between results items.

This also changed the `span` tags into `a` tags withing the autosuggest dropdown list, even though the JavaScript still handles the page change. This should make for a small accessibility improvement.

Also ads a hook for filtering ajax response data before outputting it to the autosuggest box, as some users are relying on the jQuery `ajaxSetup` dataFilter method to modify the display.

Further, this PR includes the latest version of 10up's `eslint-config` setup (2.0.0) for improved linting and code formatting.

Finally, while this doesn't change the underlying logic, a number of the functions have been simplified or broken out into smaller functions for reuse and commented, which should improve maintainability and code readability.

### Possible Drawbacks

Sites that use autosuggest and leverage jQuery's `ajaxSetup` for example will have to make adjustments. In this way, it could certainly be considered a breaking change.

### Checklist:


assets/js/autosuggest.js
```diff
@@ -1,37 +1,43 @@
...
  * Navigate to the selected item, and provides
  * event hook for JS customizations, like GA
  * @param event
  */
-function goToAutosuggestItem( $localInput, url ) {
-
+function goToAutosuggestItem( input, url ) {
 	const detail = {
-		searchTerm: $localInput[0].value,
+		searchTerm: input.value,
```
Shavindra: only property from input used here is value.

so this can be simplified.

```
function goToAutosuggestItem( searchTerm, url ) {

	const detail = {
                searchTerm,
                url,

File: assets/js/autosuggest.js [hunk 2]
```diff
@@ -1,37 +1,43 @@
...
-function selectAutosuggestItem( $localInput, text ) {
-	$localInput.val( text );
+function selectAutosuggestItem( input, text ) {
+	input.value = text;
 }
 
 /**
  * Navigate to the selected item, and provides
  * event hook for JS customizations, like GA
  * @param event
```
Shavindra: This need to change?

assets/js/autosuggest.js
```diff
@@ -62,40 +68,17 @@ function triggerEvents( detail ) {
...
- * @param $localInput
+ * @param input
  * @param element
  */
-function selectItem( $localInput, element ) {
+function selectItem( input, element ) {
 
 	if ( 'navigate' === epas.action ) {
-		return goToAutosuggestItem( $localInput, element.dataset.url );
+		return goToAutosuggestItem( input, element.dataset.url );
```
Shavindra: same way you're passing just the URL instead of `dataset` can just pass the `input.value`

assets/js/autosuggest.js
```diff
@@ -163,147 +122,135 @@ function buildSearchQuery( searchText, placeholder, { query } ) {
...
+		if( 'undefined' !== typeof window.epDataFilter ) {
+			data = window.epDataFilter( data, searchTerm );
+		}
+
+		return data;
+	} catch( error ) {
+		// eslint-disable-next-line no-console
+		console.error( error );
+		return error;
+	}
```
Shavindra: `fetch` is a promise. you don't need `try/catch`. you can just do `then().catch()`

This can be abstracted out to be a generic utility function e.g. `httpRequest` or `fetchRequest`

in that class you can do something like below. I notice that search term is not a query param but in headers. So you'd have to modify and work that in to the class. May be we have to create a new issue for this I am not sure though. 

```
class fetchRequest {
   constructor(config) {}


    // Handle response
    // Fetch doesn't distinguish Error / Success
    public status(response: Response) {
        if (response.ok) {
            return response.json();
        } else {
           const error = JSON.stringify({
                status: response.status,
                message: response.statusText
            });
           throw new Error(error);
        }
    }

    // Handle only server errors
    public handleError(error: Error) {
        throw error;
    }

    public get(url: string, searchParams?: any) {
        const options = {
            method: 'GET',
        }
        const requestUrl = this.url(url, searchParams)
    
        const request = this.request(requestUrl, options);

        return fetch(request)
        .then(this.status)
        .catch(this.handleError);
    }


}

```

assets/js/autosuggest.js
```diff
@@ -339,123 +286,272 @@ function checkForOrderedPosts( hits, searchTerm ) {
...
+				return { text, url };
 			}
 		} );
-	} );
+	};
+
+
+	// to be used by the handleUpdDown function
+	// to keep track of the currently selected result
+	let $currentIndex;
```
Shavindra: remove `$` ?

assets/js/autosuggest.js
```diff
@@ -339,123 +286,272 @@ function checkForOrderedPosts( hits, searchTerm ) {
...
+	epInputs.forEach( input => {
+		input.addEventListener( 'keyup', handleKeyup );
+		input.addEventListener( 'blur', hideAutosuggestBox );
 	} );
 
+}
+
+// Ensure we have an endpoint URL, or
+// else this shouldn't happen
+if ( epas.endpointUrl && '' !== epas.endpointUrl ) {
```
Shavindra: for readability `epas.endpointUrl && ( '' !== epas.endpointUrl )

assets/js/autosuggest.js
```diff
@@ -339,123 +286,272 @@ function checkForOrderedPosts( hits, searchTerm ) {
...
+			return;
+		}
+
+		if ( 2 <= searchText.length ) {
+			const query = buildSearchQuery( searchText, placeholder, queryJSON );
+
+			// fetch the results
+			const response = await esSearch( query, searchText );
+
+			if ( response && 0 < response._shards.successful ) {
```
Shavindra: for readability `response && ( 0 < response._shards.successful )`

assets/js/autosuggest.js
nicholasio: The try/catch here is needed because async/await is being used. I'm also not sure if it's worth abstracting this into a utility function unless we're using fetch multiple times.

assets/js/autosuggest.js
```diff
@@ -163,147 +122,135 @@ function buildSearchQuery( searchText, placeholder, { query } ) {
...
-			return;
-		}
+	// append list items to the list
+	suggestList.innerHTML = itemString;
 
+	const autosuggestItems = Array.from( document.querySelectorAll( '.autosuggest-item' ) );
+	autosuggestItems.forEach( item => {
+		item.addEventListener( 'click', event => {
+			selectItem( input, event.target );
+		} );
```
Shavindra: Also delegate the event to the parent? instead of individual items?

assets/js/autosuggest.js
```diff
@@ -163,147 +122,135 @@ function buildSearchQuery( searchText, placeholder, { query } ) {
...
 
-/**
- * Escapes double quotes for specific data-attr
- *
- * @param str
- * @returns string The escaped string
- */
-function escapeDoubleQuotes( str ) {
-	return str.replace( /\\([\s\S])|(")/g, '&quot;' );
+		let data = await response.json();
```
Shavindra: no need for `await` here.

assets/js/autosuggest.js
```diff
@@ -163,147 +122,135 @@ function buildSearchQuery( searchText, placeholder, { query } ) {
...
+
+		// allow for filtered data before returning it to
+		// be output on the front end
+		if( 'undefined' !== typeof window.epDataFilter ) {
+			data = window.epDataFilter( data, searchTerm );
```
Shavindra: can just add `return` here so it's clearer instead of assigning back to `data`

assets/js/autosuggest.js
Shavindra: I was thinking if moved the logic of fetch somewhere else here the code is simpler just to return the promise from here. I think using `try` and `catch` make sense with more than 1 request. If it's not going to get used anywhere else you're right it can be left as it is.

assets/js/autosuggest.js
```diff
@@ -163,147 +122,135 @@ function buildSearchQuery( searchText, placeholder, { query } ) {
...
 /**
  * Update the auto suggest box with new options or hide if none
  *
  * @param options
  * @return void
  */
-function updateAutosuggestBox( options, $localInput ) {
+function updateAutosuggestBox( options, input ) {
 	let i,
 		itemString = '';
```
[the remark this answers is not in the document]
mlaroy: this is done by the linter. Perhaps 2 lines instead:
```
let i;
let itemString = '';
```

assets/js/autosuggest.js
mlaroy: I'm not sure this is true - https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API/Using_Fetch uses await here.

assets/js/autosuggest.js
mlaroy: I agree with @nicholasio here - I don't see the need for abstraction at this point. For the moment, this is the only fetch request, and reading it this way is clear IMO.

assets/js/autosuggest.js
```diff
@@ -163,147 +122,135 @@ function buildSearchQuery( searchText, placeholder, { query } ) {
...
+	// get the search term for use later on
+	const { value } = input;
+	const container = findAncestor( input, 'ep-autosuggest-container' );
+	const resultsContainer = container.querySelector( '.ep-autosuggest' );
+	const suggestList = resultsContainer.querySelector( '.autosuggest-list' );
 
-	$localSuggestList.empty();
+	// empty the the list of all child nodes
+	while( suggestList.firstChild )
+		suggestList.removeChild( suggestList.firstChild );
```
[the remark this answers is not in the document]
mlaroy: either way, the while will only return true so long as there are any children, one or more.

assets/js/autosuggest.js
```diff
@@ -339,123 +286,272 @@ function checkForOrderedPosts( hits, searchTerm ) {
...
 	 *
+	 * @param {*} event
 	 */
-	$epInput.each( ( key, localInput ) => {
-		const $localInput = jQuery( localInput );
-		$localInput.on( 'keyup', debounce( ( event ) => {
-			if ( 38 === event.keyCode || 40 === event.keyCode || 13 === event.keyCode || 27 === event.keyCode ) {
-				return;
-			}
+	const handleUpdDown = event => {
```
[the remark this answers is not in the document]
mlaroy: this isn't required with a single argument, unless this is a recommended practice?

assets/js/autosuggest.js
```diff
@@ -339,123 +286,272 @@ function checkForOrderedPosts( hits, searchTerm ) {
...
+	 * ajax request.
+	 *
+	 * @param {*} event
+	 */
+	const handleKeyup = event => {
+		event.preventDefault();
+
+		const input = event.target;
+		const keyCodes = [
+			38, // up
```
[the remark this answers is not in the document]
mlaroy: perhaps - though the two instances don't share the same values. Is it overkill? It's not like we are modifying these values anywhere.

assets/js/autosuggest.js
```diff
@@ -339,123 +286,272 @@ function checkForOrderedPosts( hits, searchTerm ) {
...
+			40, // down
+			13, // enter
+			27 // esc
+		];
+
+		if( keyCodes.includes( event.keyCode ) ) {
+			handleUpdDown( event );
```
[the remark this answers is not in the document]
mlaroy: fixed ✅

assets/js/autosuggest.js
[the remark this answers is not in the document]
mlaroy: went with moving it out to a shared const

timwright12:
@mlaroy can you run this against our latest linting and fix up any errors/warning that come up? https://www.npmjs.com/package/@10up/eslint-config

goaround:
@mlaroy great idea, thank you for working on it!

mlaroy:
> @mlaroy can you run this against our latest linting and fix up any errors/warning that come up? https://www.npmjs.com/package/@10up/eslint-config

@timwright12 this is done - not much to resolve there, we were already using this in the dev workflow.

timwright12:
@mlaroy is the linting version updated? I can only see "@10up/eslint-config": "^1.0.9", in the package file. We just released 2.0

mlaroy:
> @mlaroy is the linting version updated? I can only see "@10up/eslint-config": "^1.0.9", in the package file. We just released 2.0
I'll check it out and make sure it's up to date

assets/js/admin.js
```diff
@@ -1,39 +1,110 @@
-import jQuery from 'jquery';
-import { epAdmin, ajaxurl } from 'window';
+/* eslint-disable import/no-unresolved */
```
nicholasio, line 1: Why is this necessary? If the module is installed as a npm this should not be necessary.

assets/js/admin.js
```diff
@@ -1,39 +1,110 @@
...
+
+	/**
+	 * Handles the change event
+	 *
+	 * @param {event} event - change event from the range sliders
+	 */
+	const handleRangeChange = (event) => {
+		const rangeInput = event.target;
+		const label = rangeInput.previousElementSibling.querySelector('.weighting-value');
+		label.innerHTML = rangeInput.value;
```
nicholasio, line 64: Use `label.textContent` to be safe.

assets/js/admin.js
```diff
@@ -1,39 +1,110 @@
...
+			'input[type=range]',
+		);
+		const weightDisplay = rangeInput.previousElementSibling.querySelector('.weighting-value');
+
+		// toggle range input
+		rangeInput.setAttribute('disabled', !checkbox.checked);
+
+		// get new weight display value, and set it
+		const newWeightDisplay = !checkbox.checked ? '0' : rangeInput.value;
+		weightDisplay.innerHTML = newWeightDisplay;
```
nicholasio, line 97: use `textContent`

assets/js/stats.js
```diff
@@ -1,5 +1,13 @@
+/* eslint-disable import/no-unresolved, no-plusplus, no-unused-vars */
```
nicholasio, line 1: I believe `import/no-unresolved` shouldn't be necessary. And instead of disabling `no-unused-vars` we can just delete those variables.

assets/js/admin.js
mlaroy: @nicholasio could be an oversight on my part as I was trying to get the new eslint config settings working. I will revisit this.

assets/js/stats.js
mlaroy: @nicholasio most likely the same as the other - trying to get eslint to work. I'll revisit this also.

assets/js/admin.js
```diff
@@ -1,39 +1,110 @@
...
+	 * Handles the toggling and disabling
+	 *
+	 * @param {event} event - click event from the checkboxes
+	 */
+	const handleCheckboxChange = (event) => {
+		const checkbox = event.target;
+		const rangeInput = checkbox.parentNode.nextElementSibling.querySelector(
+			'input[type=range]',
+		);
+		const weightDisplay = rangeInput.previousElementSibling.querySelector('.weighting-value');
```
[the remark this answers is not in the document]
mlaroy: Fair enough - just wondering if we want to solve for that now, or stick with the vanilla version of what was jQuery, since this is the logic that existed previously with jQuery:

```
const $rangeInput = $checkbox.parent().next( '.weighting' ).find( 'input[type=range]' );
const $weightDisplay = $rangeInput.prev( 'label' ).find( '.weighting-value' );
```

How likely are we to change the markup?

assets/js/stats.js
mlaroy: @nicholasio for `no-unused-vars`, I think it is the cleanest way since we are creating new instances of the charts like `const documentChart = new Chart(...`. Alternatively, simply calling `new Chart(...` without assigning creates different eslint warnings about creating side effects.

Is there a better approach?

assets/js/admin.js
mlaroy: @nicholasio changed this ✅

assets/js/admin.js
mlaroy: @nicholasio changed this ✅

assets/js/stats.js
mlaroy: removed `import/no-unresolved` ✅

assets/js/autosuggest.js
```diff
@@ -1,466 +1,567 @@
...
  * Submit the search form
- * @param object $localInput
+ *
+ * @param {Node} input - input element
  */
-function submitSearchForm( $localInput ) {
-	$localInput.closest( 'form' ).submit();
+function submitSearchForm(input) {
+	input.closest('form').submit();
 }
```
cedarstay, line 17:
```suggestion
/**
	* Set the expanded aria state on the input
	*
	* @param bool haveOptions - whether or not the autosuggest list contains results
	* @param {Node} input - search input
	*/
function toggleInputAria( haveOptions, input ) {
	input.setAttribute( 'aria-expanded', haveOptions );
}

/**
	* Set the active descendant aria attribute input
	*
	* @param string id - id of the currently selected element
	* @param {Node} input - search input
	*/
function setInputActiveDescendant( id, input ) {
	input.setAttribute( 'aria-activedescendant', id );
}
```
Lets assistive technology when the autosuggest is open or closed and which item in the list is currently selected.

assets/js/autosuggest.js
```diff
@@ -1,466 +1,567 @@
...
-	 */
-	$epInput.each( ( key, input ) => {
-		const $epContainer = jQuery( '<div class="ep-autosuggest-container"></div>' );
-		const $input = jQuery( input );
+/**
+ * init method called if the epas endpoint is defined
+ */
+function init() {
+	const epInputNodes = document.querySelectorAll(
+		`.ep-autosuggest, input[type="search"], .search-field, ${epas.selector}`,
```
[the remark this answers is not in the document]
mlaroy: interesting - we've had a lot of discussion on this, and I believe the consensus has been that we need to have it be this generic for the average/new user, so that it just "works" without having to make this configuration. [This PR](https://github.com/10up/ElasticPress/pull/1597) seeks to add a new field for search fields to exclude, but this current PR solves for the issue linked on 1597 about excluding the search in the facets widget. Maybe you want to weigh in @brandwaffle ?

assets/js/autosuggest.js
```diff
@@ -1,466 +1,567 @@
...
+		// TODO: use some regex magic to match upper/lower/capital case??
+		const highlightedText = escapedText.replace(
+			value,
+			`<span class="ep-autosuggest-highlight">${value}</span>`,
+		);
+		itemString += `<li class="autosuggest-item">
+				<a href="${url}" class="autosuggest-link" data-search="${escapedText}" data-url="${url}">
+					${highlightedText}
+				</a>
+			</li>`;
```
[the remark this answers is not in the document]
mlaroy: thanks!  I've adapted this a bit but it seems to work 👍🏼

assets/js/autosuggest.js
mlaroy: adding these in, thanks!

assets/js/autosuggest.js
```diff
@@ -1,466 +1,567 @@
...
+	 * Calls a debounced function to get the search results via
+	 * ajax request.
+	 *
+	 * @param {event} event - keyup event
+	 */
+	const handleKeyup = (event) => {
+		event.preventDefault();
+
+		if (event.keyCode === 27) {
+			hideAutosuggestBox();
```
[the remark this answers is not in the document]
mlaroy: adding these in, thanks!

assets/js/autosuggest.js
```diff
@@ -1,466 +1,567 @@
...
 
-			if( queryJSON.error ) {
-				return;
+		/**
+		 * helper function to selected the next result
+		 */
+		const selectNextResult = () => {
+			if (currentIndex >= 0) {
+				const el = results[currentIndex];
+				el.classList.add('selected');
```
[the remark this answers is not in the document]
mlaroy: adding this in, thanks!

assets/js/admin.js
```diff
@@ -1,39 +1,110 @@
...
+	const rangeInputs = document.querySelectorAll('.weighting-settings input[type=range]');
+
+	/**
+	 * Handles the change event
+	 *
+	 * @param {event} event - change event from the range sliders
+	 */
+	const handleRangeChange = (event) => {
+		const rangeInput = event.target;
+		const label = rangeInput.previousElementSibling.querySelector('.weighting-value');
```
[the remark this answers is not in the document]
dinhtungdu: @mlaroy @tlovett1 I created a similar helper for refactoring `admin-bar.js`, we can use that too: https://core.trac.wordpress.org/browser/trunk/src/js/_enqueues/lib/admin-bar.js#L424

assets/js/autosuggest.js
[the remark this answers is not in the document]
dinhtungdu: What about excluding search input with a class or attribute?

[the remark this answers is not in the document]
dinhtungdu: What about excluding search input with a class or attribute?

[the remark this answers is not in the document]
mlaroy: @dinhtungdu that's exactly what the linked PR does, providing a field where you can exclude using a selector

assets/js/admin.js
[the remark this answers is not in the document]
mlaroy: @dinhtungdu thanks - I already have a polyfill for the `.closest()` method which I can leverage and keep it simple

assets/js/autosuggest.js
```diff
@@ -1,466 +1,613 @@
...
+			return;
+		}
+
+		if (keyCodes.includes(event.keyCode)) {
+			handleUpDown(event);
+			return;
+		}
+
+		const input = event.target;
+		const debounceFetchResults = debounce(fetchResults, 200);
```
ivanlopez, line 595: I would make sure to test that the denounce is working we were using this code on a site that went live last week and it ended up just sending all the request per character after the 200ms. We ended up using the lodash throttle method and placing the initiation of the throttle outside of this method.

mlaroy: I have tested and it is working correctly. Do you think throttle is a better function to use here, over debounce?

assets/js/admin.js
```diff
@@ -1,39 +1,110 @@
...
+	);
+
+	/**
+	 * Handles the toggling and disabling
+	 *
+	 * @param {event} event - click event from the checkboxes
+	 */
+	const handleCheckboxChange = (event) => {
+		const checkbox = event.target;
+		const rangeInput = checkbox.parentNode.nextElementSibling.querySelector(
```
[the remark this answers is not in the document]
mlaroy: @tlovett1 I've changed this using the `closest()` helper

assets/js/admin.js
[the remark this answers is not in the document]
mlaroy: @tlovett1 I've changed this using the `closest()` helper

assets/js/autosuggest.js
```diff
@@ -1,466 +1,610 @@
...
+	epInputs.forEach((input) => {
+		input.addEventListener('keyup', handleKeyup);
+		// input.addEventListener( 'blur', hideAutosuggestBox );
+	});
+}
+
+// Ensure we have an endpoint URL, or
+// else this shouldn't happen
+if (epas.endpointUrl && epas.endpointUrl !== '') {
+	init();
```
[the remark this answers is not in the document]
mlaroy: done!

PR #1697, opened 2020-03-11
WIP: Properly populate Index Health list of indices

### LINKED ISSUE #1502: Index health dashboard shows all indexes
**Describe the bug**
When running a multisite or multiple indexes within the same ElasticSearch instance the new index health point will show all indexes and not just the relevant indexes.

**Steps to Reproduce**
- Install in a multisite from develop branch. 
- Activate plugin with same credentials/endpoint on at least two sites
- Go to index-health on either site and you will see both indexes listed in index list/documents

**Expected behavior**
Unless network activated (and in network admin) the index health dashboard should limit the results to the current site/index.

**Screenshots**
https://i.imgur.com/qdZTtwF.png

moraleida:
### Description of the Change

Checks if if this is a single or multisite install before selecting which indices to display in the Index Health listing. Fixes #1502 

Includes a new constant and some minor refactoring on how to access the plugin's main file.

### Benefits

In the current implementation, all general data is queried from the  `_all` key returned from the `_stats ` endpoint in Elasticsearch. This can be problematic because it may contain other indexes that are not related to the ones attached to ElasticPress - and the data from these other indexes (indexing operations, query times, etc) would skew our numbers.

This implementation moves towards a more direct selection of which indexes to fetch data from, using sensible defaults: if this is a network activated install, all indexable indices are returned. If this is a site within a network or a single site, only the data from the specific site is returned.

In exchange we're introducing a new filter, `ep_index_health_stats_indices` which developers can use to bring data from other indices into the dashboard if they prefer.

### Things to be aware of

In doing the change to selectively include indices we are forced to choose when to read global data from the cluster and when to re-build totals based on the selected indices. I chose to separate concerns based on the Dashboard views such that:

- For the 4 first dashboard displays (Index List, Queries and Indexing Total, Documents and Totals) we're using a sum of the data available in each index. In the code I'm calling this data as representative of each index capacity and metrics.

- For the last dashboard display (Queries and Indexing Time), we're using the data from the entire cluster. The reasoning for this is that this is performance data, so it is more relevant for the user to check how well their cluster is performing in general than to separate and average the performances of each index.

### Verification Process

Tested on multisite and single-site installs, network activated and site-by-site activated.

### Checklist:


includes/classes/Stats.php
```diff
@@ -144,6 +129,92 @@ public function build_stats( $force = false ) {
...
+		$this->localized['query_total']   += absint( $this->stats['indices'][ $index_name ]['total']['search']['query_total'] );
+		$this->localized['suggest_total'] += absint( $this->stats['indices'][ $index_name ]['total']['search']['suggest_total'] );
+
+		$this->localized['total_index_time_in_millis']   += absint( $this->stats['indices'][ $index_name ]['total']['indexing']['index_time_in_millis'] );
+		$this->localized['total_query_time_in_millis']   += absint( $this->stats['indices'][ $index_name ]['total']['search']['query_time_in_millis'] );
+		$this->localized['total_suggest_time_in_millis'] += absint( $this->stats['indices'][ $index_name ]['total']['search']['suggest_time_in_millis'] );
+
+		$this->localized['index_time_in_millis']   = ceil( $this->localized['total_index_time_in_millis'] / $i );
+		$this->localized['query_time_in_millis']   = ceil( $this->localized['total_query_time_in_millis'] / $i );
+		$this->localized['suggest_time_in_millis'] = ceil( $this->localized['total_suggest_time_in_millis'] / $i );
```
moraleida: @tlovett1 this feature is ready to go, except I need a second pair of eyes on these lines. Judging by the labels in the dashboard, these should be averages. However, if we average a very busy index with a few stale indexes in the same network we might not get any valuable information out of this. If you have any info on how exactly the averages from the `_all` key are calculated we can make the same here.

moraleida: Ok. After talking to @ayebare I've changed this approach to use the data from `_all` for the averages, as explained in the updated description.

tlovett1:
I'm testing this locally network activated. After indexing, I'm not seeing terms or user indexes show up at all in the health dashboard. Have you tested this in that scenario?

tlovett1:
@moraleida is this ready for testing?

tlovett1:
Seems to work well. Nice work!

moraleida:
props to @barryceelen for the idea of fetching indexables dynamically :tada:

Issue #834, opened 2017-05-16
400 Bad Request when using aggregations with use-filter = true


elvismdev:
I have this code to search only posts with a given category and also add an aggregation of category terms.   

```
add_action(
    'pre_get_posts',
    function set_search_filter( $query ) {
		// Exit this function if we are not in a search query.
		if ( ! $query->is_search() ) {
			return;
		}

		$tax_query = array(
			array(
				'taxonomy' => 'category',
				'field'    => 'slug',
				'terms'    => array( 'una-categories' )
				)
			);
		$query->set( 'tax_query', $tax_query );


		$aggs = array(
			'name' => 'category',
			// 'use-filter' => true,
			'aggs' => array(
				'terms' => array(
					'size' => 100,
					'field' => "terms.category.slug"
					)
				)
			);
		$query->set( 'aggs', $aggs );

	}
);
```

When I add/use the filter param `'use-filter' => true` in my aggregation arguments, I get a 400 Bad Request response from Elasticsearch failing exactly here at [this line on `ep_remote_request()`](https://github.com/10up/ElasticPress/blob/develop/classes/class-ep-api.php#L215) 

```
Array
(
    [headers] => Requests_Utility_CaseInsensitiveDictionary Object
        (
            [data:protected] => Array
                (
                    [content-type] => application/json; charset=UTF-8
                    [content-length] => 520
                )

        )

    [body] => {"error":{"root_cause":[{"type":"search_parse_exception","reason":"Expected [START_OBJECT] under [field], but got a [VALUE_STRING] in [terms]","line":1,"col":1251}],"type":"search_phase_execution_exception","reason":"all shards failed","phase":"query","grouped":true,"failed_shards":[{"shard":0,"index":"localhost8081wp-1","node":"vLfHZlyVSjWpQi7PhXEFYg","reason":{"type":"search_parse_exception","reason":"Expected [START_OBJECT] under [field], but got a [VALUE_STRING] in [terms]","line":1,"col":1251}}]},"status":400}
    [response] => Array
        (
            [code] => 400
            [message] => Bad Request
        )

    [cookies] => Array
        (
        )

    [filename] => 
    [http_response] => WP_HTTP_Requests_Response Object
        (
            [response:protected] => Requests_Response Object
                (
[...]

PR #1702, opened 2020-03-15
Update aggs for multiple aggregation

goaround:
### Description of the Change

With `'use-filter' => true` you have to use multiple aggregation. Otherwise, you get an 400 error.

See: https://github.com/10up/ElasticPress/issues/834

### Alternate Designs

I would make sense to include two examples: One with `'use-filter' => true` and multiple aggregation and one without 'use-filter'

### Benefits

Correct documentation


felipeelia:
(Finally) updated on https://www.elasticpress.io/documentation/article/indexables/. Thanks!

Issue #1718, opened 2020-03-25
Querying with both "cat" and "tag" parameters is only passing the "cat" parameter.


petenelson:
**Describe the bug**
When running the following query
```
$query = new \WP_Query(
	[
		'cat' => 123,
		'tag' => 'post-tag',
	]
);
```
the formatted args for Elasticsearch do not include the `post-tag` tag.

**Steps to Reproduce**
Discovered this will working on unit tests for the indexable post class.

**Expected behavior**
Formatted ES args should contain the post_tag taxonomy.

**Additional context**
This appears to be a bug in the switch statement here https://github.com/10up/ElasticPress/blob/develop/includes/classes/Indexable/Post/Post.php#L719
I'll be submitting a PR to address this.

petenelson:
I believe this should also be updated to `term_id` https://github.com/10up/ElasticPress/blob/develop/includes/classes/Indexable/Post/Post.php#L731

tlovett1:
@felipeelia is confirming this is working.

tlovett1:
@felipeelia does this work?

dinhtungdu:
Tested with the latest commit from `develop`, look like the issue has been fixed. both tag and cat are included in the formatted args.
```
                {
                    "bool": {
                        "must": [
                            {
                                "terms": {
                                    "terms.category.term_id": [
                                        "123"
                                    ]
                                }
                            },
                            {
                                "terms": {
                                    "terms.post_tag.slug": [
                                        "post-tag"
                                    ]
                                }
                            },
                            {
                                "terms": {
                                    "terms.post_tag.slug": [
                                        "post-tag"
                                    ]
                                }
                            }
                        ]
                    }
                },
```

PR #1725, opened 2020-04-06
Expanded indexable post unit test coverage

petenelson:
### Expanded Post unit test coverage
* Increased the code coverage on the \ElasticPress\Indexable\Post class to 100%
* Fixed a bug where both `cat` and `tag` parameters in the root WP_Query args would only pass one of those values to ES.
* Added the `false` bool value to an `ignore_sticky_posts` array check to match the strict type comparison it is doing.

### Benefits

Improved unit tests and bug fixes

### Possible Drawbacks

None

### Verification Process

Verified all existing unit tests passed and added new unit tests to verify the bug fixes.

### Checklist:

### Applicable Issues

#1718 

### Changelog Entry
* Fixed a bug where both `cat` and `tag` parameters in the root WP_Query args would only pass one of those values to ES.
* Added the `false` bool value to an `ignore_sticky_posts` array check to match the strict type comparison it is doing.

### DIFF
--- a/includes/classes/Indexable/Post/Post.php
+++ b/includes/classes/Indexable/Post/Post.php
@@ -1523,9 +1523,13 @@ class Post extends Indexable {
 	 * @return string The sanitized 'order' query variable.
 	 */
 	protected function parse_order( $order ) {
+		// Core will always set sort order to DESC for any invalid value,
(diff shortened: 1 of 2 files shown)


tlovett1:
Awesome!

Issue #1722, opened 2020-03-28
Weighting engine not working with Woocommerce search


oscarssanchez:
**Describe the bug**

It looks like the Weighting feature doesn't work properly when doing a product search. As a result, the weighting configuration gets ignored. The reason for this seems to be that when doing a Woocommerce search, `search_fields` get set: 
https://github.com/10up/ElasticPress/blob/develop/includes/classes/Feature/WooCommerce/WooCommerce.php#L351

Which is why later, this logic will be true and no weighting will get applied: 
https://github.com/10up/ElasticPress/blob/develop/includes/classes/Feature/Search/Weighting.php#L514 

If this logic is removed form the code, weighting is applied successfully. However, some of the default fields set from the feature get lost: `_sku`, `terms.category.name` and `terms.post_tag.name`. 

**Steps to Reproduce**

1.- Activate Woocommerce feature
2.- Add some weighting to the product post type
3.- Do a search with `post_type product` set
4.- With debug bar, see no weighting has been applied

**Expected behavior**
Weigthing should get applied based on weighting settings

**Screenshots**

**Environment information**
 - Device: 
 - OS: 
 - Browser and version: 
 - WordPress version:   WordPress 5.2
 

 - Plugins and version:  Elasticpress 3.4
 - Theme and version: 
 
 - <details><summary>Site Health Info:</summary>
   
   </details>

**Additional context**

tlovett1:
@moraleida can you take a look?

felipeelia:
As I have a recent background with this, I can work on it.

felipeelia:
As we've merged #1801 to fix this one, I'm closing it now.

PR #1732, opened 2020-04-08
Add WooCommerce fields without overwriting weight config

felipeelia:
### Description of the Change

As per #1722 , when we add the WooCommerce fields to the query via `search_fields` parameter, the weighting engine is completely ignored. This PR aims to add the same fields but keeping the weighting config.

This PR also creates a method that tries to guess the best way to add a field to the query weighting config: `Weighting::add_weighted_fields()`

### Benefits

Keep weighting functionality AND having WC fields.

### Verification Process

I've checked the ES query in Debug Bar.

### Checklist:

### Applicable Issues

#1722 

### Changelog Entry

Add WooCommerce fields to the query without removing weighting configurations.


felipeelia:
As I've completely messed up with the git history of this branch, I'll close it in favor of #1785 that has a much cleaner timeline. Sorry!

PR #1737, opened 2020-04-14
fixed not correct generating elastic queries for taxonomies

renakdup:
**Describe the bug**
For Custom Taxonomies elasticpress is generating not correct query.

**Steps to Reproduce**
If we want use Custom Taxonomy for our query, we fill $args for wp_query and we get next generated query for elastic:
```
{
  "from": 0,
  "size": 40,
  "query": {
    "match_all": {
      "boost": 1
    }
  },
  "post_filter": {
    "bool": {
      "must": [
        {
          "bool": {
            "should": [
              {
                "terms": {
                  "terms.cutom_tax_name.id": [
                    11111
                  ]
                }
              }
            ]
          }
        },

        {
          "terms": {
            "post_type.raw": [
              "post"
            ]
          }
        },
        {
          "term": {
            "post_status": "publish"
          }
        }
      ]
    }
  }
}
```
And elastic by this query not found some posts, because code formatting terms  `terrms.cutom_tax_name.id` has error. We need use `term_id` instead `id`, because in elastic we field stored under the name `term_id`

**Expected behavior**

**Screenshots**

**Environment information**
 - Browser and version: 
 - WordPress version:  5.3.2
- ElasticPress version 3.4.1
   </details>

**Additional context**

### DIFF
--- a/includes/classes/Indexable/Post/Post.php
+++ b/includes/classes/Indexable/Post/Post.php
@@ -1416,7 +1416,12 @@ class Post extends Indexable {
 			$single_tax_query = $tax_queries;
 			if ( ! empty( $single_tax_query['taxonomy'] ) ) {
 				$terms = isset( $single_tax_query['terms'] ) ? (array) $single_tax_query['terms'] : array();
-				$field = ( ! empty( $single_tax_query['field'] ) ) ? $single_tax_query['field'] : 'term_id';
+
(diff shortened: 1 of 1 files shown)


tlovett1:
@renakdup seems like this might be an error in your query. According to https://developer.wordpress.org/reference/classes/wp_query/ `fields` should not contain `id`

PR #1748, opened 2020-04-23
Allow custom queries in painless scripting language

scottbuckel:
This allows the user to write a custom query in the painless scripting language https://www.elastic.co/guide/en/elasticsearch/painless/current/index.html. 

For example, to search for a post by meta.birthday for all people born in March, you can use the following code:

`$args['custom_query'][] = "doc['meta.birthday'].value.getMonthValue() == 3";`

### DIFF
--- a/docs/indexables.md
+++ b/docs/indexables.md
@@ -94,6 +94,7 @@ ElasticPress integrates with `WP_Query` if the `ep_integrate` parameter is passe
     ) );
     ```
 * ```custom_query``` (*string*)
+
     ```custom_query``` allows the user to write a custom query in the painless scripting language [https://www.elastic.co/guide/en/elasticsearch/painless/current/index.html](https://www.elastic.co/guide/en/elasticsearch/painless/current/index.html).
     ```php 
           $args['custom_query'][] = "doc['meta.birthday'].value.getMonthValue() == 3";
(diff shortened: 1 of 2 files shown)


tlovett1:
@sebumd this is interesting. Assuming it works and @tott and @brandwaffle, are okay with it. I think we can add this. However, first please run your PR through PHPCS as there are some issues.

sebumd:
@tlovett1 thanks! I fixed the PHPCS issue. 

@tott @brandwaffle what's the process to get this tested, approved and released for ElasticPress? Right now I'm using a hacked version of the module in production and would like to be able to include this feature and update to the latest when it comes out.

Rahmon:
Hi @sebumd.

I would like to review this PR. Could you provide more information to test it? I've tried but looks like the filter is not applied.

sebumd:
@Rahmon This is how I'm using it myself:

```
 if (isset($_GET['birthDateYear']) && !empty($_GET['birthDateYear'])) {
    $birthDateYear = esc_html($_GET['birthDateYear']);
    $args['custom_query'][] = "doc['meta.birthdate.date'].value.getYear() == " . $birthDateYear;
}
```

felipeelia:
Putting this back into consideration for EP 5.3.0, but before doing that we need to assess if there is any performance impact on enabling painless on ES servers.

Issue #1740, opened 2020-04-16
Duplicate non-blocking requests


adiloztaser:
**Describe the bug**
When the user changes the max number of times to attempt failed remote requests with the filter `ep_max_remote_request_tries`,  ElasticPress tries to make non-blocking requests repeatedly because if the request is non-blocking, `wp_remote_request` returns an empty string as a result and empty string is not a valid response.

I think there are two options for solving this problem:
1. find a way to find out that the non-blocking request returns a real error
2. make sure non-blocking requests are not repeat

**Steps to Reproduce**
1. Return a max limit more than 1 with filter `ep_max_remote_request_tries`
2. Add any debug logic into the `Elasticsearch->remote_requst()`. I added failures count to options table `update_option('es_error_count', $failures );` 
3. Update a published post as a draft.
4. See the option value

felipeelia:
Hi @adiloztaser ,
I've tried to replicate this issue but I couldn't. Do you mind sharing with us the details of your environment? WordPress, ElasticPress, ElasticSearch versions, and which web server (Apache, Nginx, or some other) on which version would be helpful.
Thanks!

adiloztaser:
Hi @felipeelia,

I am using ElasticPress 3.4.1 with Elasticsearch 6.3.2 on WordPress 5.3.2

For more detail here is my test patch:

```
diff --git a/wp-content/plugins/elasticpress/includes/classes/Elasticsearch.php b/wp-content/plugins/elasticpress/includes/classes/Elasticsearch.php
index 48a437d4e..5da765bc6 100644
--- a/wp-content/plugins/elasticpress/includes/classes/Elasticsearch.php
+++ b/wp-content/plugins/elasticpress/includes/classes/Elasticsearch.php
@@ -929,6 +929,13 @@ class Elasticsearch {
 			if ( false === $request || is_wp_error( $request ) || ! $is_valid_res ) {
 				$failures++;
 
+				$log = array(
+					'count' => $failures,
+					'args' => $args
+				);
+
+				update_option( 'es_error', json_encode( $log ) );
+
 				/**
 				 * Filter max number of times to attempt remote requests
 				 *
@@ -938,7 +945,7 @@ class Elasticsearch {
 				 * @param  {args} $args Request arguments
 				 * @return {int} New number of tries
 				 */
-				if ( $failures >= apply_filters( 'ep_max_remote_request_tries', 1, $path, $args ) ) {
[...]

PR #1752, opened 2020-04-23
EP Requests: Don't retry if is non-blocking request

felipeelia:
### Description of the Change

As related in #1740, non-blocking requests return empty bodies and response codes and that is interpreted as an error by ElasticPress, that retries the query. This PR aims to avoid that behavior, simply avoiding a retry if the response code is 0.

### Alternate Designs

As non-blocking requests will fail async, I don't think we have much choice here.

### Benefits

Stop EP from doing unnecessary requests.

### Possible Drawbacks

none

### Verification Process

1. Set a plugin with 
```
add_filter(
	'ep_max_remote_request_tries',
	function( $num ) {
		return 5;
	}
);
```
2. Added this to the `Elasticsearch::remote_request()`
```
error_log( $failures . ' / ' . apply_filters( 'ep_max_remote_request_tries', 1, $path, $args ) . ' / ' . $request_response_code );
```
3. Saved a post as Draft and
4. Read logs.

### Checklist:

### Applicable Issues

#1740

### DIFF
--- a/includes/classes/Elasticsearch.php
+++ b/includes/classes/Elasticsearch.php
@@ -924,9 +924,10 @@ class Elasticsearch {
 
 			$request_response_code = (int) wp_remote_retrieve_response_code( $request );
 
-			$is_valid_res = ( $request_response_code >= 200 && $request_response_code <= 299 );
+			$is_valid_res            = ( $request_response_code >= 200 && $request_response_code <= 299 );
+			$is_non_blocking_request = ( 0 === $request_response_code );
 
-			if ( false === $request || is_wp_error( $request ) || ! $is_valid_res ) {
+			if ( false === $request || is_wp_error( $request ) || ( ! $is_valid_res && ! $is_non_blocking_request ) ) {
 				$failures++;
 
 				/**


includes/classes/Elasticsearch.php
```diff
@@ -924,9 +924,10 @@ public function remote_request( $path, $args = [], $query_args = [], $type = nul
 
 			$request_response_code = (int) wp_remote_retrieve_response_code( $request );
 
-			$is_valid_res = ( $request_response_code >= 200 && $request_response_code <= 299 );
+			$is_valid_res            = ( $request_response_code >= 200 && $request_response_code <= 299 );
+			$is_non_blocking_request = ( 0 === $request_response_code );
 
-			if ( false === $request || is_wp_error( $request ) || ! $is_valid_res ) {
+			if ( false === $request || is_wp_error( $request ) || ( ! $is_valid_res && ! $is_non_blocking_request ) ) {
```
tlovett1, line 930: Why not check `$query['blocking']`?

felipeelia: As we call `wp_remote_request( $query['url'], $args )` (passing only the URL and not the whole `$query` array), having `$query['blocking']` as `true` doesn't really affect the request. If we had `$args['blocking'] = true;` then we'd block the request and could use that as a conditional.
In fact, we have a `if ( isset( $args['blocking'] ) && false === $args['blocking'] )` (testing against `$args`) later in the code but even that can leave to a wrong assumption: we can send `$args['blocking']` as `true` for `wp_remote_request()` and it can be changed later by some other code through the `http_request_args` filter.

tlovett1: Is there somewhere in the docs that says wp_remote_request will always return 0 for a non-blocking request?

felipeelia: No, unfortunately it is not documented. But `wp_remote_request()` is a wrapper for `WP_Http::request` that contains this code:
```
if ( ! $parsed_args['blocking'] ) {
	return array(
		'headers'       => array(),
		'body'          => '',
		'response'      => array(
			'code'    => false,
			'message' => false,
		),
		'cookies'       => array(),
		'http_response' => null,
	);
}
```
The function we're using to retrieve the status code, `wp_remote_retrieve_response_code()`, returns `$response['response']['code']`, which will be `false`. As we're using it like this
```
$request_response_code = (int) wp_remote_retrieve_response_code( $request );
```
That `(int)` will cast `false` to `0`.

PR #1766, opened 2020-05-08
Allow filtering sticky posts support

shadyvb:
Allow users to turn on sticky posts handling for any request not just when `is_home() === true`, like for REST API calls for example.

### Description of the Change

This introduces a new filter that allows developers to enable sticky posts on other requests than when `is_home()` is true, eg: for REST API calls ( to provide parity with server-side preloaded states ).

### Benefits

- Allows parity between Server-side loaded content and REST API calls if needed.
- Allows more flexibility for developers to enable/disable sticky posts at their discretion for any query.
 

### Possible Drawbacks

- This does not change any logic, just adds the filter.

### Verification Process

- This does not change any logic, just adds the filter.

### Checklist:

### Changelog Entry

- Adds `ep_enable_sticky_posts` filter to control sticky posts activation per query.


tlovett1:
Looks good.

PR #1771, opened 2020-05-13
CI: Use Travis jobs to speed up testing process

dinhtungdu:
### Description of the Change
We're using one job for linting, wpa and unit test. These three tasks are run in sequence and it's hard to check which test causes the build fails when it happens.

This PR split the testing stage into 3 jobs. This makes the test run faster because jobs are run in parallel. In addition, it's easier to see which test is failed and easier to trigger the build manually that targets only one test suite.

### Verification Process
See Travis passes.

### Checklist:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,13 +20,6 @@ env:
   global:
   - WP_VERSION=latest WP_MULTISITE=1 ES_VERSION="7.5.0"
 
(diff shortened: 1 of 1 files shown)


tlovett1:
@tuanmh wow! Didn't know this was possible. Great idea.

PR #1779, opened 2020-05-18
Improve check for WP_IMPORTING in SyncManager classes

### LINKED ISSUE #1757: WP_IMPORTING check in SyncManager too early
The SyncManager setup() methods [contain a check if WP_IMPORTING is true](https://github.com/10up/ElasticPress/blob/develop/includes/classes/Indexable/Post/SyncManager.php#L30), and do not add their hooks if WP_IMPORTING is true.

It seems this check is performed on the `init` hook, which can be too early: other functionality may define WP_IMPORTING on a later hook.

To improve Elasticpress responding to WP_IMPORTING and prevent the need for other functionality to define WP_IMPORTING on an earlier hook, we could add the conditional to each callback function in the SyncManager classes separately in stead of in the `setup()` function.

An alternative to looking at WP_IMPORTING could be adding a specific filter that prevents indexing.

barryceelen:
Fixes #1757

This update adds a conditional that checks if WP_IMPORTING is `true` to each callback function in the `SyncManager` classes for each indexable  in stead of performing that check in the `SyncManager` class `setup()` function.

A `kill_sync()` method is added to the `SyncManager` base class that is used to perform the check against WP_IMPORTING. It contains an `ep_sync_indexable_kill` filter that allows overriding the default behaviour.


tlovett1:
Awesome! Thank you.

PR #1780, opened 2020-05-19
Facets widget: Move <div> outside ob_start()

kallehauge:
### Description of the Change

This PR makes the HTML that is passed to the `ep_facet_search_widget` filter valid. This is required because `ob_start()` is started inside the `<div class="terms">` element but closing element is included in the `ob` result

### Benefits
The main benefit here is that it is valid/complete HTML that is being passed.
It is not necessarily a problem is the markup is only being manipulated with str_replace() or similar functions but HTML parsers like `DOMDocument()` will try to auto-correct broken markup and add/remove a `<div>` which will, in the end, cause broken markup

### Checklist:

### DIFF
--- a/includes/classes/Feature/Facets/Widget.php
+++ b/includes/classes/Feature/Facets/Widget.php
@@ -270,9 +270,9 @@ class Widget extends WP_Widget {
 					</div>
 				<?php endforeach; ?>
 			</div>
+			<?php $facet_html = ob_get_clean(); ?>
 		</div>
 		<?php
-		$facet_html = ob_get_clean();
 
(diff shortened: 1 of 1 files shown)


VoroDk:
I've tested this, and it fixes the issue. As stated in the PR, the widget filters output contains a closing div, that is opened outside of the passed output. This is fine when the widget run as is, but when manipulating the output with DomDocument(), the HTML is re-validated and throws a warning: 
```
Warning: DOMDocument::loadHTML(): Unexpected end tag : div in Entity
```

tlovett1:
Thank you!

PR #1789, opened 2020-05-27
Exclude password protected posts from search

cristianuibar:
### Description of the Change
The problem we faced was the display of password-protected posts in the search results. Upon further investigation we were not able to hide password-protected posts via any existing hooks. Looking at past issues, we found this one: #1083 which had a PR: #1094 that was reverted at a later time. Upon testing the PR we managed to successfully hide posts from search results using it hence I am re-opening the PR with an additional hook to be able to show posts regardless of admin status if one might want that.
 
### Alternate Designs
We attempted to hide password-protected posts from the search results (which is default WP behavior) via existing hooks with no success.

### Benefits
Hide password-protected posts from search results. This behavior can be disabled via the included hook.

### Possible Drawbacks
None of which I can tell so far.

### Verification Process
Upon implementing this code on our test and production sites, we re-run the indexing process and then tested the default WP search results in the front-end of the site using different keywords that we knew previously returned password-protected posts.

### Checklist:

### Applicable Issues
None so far.

### Changelog Entry
Add support for hiding password-protected posts from search-results. Props @allan23 (#1094)
Add filter hook so you can still display password-protected posts from search-results. @cristianuibar

### DIFF
--- a/includes/classes/Indexable/Post/Post.php
+++ b/includes/classes/Indexable/Post/Post.php
@@ -521,6 +521,16 @@ class Post extends Indexable {
 		 */
 		$meta = apply_filters( 'ep_prepare_meta_data', (array) get_post_meta( $post->ID ), $post );
 
+		/**
+		 * Make sure that proper post meta is being saved for
+		 * password protected posts.
+		 * 
+		 * @since 3.4.2
+		 */
+		if ( ! empty( $post->post_password ) ) {	
+			$meta['password_protected'] = true;	
+		}
+
 		if ( empty( $meta ) ) {
 			/**
 			 * Filter final list of prepared meta.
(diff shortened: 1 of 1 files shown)


tlovett1:
Hmm, interesting.

Excluding password protected posts absolutely makes sense. I think there is a better way to do it than tricking the plugin into using meta.

I also wonder if ElasticPress should just not index password protected posts which people should not be using anyway.

dinhtungdu:
`WP_Query` has an argument named `has_password`. IMO, we can add support for that arg to Post Indexable, then from another plugin/theme, we can filter the search query to exclude password-protected posts.

tlovett1:
@dinhtungdu can you create a PR for that?

sagarsdeshmukh:
Hi @tlovett1 , Are you planning to merge one PR from [#1789](https://github.com/10up/ElasticPress/pull/1789) or [#1913](https://github.com/10up/ElasticPress/pull/1913) which allow to skip password protected content from search?

mckdemps:
Being addressed through #1913

PR #1801, opened 2020-06-12
[WooCommerce] Add Fields to the Weighting Dashboard

felipeelia:
### Description of the Change

As per #1722 , when we add the WooCommerce fields to the query via search_fields parameter, the weighting engine is completely ignored.

This PR aims to add WooCommerce fields to the _Manage Search Fields & Weighting_ Screen, giving the user the ability to enable or not WooCommerce fields and change their weights. This is a cleaner and more elegant approach than the one suggested on #1785 .

### Screenshot

This is how the _Product_ post type box looks like with the initial version of the PR:

image

Note the "SKU" field and the absence of the "Author", which is removed by the current code.

### Benefits

Keep weighting functionality AND having WC fields.

### Verification Process

I've checked the ES query in Debug Bar

### Checklist:

### Applicable Issues

#1722 

### Changelog Entry

Add WooCommerce fields to the _Manage Search Fields & Weighting_ Screen.

### DIFF
--- a/includes/classes/Feature/WooCommerce/WooCommerce.php
+++ b/includes/classes/Feature/WooCommerce/WooCommerce.php
@@ -433,25 +433,6 @@ class WooCommerce extends Feature {
 					);
 
 					$query->set( 'search_fields', $search_fields );
-				} elseif ( 'product' === $post_type ) {
-					$search_fields = $query->get( 'search_fields', array( 'post_title', 'post_content', 'post_excerpt' ) );
-
-					// Remove author_name from this search.
-					$search_fields = $this->remove_author( $search_fields );
-
-					foreach ( $search_fields as $field_key => $field ) {
-						if ( 'author_name' === $field ) {
-							unset( $search_fields[ $field_key ] );
-						}
-					}
-
-					$search_fields['meta']       = ( ! empty( $search_fields['meta'] ) ) ? $search_fields['meta'] : [];
-					$search_fields['taxonomies'] = ( ! empty( $search_fields['taxonomies'] ) ) ? $search_fields['taxonomies'] : [];
-
-					$search_fields['meta']       = array_merge( $search_fields['meta'], array( '_sku' ) );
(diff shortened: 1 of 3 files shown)


felipeelia:
I've confirmed that users upgrading their EP version will have to manually go to the Weighting Screen and enable product taxonomies and SKU, unless we do something about it.

@brandwaffle @tott @tlovett1 how do you want to handle it? I see some possibilities:
1. Enable it for them during the upgrade. I didn't find an "easy" way to do it though, as we don't seem to have a mechanism in place that deals with specificities like these. We do have a way to identify [versions that needs to be reindexed](https://github.com/10up/ElasticPress/blob/develop/elasticpress.php#L188) but we'd have to improve that (I may be missing something here).
2. Create an admin notice with it. My question here would be: what conditions we'd have to meet to display the notice? I can see these:
- Both features enabled (post search and WooCommerce)
- EP not being network enabled
- The "product" in the Weighting Config array should not have sku or taxonomies (or just sku, to make the test easier)
- We'd need a flag to know that the user recently saved the weighting config and decided to not use skus or taxonomies. Right?
3. Just make a note in the changelog.

includes/classes/Feature/WooCommerce/WooCommerce.php
```diff
@@ -433,7 +433,7 @@ public function translate_args( $query ) {
 					);
 
 					$query->set( 'search_fields', $search_fields );
-				} elseif ( 'product' === $post_type ) {
+				} elseif ( 'product' === $post_type && defined( 'EP_IS_NETWORK' ) && EP_IS_NETWORK ) {
```
Rahmon, line 436: Why is it necessary to check EP_IS_NETWORK?

felipeelia: Because the weighting dashboard is not something available for multisite installations. In these cases, we should default to the current behavior.

tlovett1:
Looks good. Definitely needs lots of testing.

PR #1552, opened 2019-11-13
Add search term highlighting

mlaroy:
### Description of the Change

This pull request adds a new Feature, giving the plugin the ability to inject tags around search terms in search results. Also adds an admin screen to select the tag, add a colour, and whether to include in the_excerpt, since excerpts filter out html by default.

Screen Shot 2019-11-12 at 4 31 39 PM
Screen Shot 2019-11-12 at 4 31 56 PM
Screen Shot 2019-11-12 at 4 31 13 PM

### Benefits

Search terms can be styled to stand out within search results, like you might find in any major search engine, for visual cues to how that result is relevant to your search.

### Possible Drawbacks

On the off chance that the plugin user already has styles applied to the class applied to the injected tag, it will be styled accordingly.

### Verification Process

Inspected source code, used the EP Debug Bar to confirm the query args were being properly modified using `ep_formatted_args`, as well as verifying with the debug bar the the highlighting was returned from the ElasticSearch request.

### Checklist:

- [x] I have read the [**CONTRIBUTING**](/CONTRIBUTING.md) document.
- [x] My code follows the code style of this project.
- [x] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [ ] I have added tests to cover my change.
- [ ] All new and existing tests passed.

## 2019-11-XX

* Add search term highlighting to search results
* Add new admin screen to configure highlighting settings

PR #1804, opened 2020-06-14
Feature/search term highlight

oscarssanchez:
### Description of the Change

This PR builds on the work done by @mlaroy here https://github.com/10up/ElasticPress/pull/1552 

### Checklist:


oscarssanchez:
Hi @tlovett1 ,

As per your comments here https://github.com/10up/ElasticPress/pull/1552 

I have modified the Unit Tests so they are passing now. I have also made some other modifications like defining class properties and forced boolean values (as it looks the values were inconsistent in unit tests and inside class methods), so It would be great if you can have another look at this. 

I also noticed the change color button text gets lost if you have a color similar to blue:

color


As I am not familiar with the discussion of this feature, just wanted to flag if this is ok or do we want to have something else to see the color picked like a box next to the button.

Thanks!

PR #1808, opened 2020-06-18
Correct autosuggest notice

oscarssanchez:
### Description of the Change

Hi @tlovett1 ,

This PR simply corrects this autosuggest notice:

autosuggest

In favor of this one: `Autosuggest feature is enabled. If documents feature is enabled, your media will also become searchable in the frontend.`

As the safe values option was removed from autosuggest settings.

### Benefits

Less confusion with autosuggest.

### Checklist:

### DIFF
--- a/includes/classes/AdminNotices.php
+++ b/includes/classes/AdminNotices.php
@@ -124,7 +124,7 @@ class AdminNotices {
 		}
 
 		return [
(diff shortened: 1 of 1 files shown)


brandwaffle:
@tlovett1 @oscarssanchez I think we should still be reminding folks that their protected content will be searchable, at least if they're not using ElasticPress.io.

Issue #1800, opened 2020-06-11
The condition for building the post_status parameter in a simple query


vnmedeiros:
Hey guys!
I have a doubt, on `post_status` parameter of the WP_Query class the default value is‘ publish ‘, but if the user is logged in,‘ private ‘is added, see:
https://developer.wordpress.org/reference/classes/wp_query/#status-parameters

but the elasticpress when on perform query only adds the private status if the request is made through an administrative page and if the user is logged in, see here:
https://github.com/10up/ElasticPress/blob/master/includes/classes/Indexable/Post/Post.php#L1232

any special reason for having this difference?
Thanks!!

oscarssanchez:
Hey @vnmedeiros , 

I think you are right! I did a search with and without elasticpress. 

Without, i get some results when looking for a private post. With EP enabled, I get no results. 

I'll craft a PR for this.

oscarssanchez:
Hi @vnmedeiros ,

I made https://github.com/10up/ElasticPress/pull/1813 to address this. Could you give it a try see if it solves the issue for you?

vnmedeiros:
Hi @oscarssanchez 
cool!
I'll test, thank very much!

oscarssanchez:
@brandwaffle even with protected content enabled the query that's being run with ElasticPress does not return results. As we are returning them in the admin, I think we should include them if protected content is enabled and you are logged in as an admin or you have the correct capability

PR #1813, opened 2020-06-22
Look for private status posts if editor or admin

oscarssanchez:
### Description of the Change

Hi @tlovett1 ,

Could you please review this PR which addresses #1800?

This changes the `is_user_logged_in()` check in favor of checking if the current user is either an `editor` or an `administrator`, and displays private posts only to them. Both in the frontend and in the admin.

As per WP documentation, private posts should only be visible to editors and administrators: https://developer.wordpress.org/reference/classes/wp_query/#status-parameters and without Elasticpress enabled, I could confirm this was true.

Additionally, the PR solves another bug I spotted, which lets authors and contributors see just anyone private posts in the admin if Elasticpress is enabled.

### Benefits
Not allowed users won't see post they should not. Allowed users should be able to see private posts in the frontend.

### Verification Process
* Create a private post.
* Index with Elasticpress.
* Query in the frontend as an editor/admin and see private post.
* Query in the backend as an author/contributor and no private post is returned.

### Checklist:


oscarssanchez:
Also confirmed it fixes https://github.com/10up/ElasticPress/issues/1870

felipeelia:
As we've done a lot of work related to which posts should and should not be displayed for certain roles, I'm adding this PR to EP 5.3.0 so we can check if this is still needed or not.

burhandodhy:
@felipeelia This PR can be closed in favour of #4169

Issue #1798, opened 2020-06-10
Delete an attachment with the CLI does not tigger ES sync


photocurio:
The issue is the exact same as [WP-CLI Post Delete does not trigger es sync](https://github.com/10up/ElasticPress/issues/152), except I am deleting attachments not posts.

**Steps to Reproduce**
1. Delete an attachment with the CLI. The command is the same as for deleting a post: `wp post delete <id> --force`
2. Search for the attachment in the media library (its gone)
3. Search for the attachment with Elasticsearch (its still there)

**Expected behavior**
When I delete an attachment with the CLI, it should not be findable in search.

**Additional context**
It seems that issue 152 was resolved by hooking ES Sync to `delete_post`.  Maybe Sync can also hook to `delete_attachment`?
Also: when deleting attachments with the CLI, the --force flag must be used, because there is no trash status for attachments.

photocurio:
P.S. can a sync be triggered from the CLI? I could just run a second command, as a workaround.

photocurio:
I tried these:

-  adding `add_action( 'delete_attachment', array( $this, 'action_delete_post' ) );` to SyncManager.php in the elasticpress plugin, at line 42. This does not work.
- I also tried disabling the early return in action_delete_post, thinking maybe attachments were not in $indexable_post_types.  That did not work.

photocurio:
I resolved this in my app by sending a `_delete_by_query` post request to ElasticSearch, right after deleting the attachment.  https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-delete-by-query.html.  It's not an optimal solution, but it will do.

oscarssanchez:
Hey @photocurio ,

I did some testing here. Turns out this happens on all post types that are force deleted. I will take a look.

oscarssanchez:
Hi @photocurio ,

I made PR https://github.com/10up/ElasticPress/pull/1814 to address this. Can you please give it a shot check if this solves the issue for you?

photocurio:
OMG thank you @oscarssanchez for working on this.  I forgot all about it after I found a workaround.  But I'll test this fix.

felipeelia:
Closing this one as https://github.com/10up/ElasticPress/pull/1968 was merged.

PR #1814, opened 2020-06-23
Skip permissions check on deleting post if doing WP_CLI

oscarssanchez:
### Description of the Change

Hi @tlovett1,

Could you please review this PR fixing #1798 ?

The change adds a check to see if we are doing WP_CLI and if we are, skip the permissions check if we are deleting a post. This is because if we are doing WP_CLI, when the previous code checks if the current user can edit posts `wp_get_current_user()` is called and it will return no user, returning false and skipping the document deletion.

### Benefits

Solves #1798 

### Verification Process

* Create a post
* Delete it via WP CLI with the force flag `wp post delete <id> --force`
* Check in the admin the post is not returned with Elasticpress enabled.

### Checklist:

- [] All new and existing tests passed.

### DIFF
--- a/includes/classes/Indexable/Post/SyncManager.php
+++ b/includes/classes/Indexable/Post/SyncManager.php
@@ -141,7 +141,7 @@ class SyncManager extends SyncManagerAbstract {
 		 * @param  {int} $post_id ID of post
 		 * @return {boolean} New value
 		 */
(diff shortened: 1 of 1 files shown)


tlovett1:
This is a big problem. @oscarssanchez it feels like we need to do this check in a lot more places.

felipeelia:
Closing this one in favor of #1968 .

Issue #1818, opened 2020-06-25
Indroduce a way to specify credentials via PHP constant.


ChrisWiegman:
**Is your enhancement related to a problem? Please describe.**
While _ep_host_ configuration is available via PHP constant, this is not extended to related fields, particularly the credentials when using elasticpress.io. In order to better help our users configure elasticpress.io on their sites, we would like to see the ability to apply all required configuration items, especially credentials, via php constant.

**Describe the solution you'd like**
Implement an EP_CREDENTIALS constant that can be used in the same fashion as EP_HOST and EP_INSTALL_PREFIX

**Designs**
N/A

**Describe alternatives you've considered**
For our use case, the only other option is to inject the credentials into the database via wp-cli or similar. This, however, is not ideal in a hosted installation setting.

**Additional context**
N/A

PR #1819, opened 2020-06-25
Feature/improve credential storage

ChrisWiegman:
### Description of the Change

This implements an EP_CREDENTIALS constant to store elasticpress.io credentials outside of the WordPress database. The current change utilizes a colon delimited username:token pair in a single constant to reflect usage in ep shield and other areas in which the username and token are simply concatenated with a colon for implementation. In addition, this extends the ability to hide and other language enhancements to the credentials form in the same manner as has been done for EP_HOST and EP_INSTALL_PREFIX.

### Alternate Designs

As a follow-up, we would eventually like to see hiding of fields and other config items available via PHP constant however, this will allow us to implement elasticpress.io accounts on user sites using a consistent configuration format for all required configuration items.

### Benefits

This will allow us to implement elasticpress.io accounts on user sites using a consistent configuration format for all required configuration items.

### Possible Drawbacks

This probably doesn't go far enough. For example, hiding these same configuration variables via constant would be ideal however, this would require changes to established precedent already implemented in EP_HOST and EP_INSTALL_PREFIX. 

### Verification Process

Automated and manual testing has been included to verify correct operation of the new configuration inputs.

### Checklist:

### Applicable Issues

https://github.com/10up/ElasticPress/issues/1818

### Changelog Entry

* [Feature] Implement EP_CREDENTIALS constant to store credentials in a _username:token_ format.

### DIFF
--- a/includes/partials/settings-page.php
+++ b/includes/partials/settings-page.php
@@ -132,8 +132,23 @@ if ( defined( 'EP_IS_NETWORK' ) && EP_IS_NETWORK ) {
 									<label for="ep_username"><?php esc_html_e( 'Subscription Username', 'elasticpress' ); ?></label>
 								</th>
 								<td>
-									<input type="text" value="<?php echo esc_attr( $credentials['username'] ); ?>" name="ep_credentials[username]" id="ep_username">
-									<legend class="description"><?php esc_html_e( 'Plug in your subscription username here.', 'elasticpress' ); ?></legend>
+									<?php
+									/**
+									 * Filter whether to show epio credentials fields in admin UI or not
+									 *
+									 * @hook ep_admin_show_credentials
+									 * @param  {boolean} $show True to show
+									 * @return {boolean} New value
+									 */
+									if ( apply_filters( 'ep_admin_show_credentials', true ) ) :
+										?>
+										<input <?php if ( defined( 'EP_CREDENTIALS' ) && EP_CREDENTIALS ) : ?>disabled<?php endif; ?> type="" value="<?php echo esc_attr( $credentials['username'] ); ?>" name="ep_credentials[username]" id="ep_username">
+									<?php endif ?>
+									<?php if ( defined( 'EP_CREDENTIALS' ) && EP_CREDENTIALS ) : ?>
+										<legend class="description"><?php esc_html_e( 'Your Subscription Username is set in wp-config.php', 'elasticpress' ); ?></legend>
+									<?php else : ?>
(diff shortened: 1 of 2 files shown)


brandwaffle:
@ChrisWiegman makes sense to me! Thanks for submitting.

@oscarssanchez can you test this and let us know if it can be added to the next release?

oscarssanchez:
Hi @ChrisWiegman ,

Thanks for submitting this PR.

While testing this, it looked to me like `EP_CREDENTIALS` accomplishes the same `ES_SHIELD` does (connecting and authenticating elasticpress to elasticsearch if it is defined). `ES_SHIELD` would be defined the same way: `username:token`. Could you please clarify what would be the difference with `EP_CREDENTIALS`?

ChrisWiegman:
I looked at that but, according to this (https://10up.github.io/ElasticPress/tutorial-security.html), that's for another plugin, the Elasticsearch Shield plugin. Is that not the case? Looking at the code functions I'll admit it is not clear.

ChrisWiegman:
A bit further digging and I'm having a hard time understanding the difference between the two given the documentation. That said, if we can confirm it will work with just ES_SHIELD credentials I can modify the PR to simply allow hiding of the credentials field based on that constant alone. The question then becomes, is there any place at which the credentials for using the elasticsearch shield credentials and the elasticpress credentials would differ?

oscarssanchez:
Hi @ChrisWiegman ,

I have reached out internally and turns out `ES_SHIELD` is a carry over from years ago. It seems elasticpress.io doesn't use elasticsearch shield but custom security mechanism. 

Although `ES_SHIELD` would accomplish the same thing, do you think having a different/another constant makes sense ? It is a good point that this could be confusing for some people.

ChrisWiegman:
Hi @oscarssanchez,

Given the documentation and history of the constant would it not be better to add the es_credentials constant (and associated documentation) for now, allowing them both to work while beginning the process to deprecate the confusing es_shield constant?

This is all up to y'all of course. Please advise on how you would like me to proceed and I can update this PR accordingly. If you stick with es_shield I would still like to get the ability to hide the credentials fields like the other setup fields as is also in this PR. If needed, I can also help you by updating this in ways that could help you move closer to deprecating es_shield. That's all up to y'all.

mikelking:
+1 to what @ChrisWiegman said.

ChrisWiegman:
Hey @oscarssanchez, any thoughts on this from your end? We'd like to try to get this closed out on ours.

oscarssanchez:
@ChrisWiegman , all looks good, thanks for this!

Issue #1791, opened 2020-05-27
Search "E-mail" vs "Email"


sinusiks:
We have a post with title "Best E-mail clients for mobile phones"

- When searching for "E-mail" (without quotes) a lot of posts are found, none of them not even including word "mail" in title or content even though title has highest weight set.

- When searching for "Email" without dash, the correct post is found and also a first result.

ElasticPress is configured with default settings, except titles have highest weight set. If user searches for "E-mail" or "mobile e-email" or "mobile client" etc. I would want this post to be one of the first results. What can I do to fine-tune this kind of search?

oscarssanchez:
Hi @sinusiks ,

I think you could test how things look to you by adjusting the fuzziness and also look into Elasticsearch synonyms for the specific Email vs E-mail issue. Perhaps this link can help: https://www.elastic.co/guide/en/elasticsearch/reference/current/analysis-synonym-tokenfilter.html

sinusiks:
Thanx but can you please tell me how to adjust fuzziness and synonyms? I use default Elasticpress with default Elasticsearch installation - can I solve both in the code or in the elasticsearch configuration? Because I can't find this anywhere through the module settings..

oscarssanchez:
Hi @sinusiks ,

Yes you would need to add this through code. 

Here is a sample on how to set a synonym:

```
add_filter( 'ep_config_mapping', 'search_config_mapping' );

function search_config_mapping( $mapping ) {
// Ensure we have filters and is array.
if ( ! isset( $mapping['settings']['analysis']['filter'] ) || ! is_array( $mapping['settings']['analysis']['filter'] ) ) {
return $mapping;
}
// Ensure we have analyzers and is array.
if ( ! isset( $mapping['settings']['analysis']['analyzer']['default']['filter'] ) || ! is_array( $mapping['settings']['analysis']['analyzer']['default']['filter'] ) ) {
return $mapping;
}
// Define the custom filter.
$mapping['settings']['analysis']['filter']['our_synonym_filter'] = array(
'type' => 'synonym',
'lenient' => true,
'synonyms' => array( 'Look => See' ),
);
// Tell the analyzer to use our newly created filter.
$mapping['settings']['analysis']['analyzer']['default']['filter'] = array_merge( [ 'our_synonym_filter' ], $mapping['settings']['analysis']['analyzer']['default']['filter'] );
return $mapping;
}
```

[...]

Issue #218, opened 2015-03-02
Custom Analyzer


GhostToast:
This is definitely a "would be nice" item:

To be able to associate relevant terms from within WP dashboard. What I mean is terms which have synonymous meaning within the context of a particular site and would help users find content based on the potential intention of their words.

As an example, let's say there is a restaurant review site which writes about a bar in Chicago. Only, they insist on referring to it as being in "Chi-Town" or "The Windy City" or "Second City" and never actually call Chicago by the real name. A user wants to find reviews for "Chicago Bars" and no results are found. I know this is an edge case, but one can find other uses, such as abbreviations. (New York City: "NYC").

Would be amazing to someday plug in these pieces of info directly into a WP dashboard. I may investigate doing this manually with a custom analyzer for an upcoming project which deals with medical issues (and their common abbreviations).

GhostToast:
Related reading: http://www.elasticsearch.org/guide/en/elasticsearch/guide/current/custom-analyzers.html

tlovett1:
Definitely an idea. Please post the code you use to add the custom analyzer. Thanks!

GhostToast:
More related reading: http://www.elasticsearch.org/guide/en/elasticsearch/reference/current/analysis-synonym-tokenfilter.html

GhostToast:
Some basic synonym stuff:

```<?php

namespace SECRET\CLIENT\EP;

/**
 * Custom mapping for ElasticPress
 *
 * @param $mapping
 *
 * @return mixed
 */
function elasticpress_config_mapping( $mapping ) {
	if ( ! isset( $mapping ) || ! is_array( $mapping ) ) {
		return false;
	}

	if ( ! isset( $mapping['settings']['analysis']['filter'] ) || ! is_array( $mapping['settings']['analysis']['filter'] ) ) {
		return false;
	}

	if ( ! isset( $mapping['settings']['analysis']['analyzer']['default']['filter'] ) || ! is_array( $mapping['settings']['analysis']['analyzer']['default']['filter'] ) ) {
		return false;
	}

	$mapping['settings']['analysis']['filter']['secret_client_synonym_filter'] = array(
		'type' => 'synonym',
		'synonyms' => array(
			'nyc,new york,new york city',
		),
	);

	$mapping['settings']['analysis']['analyzer']['default']['filter'][] = 'secret_client_synonym_filter';

	return $mapping;
}
[...]

PR #1824, opened 2020-07-02
Add a synonyms feature to support a user defined synonym filter

christianc1:
### Description of the Change

**Work in Progress**

A EP Core Feature to add synonym support to ElasticPress.  Synonyms are stored as post content of a post in a private post type, though the edit UI is decoupled from the standard post edit screen.  Synonym lists follow the Solr spec defined [here](https://www.elastic.co/guide/en/elasticsearch/reference/current/analysis-synonym-tokenfilter.html#_solr_synonyms).
 
### Alternate Designs

Needs UX/Visual Design.

### Benefits

Users can define synonyms or alternate spellings for their most popular search terms from the UI.

### Possible Drawbacks

Large synonym filters can affect elasticsearch query performance.

### Verification Process

Unverified.  Work in progress, looking for 👍 's to proceed beyond proof of concept.

### Checklist:

### Applicable Issues

#218 #1791


includes/classes/Feature/Synonyms/Synonyms.php
```diff
@@ -0,0 +1,524 @@
...
+
+	/**
+	 * Initialize feature setting it's config
+	 *
+	 * @since  3.4
+	 */
+	public function __construct() {
+		$this->slug                     = 'synonyms';
+		$this->title                    = esc_html__( 'Synonyms', 'elasticpress' );
+		$this->requires_install_reindex = false;
```
felipeelia, line 45: As we are changing the mapping, shouldn't this be true? In my tests locally the feature just works after reindexing.

includes/classes/Feature/Synonyms/Synonyms.php
```diff
@@ -0,0 +1,524 @@
...
+	 */
+	public function register_post_type() {
+		$args = [
+			'description'        => esc_html__( 'Elasticsearch Synonyms', 'elasticpress' ),
+			'public'             => false,
+			'publicly_queryable' => false,
+			'show_ui'            => false,
+			'show_in_menu'       => false,
+			'query_var'          => true,
+			'rewrite'            => [ 'slug' => 'ep-synonyms' ],
```
felipeelia, line 299: Sorry if I'm missing something obvious but do we need rewrite for this CPT?

includes/classes/Feature/Synonyms/Synonyms.php
```diff
@@ -0,0 +1,524 @@
...
+	 * @return void
+	 */
+	public function output_feature_box_long() {
+		?>
+		<p><?php esc_html_e( 'Create a custom synonym filter to allow ElasticPress to match alternative spellings or synonyms of your most popular search terms.', 'elasticpress' ); ?></p>
+		<?php
+	}
+
+	/**
+	 * Adds search synonyms.
```
felipeelia, line 112: Just to keep consistency, this should be "Add search synonyms.", right?

includes/classes/Feature/Synonyms/Synonyms.php
christianc1: I'm not sure, and it's one of the things I think is worthy of a conversation.  Upon activating the feature, I add an annotated synonym file which is only really helpful as a starting point to help users understand the syntax for the file.  (It's the same as the example as the Elasticsearch and Solr docs.

Re-indexing on install would actually put this initial mapping to ES, which I'm not sure is the right behavior.  There is the case where users deactivate then re-activate the feature after adding their own synonyms.  In that case, it makes sense to put that mapping on reactivation.

In an ideal world, I think we put an updated mapping to ES when a user saves the synonym file, or provide the reindex button on the header bar on the synonym page.

includes/classes/Feature/Synonyms/Synonyms.php
christianc1: updated

includes/classes/Feature/Synonyms/Synonyms.php
christianc1: removed in 4c73c9c

christianc1:
Update admin UI per @helen's draft.

[](url)

christianc1:
@helen This is ready for any frontend tweaks.  Additionally mentioned elsewhere, one of the editor mode toggle buttons should be removed, whichever you want.  Editor mode preference is saved in the feature settings so it persists (globally).  We could save to user meta but I think that's a bit overkill.

PR #941, opened 2017-09-29
#781 - Search by product variations SKU

lukaspawlik:
@tlovett1 This is a proof of concept of handling searching by products variations SKU and returning parent product (https://github.com/10up/ElasticPress/issues/781).

This is quite experimental approach as it seems that the easiest way of achieving this is creating artificial meta with SKUs of _product_variation_ children posts. Values of this artificial meta is calculated during posts indexing and it supports scenarios when variation product is edited/added/removed.

If you don't like using _meta_ for this then how about creating extra post property for storing this kind of "dynamically calculated static data" separately from metas?

Issue #781, opened 2017-03-22
WooCommerce feature variation search on Product edit.php


kallehauge:
Looking at `WC_Admin_Post_types::product_search` then they make a direct SQL Query to fetch product and variable product ids where `_sku` is a match (among others) and search replace the `WHERE` condition used to include variations.

It is currently not possible to do the same with ElasticPress (unless I have completely failed) which means that the functionality isn't 1:1.

_Sorry if this have been documented somewhere and I've missed it_

```
/**
	 * Search by SKU or ID for products.
	 *
	 * @param string $where
	 * @return string
	 */
	public function product_search( $where ) {
		global $pagenow, $wpdb, $wp;

		if ( 'edit.php' != $pagenow || ! is_search() || ! isset( $wp->query_vars['s'] ) || 'product' != $wp->query_vars['post_type'] ) {
			return $where;
		}

		$search_ids = array();
		$terms      = explode( ',', $wp->query_vars['s'] );

		foreach ( $terms as $term ) {
			if ( is_numeric( $term ) ) {
				$search_ids[] = $term;
			}

			// Attempt to get a SKU
			$sku_to_id = $wpdb->get_results( $wpdb->prepare( "SELECT ID, post_parent FROM {$wpdb->posts} LEFT JOIN {$wpdb->postmeta} ON {$wpdb->posts}.ID = {$wpdb->postmeta}.post_id WHERE meta_key='_sku' AND meta_value LIKE %s;", '%' . $wpdb->esc_like( wc_clean( $term ) ) . '%' ) );
			$sku_to_id = array_merge( wp_list_pluck( $sku_to_id, 'ID' ), wp_list_pluck( $sku_to_id, 'post_parent' ) );

			if ( sizeof( $sku_to_id ) > 0 ) {
				$search_ids = array_merge( $search_ids, $sku_to_id );
			}
		}

		$search_ids = array_filter( array_unique( array_map( 'absint', $search_ids ) ) );

		if ( sizeof( $search_ids ) > 0 ) {
			$where = str_replace( 'AND (((', "AND ( ({$wpdb->posts}.ID IN (" . implode( ',', $search_ids ) . ")) OR ((", $where );
		}

		return $where;
	}
```

tlovett1:
You can definitely look up products by SKU with EP/ES. Not sure I follow the question.

kallehauge:
[...]

PR #1828, opened 2020-07-06
Allow searches for variation SKU in admin interface

kallehauge:
### Description of the Change

This is a proof of concept of handling searching by products variations SKU and returning parent product (#781) and is a refactored version of https://github.com/10up/ElasticPress/pull/941 since it got closed due to major code changes in 3.0.

The point of this is to be able to search for a variation sku and then get the parent product in the admin interface (this is a default feature in WooCommerce).

### Verification Process

I created a new variable product with a couple of variants. These all have a new `__wc_ep_variations_skus` meta value in Elasticsearch.
The value of `__wc_ep_variations_skus` is being updated when the parent product is updating which is the core logic of the PR but the value have also been tested to update when a variation is being deleted through the admin interface without pressing the "Update" button for the parent product and the value have also been tested to update when using the variation _Save Changes_ button below variations in the interface (hence; not the "Update parent post button")

I have also tested that the admin product overview list changes only affect the mentioned list by test other overview lists (like Orders, Pages, Posts) and on the front-end when Elasticsearch is enabled.

### Checklist:

- [ ] I have read the [**CONTRIBUTING**](/CONTRIBUTING.md) document. **This file no longer exists**

### Applicable Issues

https://github.com/10up/ElasticPress/issues/781

### Changelog Entry

WC Enhancement: add variation sku as a searchable field and apply it on the product overview list.

### DIFF
--- a/includes/classes/Feature/WooCommerce/WooCommerce.php
+++ b/includes/classes/Feature/WooCommerce/WooCommerce.php
@@ -705,7 +705,7 @@ class WooCommerce extends Feature {
 	 * @since TBD
 	 *
 	 * @param array $post_args Post arguments.
-	 * @param int $post_id Post ID.
+	 * @param int   $post_id Post ID.
 	 *
 	 * @return array
 	 */
@@ -1008,7 +1008,7 @@ class WooCommerce extends Feature {
 	 * @since TBD
 	 *
 	 * @param array $post_args Post arguments.
-	 * @param int $post_id Post ID.
+	 * @param int   $post_id Post ID.
 	 *
 	 * @return array
 	 */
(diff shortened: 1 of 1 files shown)


kallehauge:
Just a random ping in the hopes of some sort of response :-) @dinhtungdu @jeffpaul 

Is the initial approach still valid in your opinion? I'm currently running with this PR on a couple of live sites

tlovett1:
@oscarssanchez can you test this one?

dinhtungdu:
Hi @kallehauge, I'm sorry for this late reply! Here is the result of my test.

I tested your PR with WP 5.5.1 and WC 4.5.2. The Protected Content feature is enabled and posts are reindexed before testing.

The product search "works", but there is a problem: the formatted args is not correct, it doesn't include `s` but `post__in`. This means that the query is modified before coming to the `format_args` function.

Trace down the code I realize that WC uses a method named `search_products` that WC uses to search for products in the products admin list table, and if this method finds anything, `s` in query args will be unset and replace by `post__in`. This means that EP doesn't really search for products.

Can you confirm this behavior @kallehauge?

cc @tlovett1.

kallehauge:
@dinhtungdu I don't really have time to verify it for the foreseeable future. All of my resources (both at work and in private) is allocated elsewhere by now.
The PR is basically just a rewrite of the [original PR by lukaspawlik](https://github.com/10up/ElasticPress/pull/941) so I think there - in general - might be a few issues since it was a "pretty quick port" and I wanted to know if it was still the same direction you still wished to go before I dedicated more time. But I hope it can be a good starting point for someone else

felipeelia:
Closed by #2763 and #2757.

PR #1829, opened 2020-07-07
Type check $feature before using method

pdewouters:
### Description of the Change
The `get_registered_feature` method can return boolean `false`, so in that case, calling `is_active()` on this value will cause an error:

```Uncaught Error: Call to a member function is_active() on boolean in /usr/src/app/vendor/10up/elasticpress/includes/classes/AdminNotices.php on line 84
process_using_autosuggest_defaults_notice()
call_user_func()
vendor/10up/elasticpress/includes/classes/AdminNotices.php:63
process_notices()
vendor/10up/elasticpress/includes/dashboard.php:302
ElasticPress\D\maybe_notice()
wp-includes/class-wp-hook.php:287
apply_filters()
wp-includes/class-wp-hook.php:311
do_action()
wp-includes/plugin.php:478
do_action()
wp-admin/admin-header.php:267
require_once()
wp-admin/network/index.php:65
```

### Benefits

It will fix a bug 

### Possible Drawbacks

none

### Verification Process
There is no error message at `wp-admin/network/`

### Checklist:

### DIFF
--- a/includes/classes/AdminNotices.php
+++ b/includes/classes/AdminNotices.php
@@ -80,7 +80,10 @@ class AdminNotices {
 	 */
 	protected function process_using_autosuggest_defaults_notice() {
 		$feature = Features::factory()->get_registered_feature( 'autosuggest' );
-
+		if ( ! $feature instanceof Feature ) {
+			return false;
+		}
+		
 		if ( ! $feature->is_active() ) {
 			return false;
 		}


tlovett1:
Interesting. Not sure why that wouldn't be registered, but thank you!

pdewouters:
@tlovett1 to be fair, it's a bit of an edge case, caused by loading dependencies in the wrong order on my side.

PR #1835, opened 2020-07-14
Add a separate sync page

helen:
### Description of the Change

Adds a separate sync screen that explains the different syncing options.

This is not yet complete - it needs to disable the syncing buttons while a sync is running and to actually trigger the appropriate sync from each button. I am not currently planning on handling this, so somebody else will need to take over this PR to complete it :)

Screen Shot 2020-07-14 at 4 55 05 PM

### Alternate Designs

n/a

### Benefits

Two separate sync paths, no more one-click access to a destructive syncing action without warning or explanation.

### Possible Drawbacks

More clicks to sync again.

### Verification Process

Clicked sync button in header.

### Checklist:

### Applicable Issues

Not sure if there is an issue or a PR for the actual non-destructive reindexing part.

### Changelog Entry

Still needs


includes/classes/Indexable/Post/Post.php
```diff
@@ -88,7 +88,7 @@ public function query_db( $args ) {
 		 */
 		$args = apply_filters( 'ep_index_posts_args', apply_filters( 'ep_post_query_db_args', wp_parse_args( $args, $defaults ) ) );
 
-		if ( isset( $args['include'] ) || isset( $args['post__in'] ) || 0 < $args['offset'] ) {
+		if ( isset( $args['post__in'] ) || 0 < $args['offset'] ) {
```
tott, line 91: Why did we drop include?

tott: nevermind - i see it here 
if ( isset( $args['include'] ) ) {...}

tott (approved):
Looks solid to me. we need to confirm which methods might be used for integrations and make sure people are not querying any transients or options directly and make sure the methods people use will still work.

includes/classes/Screen/Sync.php
```diff
@@ -0,0 +1,207 @@
...
+	public function action_wp_ajax_ep_cancel_index() {
+		if ( ! check_ajax_referer( 'ep_dashboard_nonce', 'nonce', false ) || ! EP_DASHBOARD_SYNC ) {
+			wp_send_json_error();
+			exit;
+		}
+
+		$index_meta = Utils\get_indexing_status();
+
+		if ( isset( $index_meta['method'] ) && 'cli' === $index_meta['method'] ) {
+			set_transient( 'ep_wpcli_sync_interrupted', true, 5 );
```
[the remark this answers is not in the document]
felipeelia: You are totally right @tott. I've bumped it to one minute.

PR #1880, opened 2020-09-01
Return status early if elasticsearch not available

oscarssanchez:
### Description of the Change

Hi @tlovett1 ,

Could you please review this PR? 

This adds a check to see if elastisearch request returns the ES version, and if it doesn't, skip the next check (ES plugins). It looks like if your ES goes off for any reason (the case that brought this was a timeout from indexing documents) your documents feature will deactivate automatically. 

I think the check should only happen if ES is online and then check if we indeed have the ingest attachment plugin. This seems to solve the issue for the client who brought this to our attention as well. 

### Checklist:

### DIFF
--- a/includes/classes/Feature/Documents/Documents.php
+++ b/includes/classes/Feature/Documents/Documents.php
@@ -308,6 +308,10 @@ class Documents extends Feature {
 	public function requirements_status() {
 		$status = new FeatureRequirementsStatus( 1 );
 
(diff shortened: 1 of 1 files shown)


go-techreative:
Hi there, I am the client in question. This change would be much appreciated. I have tested this solution for over a month now and the issue with automatic deactivation no longer came up for my 3 instances of a website. Thank you @oscarssanchez

tlovett1:
Looks good to me.

PR #1884, opened 2020-09-08
Fixes current price filter query unsetting old query 

oscarssanchez:
### Description of the Change

Hi @tlovett1 ,

Can you please review this PR? It looks like by unsetting `$args['query']['bool']['should']` when we want to reassign the search params later here:

`$args['query']['bool']['must'][1]['bool']                               = $args['query']['bool'];`

`$args['query']['bool']` would be equal to the content of this array: 

`$args['query']['bool']['must']`

But what we want to do is keep the search params intact so that we can search by price, but based on a given search term. We had a client who needed the price filtering and this solves their issue.

### Checklist:

### DIFF
--- a/includes/classes/Feature/WooCommerce/WooCommerce.php
+++ b/includes/classes/Feature/WooCommerce/WooCommerce.php
@@ -840,6 +840,7 @@ class WooCommerce extends Feature {
 			/**
 			 * This logic is iffy but the WC price filter widget is not intended for use with search anyway
 			 */
+			$old_query = $args['query']['bool'];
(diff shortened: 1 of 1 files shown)


tlovett1:
How do I reproduce this bug?

Rahmon:
Steps I took to reproduce this behavior:

1. Add Product Search and Filter Products by Price widgets 
2. Search for 'shirt'
3. Change the price filter to min: 20 and max: 20.
4. Change the price filter to min:10

After that, the result of the search includes all products and not just t-shirts.

ep-pr-1884

Issue #1854, opened 2020-08-01
Error: The element input must not appear as a descendant of the a element.


MediaMaquina:
**Describe the bug**

The facets links are not hmtl5 valid.
**Steps to Reproduce**

1. Have facets
2. validate page with https://validator.w3.org/
3. Scroll down to 'Error: The element input must not appear as a descendant of the a element'

```
<a href="/movie/?filter_genre=comedy" rel="nofollow">
    <input type="checkbox">Comedy
</a>
```

**Expected behavior**

it should use valid html.

**Screenshots**

**Environment information**
 - Device: 
 - OS: 
 - Browser and version: 
 - WordPress version: 
 
 - Plugins and version: 
 - Theme and version: 
 
 - <details><summary>Site Health Info:</summary>
   
   </details>

**Additional context**

MediaMaquina:
Also another error with facets:

**Error: Element a is missing required attribute href.**

This happens when we have 
`<a  rel="nofollow">`
On terms that are disabled (no posts with that term). a lacks href.

amesplant:
### Additional A11y Issues to address with this ticket 
**1.3.1**

- This form field should be labelled in some way. Use the label element (either with a "for" attribute or wrapped around the form field), or "title", "aria-label" or "aria-labelledby" attributes as appropriate.
facet checkbox: 
`<input type=""checkbox"" checked="""">`

**1.4.3**

- when a facet term is 'inactive' the minimum color contrast fails `.empty-term`

**1.4.11**

- when a facet term is 'inactive' the minimum color contrast fails `.empty-term`

**2.1.1**

- while tabbing through, the checkbox appears to be more of a visual indicator and not a funcitonal one, through the use of the keyboard, when a user tabs through and checks a checkbox, nothing happens as there is not submit button
facet checkbox:
`<input type=""checkbox"" checked="""">`

**2.5.3**

- Text in label element should provides name for input via for attribute. Consider markup to take advantage of labels:
```
<input id=""cat-1"" type=""checkbox"">
    	<label for=""cat-1"">Meal</label>
```

**3.2.2**

- Form does not contain a submit button
- No notification to user what will happen if checkbox is ""checked""

**3.3.2**

- Labels or instructions are not provided when content requires user input.

**4.1.2**

- This checkboxinput element does not have a name available to an accessibility API. Valid names are: label element, title undefined, aria-label undefined, aria-labelledby undefined.
facet checkbox:
`<input type=""checkbox"" checked="""">`

[...]

PR #1886, opened 2020-09-10
Fix/1854

amesplant:
### Description of the Change

Resolves invalid HTML by:
 - Replacing invalid `<input>` checkboxes in facet links, and replace with presentational divs styled as checkboxes.
 - No longer adding the `rel="nofollow"` attribute to links without an `href` (i.e. empty terms).

### Alternate Designs

One alternative option would be to remove the links and implement facets as a `<form>` element. Downsides of that approach are that it would either add the requirement to submit facets manually after selection, or the page would need to refresh upon selection, which would be unexpected behaviour for a form. Additionally, manual submission would require back-end changes to accommodate.

### Benefits

Resolves invalid HTML of the current facets, and a custom design for the checkboxes would not imply standard form behaviour, which selecting an option does not follow.

### Possible Drawbacks

Any custom styles site owners or theme authors may have applied to checkboxes, either generally or just for ElasticPress, would not apply to the new checkboxes.

### Verification Process

Running a page with ElasticPress facets through the W3 Validator should not return any errors for facets. Make sure to check the HTML when some empty terms are in the list.

### Checklist:

### Applicable Issues

#1854

### Changelog Entry

Fixed invalid HTML for facets by replacing the `<input type="checkbox">` elements in option links with a custom `.ep-checkbox` presentational div.


oscarssanchez:
Do we want to assign someone and finish the PR per Ames comment here https://github.com/10up/ElasticPress/issues/1854#issuecomment-692812901 ?

JakePT:
As Ames mentioned on #1854, this will need back-end engineering to support the noscript implementation, so it probably can't be merged yet. 

There's another direction we could go, however. We could just use `<a>` tags only, without the `<form>` or checkbox 
`<inputs>`, as this would satisfy the original issue of invalid markup, but wouldn't require any back-end engineering to implement, as the links would work as they did before. Since the behaviour of selecting a facet is to direct the user to a new results page, rather than load results in with AJAX, I'd argue an anchor tag is probably the more appropriate tags to use accessibility-wise too.

```
<a href="/movie/?filter_genre=comedy" rel="nofollow">
  <span class="checkbox">Comedy</span>
</a>
```

The only missing piece would be that the links wouldn't look like checkboxes anymore, which would require some styling, but this PR includes custom styling for checkbox inputs which could be adapted to links fairly easy. 

Alternatively, we could keep the checkbox, but move it outside the anchor tag, and mark it up as presentation to avoid accessibility issues:

```
<input type="checkbox" role="presentation">
<a href="/movie/?filter_genre=comedy" rel="nofollow">
  Comedy
</a>
```

This approach would probably have the least impact on the front end design of any existing sites too.

mckdemps:
@JakePT Yes, we like this approach. And what does it look like to replace the checkbox?

JakePT:
@mckdemps It just means that the checkboxes would look like however we define them, rather than using the styles from the theme, or the default system/browser checkboxes. This is how Ames' checkboxes look in a bare-bones theme:

Screen Shot 2021-03-08 at 2 08 44 pm

Whereas the current checkboxes, if they aren't styled by the theme, would look like the Checklist checkboxes in the PR description above.

This also means that any custom styling users might've applied to facets will not apply to the new checkboxes.

An alternative approach would be to include an actual checkbox, but marked up as `readonly` and `role="presentation"`, so that they are there purely for the visuals. This means that the appearance of the checkboxes would not change, but the spacing around them would, as we would need to 'cover' the checkboxes with the link, so that they're clickable.

I prefer the custom checkbox styling approach, as the results will be more consistent between themes, users will be able to more easily style them independently of other checkboxes in their theme, and they would not give the false impression that they will behave the same as system/browser checkboxes.

brandwaffle:
@JakePT I'm good with the custom checkboxes as well. Let's go with that approach.

JakePT:
@mckdemps @brandwaffle I have updated the PR to use links rather than forms, and I've tweaked the styling of the custom checkboxes to use `em` units so that they scale appropriately to the theme's font size, and I used flexbox for aligning the checkboxes for more consistent results across themes and font sizes/line heights.

oscarssanchez:
Hi @JakePT code looks good and it works correctly. There's a minor space missing from what I can tell on widget.php https://github.com/10up/ElasticPress/pull/1886/checks?check_run_id=2063020128 

Can you please take care of it and merge master into this branch? It seems some tests failed on EPIO but I wonder if they would go away once the branch is updated to the latest.

JakePT:
@oscarssanchez I've merged develop into this branch and fixed the coding standards errors. I'm not sure what the EP.io test you're referring to is so I'll let you have a look at that when the tests are finished.

Issue #1604, opened 2019-12-17
Image block and post_mime_type


ethanclevenger91:
**Describe the bug**
The Gutenberg image block adds `'post_mime_type' => 'image'` to the media modal's search queries (the `query-attachments` AJAX action). The classic editor didn't add this parameter. 

ElasticPress translates post_mime_type to an exact match, but in `wp_post_mime_type_where()`, which WP_Query uses to parse that argument, if there's no `/` in the post_mime_type, the resulting SQL wildcards the second half of the mime type, so `LIKE 'image/%'` (for this specific case).

While admin-side functionality isn't the default for this plugin, updating this logic would make it more seamless.

**Steps to Reproduce**
1. Install/configure ElasticPress, Classic Editor plugin
2. Enable document indexing
3. Enable admin and AJAX via `ep_admin_wp_query_integration` and `ep_ajax_wp_query_integration`
4. Add an image to the media library
5. Create a new post
6. Edit with classic editor
7. Use the "Add Media" button to open the media modal
8. See your image
9. Use the search bar to search for your image
10. Still see your image
11. Close out. Edit a post with the block editor.
12. Add a new image block to bring up the media modal
13. See your image
14. Use the search bar to search for your image
15. See no results returned

**Expected behavior**
Image search results in the admin should be consistent.

**Environment information**
 - Device: ThinkPad
 - OS: Windows
 - Browser and version: Firefox
 - Plugins and version: ElasticPress, latest
 - Theme and version: Custom
 - Other installed plugin(s) and version(s): I can provide this if the above steps fail to reproduce

**Additional context**
Current workaround is a `pre_get_posts` filter:

```
public function query_attachments( &$query ) {
	if( !wp_doing_ajax() || $_POST['action'] != 'query-attachments' ) {
		return;
	}

	$query->set('post_mime_type', '');
}
```

brandwaffle:
@ethanclevenger91 agreed this would be a good thing to get working. Currently, the only way I'm aware of to create a partial match in Elasticsearch is to map this particular field with the ngram analyzer, which enables partial string matches. That could create some indexing overhead, so I'm also wondering if perhaps we could shortcut this by splitting the 'image/<format>' string into an array of each part, so that we could exact match on just 'image' across any images regardless of format. @tlovett1 any thoughts?

tlovett1:
This is admin search correct @ethanclevenger91? I assume you have the Protected Content feature enabled?

ethanclevenger91:
[...]

PR #1887, opened 2020-09-10
Improve mime type filtering

roborourke:
### Description of the Change

As noted in issue #1604 the image block sends a mime type value of `image` through as an array which goes against the convention this part of the query filter relied on. This update makes the filter more agnostic, working with a single string (as `wp_post_mime_type_where()` does) or an array, plus it will then separate the values into full terms and prefixes.

### Benefits

This has the benefit of supporting almost any combination of values given in the `post_mime_type` argument and avoids the use of a `regexp` query which can be slower to run.

This fixes issue #1604

### Verification Process

Searching media in the media modal opened from an image block in the block editor now returns results.

### Checklist:

### Applicable Issues

- #1604 

### Changelog Entry

- Changed `post_mime_type` filtering to support more edge cases

### DIFF
--- a/includes/classes/Indexable/Post/Post.php
+++ b/includes/classes/Indexable/Post/Post.php
@@ -907,25 +907,61 @@ class Post extends Indexable {
 		/**
 		 * Add support for post_mime_type
 		 *
-		 * If we have array, it will be fool text search filter.
-		 * If we have string(like filter images in media screen), we will have mime type "image" so need to check it as
-		 * regexp filter.
+		 * Values can be in multiple forms such as "image", "image/*" or "image/jpeg". A single string
+		 * or an array of mime types can be passed 
 		 *
 		 * @since 2.3
 		 */
 		if ( ! empty( $args['post_mime_type'] ) ) {
-			if ( is_array( $args['post_mime_type'] ) ) {
-				$filter['bool']['must'][] = array(
+			if ( is_string( $args['post_mime_type'] ) ) {
+				$mime_types = array_map( 'trim', explode( ',', $args['post_mime_type'] ) );
+			} else {
+				$mime_types = (array) $args['post_mime_type'];
+			}
+			
+			// Collect fully qualified mime types here e.g. image/jpeg.
+			$terms = array();
+			// Collect prefix types here e.g. image, image/*.
+			$prefixes = array();
+
+			foreach ( $mime_types as $type ) {
+				// Remove trailing slashes and wildcards.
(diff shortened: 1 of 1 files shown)


roborourke:
I'll look into updating the tests too, just wanted to get this up.

roborourke:
Ah seems the test suite is a bit broken right now.

includes/classes/Indexable/Post/Post.php
```diff
@@ -907,25 +907,61 @@ function( $tax_query ) use ( $args ) {
...
+		 * or an array of mime types can be passed 
 		 *
 		 * @since 2.3
 		 */
 		if ( ! empty( $args['post_mime_type'] ) ) {
-			if ( is_array( $args['post_mime_type'] ) ) {
-				$filter['bool']['must'][] = array(
+			if ( is_string( $args['post_mime_type'] ) ) {
+				$mime_types = array_map( 'trim', explode( ',', $args['post_mime_type'] ) );
+			} else {
```
tlovett1, line 918: This doesn't make total sense to me. I think it would be better to check if it's a string, then check if it's an array. If it's anything else, ignore the parameter.

includes/classes/Indexable/Post/Post.php
```diff
@@ -907,25 +907,61 @@ function( $tax_query ) use ( $args ) {
...
 				);
+			}
 
-				$use_filters = true;
-			} elseif ( is_string( $args['post_mime_type'] ) ) {
+			if ( ! empty( $prefixes ) ) {
+				foreach ( $prefixes as $prefix ) {
+					$mime_type_filter[] = array(
+						'prefix' => array(
+							'post_mime_type' => $prefix,
```
tlovett1, line 954: Is mime type supposed to support wildcards?

tlovett1 (changes requested):

includes/classes/Indexable/Post/Post.php
roborourke: Yeah but only in a limited way - this function is used to build the SQL in WP core https://developer.wordpress.org/reference/functions/wp_post_mime_type_where/

You can query by subgroup eg. `image` or by fully qualified mime types like `image/jpeg`

includes/classes/Indexable/Post/Post.php
roborourke: Yep, good shout 👍

felipeelia:
Closing this as we've merged #2222 to fix the same issue.

Issue #1857, opened 2020-08-05
Every edit page view triggers ES sync


adiloztaser:
**Describe the bug**

`_edit_lock` meta is updated each time visiting post edit page and then ElasticPress triggers ES sync. It seems like an unnecessary request.

**Environment information**
 - WordPress version: 5.3.4
- ElasticPress: 3.4.1

PR #1899, opened 2020-09-23
Post metadata: Just queue post if the meta changed is allowed

felipeelia:
### Description of the Change

As described in #1857, EP will trigger a sync every time any metadata is changed (including `_edit_lock` and any other). This PR aims to only sync when metadata that will really be stored in ES is changed.

To narrow down the allowed meta I'm using the `ElasticPress\Indexable\Post\Post::prepare_meta()` method, as it'll run all filters a dev would use: ep_prepare_meta_data, ep_prepared_post_meta, ep_prepare_meta_allowed_protected_keys, ep_prepare_meta_excluded_public_keys. ep_prepare_meta_whitelist_key, and ep_prepared_post_meta.

### Alternate Designs

We could avoid a ES sync just for `_edit_lock`.

### Benefits

Less requests made to the ES server.

### Possible Drawbacks

n/a

### Verification Process

I've looked at docker logs and made some `error_log()` calls. I've changed a default meta data (`my_test`) and also WooCommerce SKUs. Both were updated successfully.

### Checklist:

### Applicable Issues

#1857

### Changelog Entry

- Fixed unnecessary requests to the ES server while saving posts.

### DIFF
--- a/includes/classes/Indexable/Post/SyncManager.php
+++ b/includes/classes/Indexable/Post/SyncManager.php
@@ -76,6 +76,11 @@ class SyncManager extends SyncManagerAbstract {
 
 		$post = get_post( $object_id );
 
+		$allowed_meta_to_be_indexed = $indexable->prepare_meta( $post );
+		if ( ! in_array( $meta_key, array_keys( $allowed_meta_to_be_indexed ), true ) ) {
+			return;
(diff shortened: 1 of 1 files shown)


tlovett1:
Nice work!

PR #1927, opened 2020-10-22
Fix autosuggest tag and class

felipeelia:
### Description of the Change

1. Only highlights terms when Highlighting is enabled
2. Apply the correct class and the selected tag to autosuggest highlights
3. Adjust the regex so partial matches will be highlighted


felipeelia:
Closing this in favor of #1928 (0f8a044 has all the changes we had here).

PR #1936, opened 2020-10-26
Feature activation: Alert when a new reindex needs to happen

### LINKED ISSUE #1932: notify of required resync while editing feature settings
This morning, I enabled the new "Protected Content" option, and then a complete resync began, which immediately killed my entire previous sync.  It then took over an hour to resync, as our site has hundreds of thousands of posts in it.  This is why we only perform a sync in the early morning during the lowest traffic hours.  

So obviously, if I had known that my setting change required a complete resync, I would not have done so at that time of day.  It be so helpful if there was a warning that a resync will be required before saving a change.  For example, when you select the `Enable` radio button under "Protected Content", a warning would appear next to the `Save` button that says "resync required" or something like that.

felipeelia:
### Description of the Change

Currently, users don't have much info that a new reindex will start when a certain feature is activated. This PR adds a notice based on the Feature `requires_install_reindex` attribute.

### Alternate Designs

Perhaps we could place the message somewhere else.

### Benefits

Users won't start a reindex without notice.

### Possible Drawbacks

n/a

### Verification Process

Created the code and checked in the Features Screen.

### Checklist:

### Applicable Issues

Closes #1932

### Changelog Entry

Adds a new notice to features that need a resync after activation.


includes/classes/Feature.php
```diff
@@ -288,7 +304,7 @@ public function output_settings_box() {
 				<?php esc_html_e( 'Setting adjustments to this feature require a re-sync. Use WP-CLI.', 'elasticpress' ); ?>
 			</span>
 
-			<a data-feature="<?php echo esc_attr( $this->slug ); ?>" class="<?php if ( 2 === $requirements_status->code || ( $this->requires_install_reindex && defined( 'EP_DASHBOARD_SYNC' ) && ! EP_DASHBOARD_SYNC ) ) : ?>disabled<?php endif; ?> button button-primary save-settings"><?php esc_html_e( 'Save', 'elasticpress' ); ?></a>
+			<a data-feature="<?php echo esc_attr( $this->slug ); ?>" class="<?php echo ( ! $can_save ) ? 'disabled' : ''; ?> button button-primary save-settings"><?php esc_html_e( 'Save', 'elasticpress' ); ?></a>
```
Rahmon, line 307: @felipeelia could you update to use `esc_attr` for `class`?

includes/classes/Feature.php
```diff
@@ -254,6 +254,22 @@ public function output_feature_box_settings() {
...
+		 *     - If the feature requires a reindex, it is really possible to reindex via Dashboard.
+		 */
+		$can_save = (
+			2 !== $requirements_status->code &&
+			( ! $this->requires_install_reindex || ( ! defined( 'EP_DASHBOARD_SYNC' ) || EP_DASHBOARD_SYNC ) )
+		);
+
+		if ( ! $this->is_active() && $this->requires_install_reindex && $can_save ) {
+			$requirements_status->message   = (array) $requirements_status->message;
+			$requirements_status->message[] = __( 'A new reindex will start when you activate this feature.', 'elasticpress' );
```
Rahmon, line 271: Just wondering, should we put this message at the top? 

image

includes/classes/Feature.php
felipeelia: @Rahmon to escape a fixed string?

includes/classes/Feature.php
felipeelia: That is a good point @Rahmon. I think all messages have the same importance, so IDK if we want to come up with a prioritization thing but we can bring that up during our internal sync.
Also, we need to change that `__()` call to use `esc_html__()`

felipeelia:
@JakePT I'm assigning this one to you as it can be used to some extent to fix the problem outlined in #1932. Also, this only can be worked when the new sync page is done, as activating a new feature that requires a reindex will now point the user to the new sync page. cc @Rahmon

felipeelia:
Closing this in favor of #2491

PR #1949, opened 2020-11-03
Feature: Dynamic index

dinhtungdu:
### Description of the Change
This PR adds a basic algorithm is to increase/decrease the number of posts/documents while indexing by timing each request. This also fixes the issue with very large documents (over 100MB) which exceed the `max_content_length` of Elasticsearch.

### Alternate Designs
n/a

### Benefits
A more reliable sync process that can self-tune base on the network and hardward condition at the sync time.

### Possible Drawbacks
None that I'm aware of.

### Verification Process
- Set bulk index setting to `1` and run sync. Check the setting value after the sync, see it changed.
- Upload some large documents then turn on document indexing. Run a sync, see the index succeeds.

### Checklist:


brandwaffle:
@dinhtungdu This is looking great! For release, we'd like to have the Settings menu display the current number of posts indexing per cycle using Dynamic Indexing. Then we'd like another field where someone can "override" the value and set it to a certain number per index cycle (i.e. what happens today). This is so someone can set a value in case there is a reason not to use dynamic indexing.

dinhtungdu:
@brandwaffle I add a checkbox to toggle the dynamic index feature.

When the box is checked. The value of the bulk setting will change dynamically. I decided to keep that behavior because:
- I don't want to add more data to the database.
- Site admins have a way to know the current bulk setting value set by the algorithm.

When the checkbox is unchecked (which is the default setting), ElasticPress will use the bulk setting as before.

For both scenarios, an improvement introduced in this PR is applied. EP will send only requests which are less than 100MB, and continue where it left off in the next index cycle. 100MB is is the default value of ES `[max_content_length](https://www.elastic.co/guide/en/elasticsearch/reference/master/modules-http.html#_http_settings)` and this is filterable by `ep_max_content_length`.

Rahmon:
Hi @dinhtungdu 

I started to review this PR and seems the dynamic index is working but when I tried to index a document I got an error. 

Also, I've updated the PR, could you check if everything is still working?

includes/partials/settings-page.php
```diff
@@ -227,6 +223,17 @@
...
 				</tr>
 			<?php endif; ?>
+			<?php if ( ! empty( $host ) ) : ?>
+				<tr>
+					<th scope="row">
+						<label for="ep_enable_dynamic_index"><?php esc_html_e( 'Enable dynamic index?', 'elasticpress' ); ?></label>
+					</th>
+					<td>
+						<input type="checkbox" name="ep_enable_dynamic_index" id="ep_enable_dynamic_index" value="1" <?php checked( $enable_dynamic_index, 1 ); ?>>
+						<legend class="description">Enabling this setting let ElasticPress adjust the bulk index setting automatically for a more reliable and optimized index process.</legend>
```
felipeelia, line 233: @dinhtungdu @Rahmon let's use an i18n function like `esc_html_e()` here please.

felipeelia (changes requested):

felipeelia:
@Rahmon @dinhtungdu not sure if this is something already under your radars, but before we merge this one, let's make sure to add all `@since` tags we need, okay? Thanks!

dinhtungdu:
@Rahmon Are you getting memory limit fatal error? If yes, please increase your PHP memory_limit setting. For large documents, we need a high limit because of the encoding process. I can index 100MB documents with memory_limit set to 512M.

dinhtungdu:
@felipeelia I added since tags and fix the i18n issue.

Rahmon (approved):

includes/dashboard.php
```diff
@@ -937,6 +952,14 @@ function action_admin_init() {
 				'sanitize_callback' => __NAMESPACE__ . '\sanitize_bulk_settings',
 			]
 		);
+		register_setting(
+			'elasticpress',
+			'ep_enable_dynamic_index',
+			[
+				'type'              => 'boolean',
+				'sanitize_callback' => __NAMESPACE__ . '\sanitize_enable_dynamic_index',
```
felipeelia, line 960: Any reason why we are not calling absint here directly?

includes/partials/settings-page.php
```diff
@@ -227,6 +223,17 @@
...
 				</tr>
 			<?php endif; ?>
+			<?php if ( ! empty( $host ) ) : ?>
+				<tr>
+					<th scope="row">
+						<label for="ep_enable_dynamic_index"><?php esc_html_e( 'Enable dynamic index?', 'elasticpress' ); ?></label>
+					</th>
+					<td>
+						<input type="checkbox" name="ep_enable_dynamic_index" id="ep_enable_dynamic_index" value="1" <?php checked( $enable_dynamic_index, 1 ); ?>>
+						<legend class="description"><?php esc_html_e( 'Enabling this setting let ElasticPress adjust the bulk index setting automatically for a more reliable and optimized index process.', 'elasticpress' ); ?></legend>
```
felipeelia, line 233: Looking at how it looks like on the Settings page, shouldn't this be a label instead of a legend? Also, if that is the case, let's add the proper "for" to it, for accessibility and ease of use.

includes/utils.php
```diff
@@ -467,6 +467,37 @@ function get_language() {
 	return apply_filters( 'ep_default_language', $ep_language );
 }
 
+/**
+ * Get EP option for both single and multisites.
+ *
+ * @since 3.6
+ *
+ * @param string $key Option key.
+ */
```
felipeelia, line 476: As we are returning things here, can we add the proper `@return` tag to the doc?

includes/utils.php
```diff
@@ -467,6 +467,37 @@ function get_language() {
...
+function get_ep_option( $key ) {
+	if ( defined( 'EP_IS_NETWORK' ) && EP_IS_NETWORK ) {
+		return get_site_option( $key );
+	} else {
```
felipeelia, line 480: We don't really need this else here, right?

includes/utils.php
```diff
@@ -467,6 +467,37 @@ function get_language() {
...
 }
 
+/**
+ * Get EP option for both single and multisites.
+ *
+ * @since 3.6
+ *
+ * @param string $key Option key.
+ */
+function get_ep_option( $key ) {
```
felipeelia, line 477: Can we add a `$default` parameter here to properly mimic WP functions?

includes/utils.php
```diff
@@ -467,6 +467,37 @@ function get_language() {
 	return apply_filters( 'ep_default_language', $ep_language );
 }
 
+/**
+ * Get EP option for both single and multisites.
+ *
+ * @since 3.6
```
felipeelia, line 473: Same about SemVer, let's go with `3.6.0`.

includes/utils.php
```diff
@@ -467,6 +467,37 @@ function get_language() {
...
+		return get_site_option( $key );
+	} else {
+		return get_option( $key );
+	}
+}
+
+/**
+ * Update EP option for both single and multisites.
+ *
+ * @since 3.6
```
felipeelia, line 488: Same

includes/utils.php
```diff
@@ -467,6 +467,37 @@ function get_language() {
...
+ *
+ * @param string $key   Option key.
+ * @param mixed  $value Option value to be set.
```
felipeelia, line 491: The `@return` tag is needed here as well.

includes/utils.php
```diff
@@ -467,6 +467,37 @@ function get_language() {
...
+ */
+function update_ep_option( $key, $value ) {
```
felipeelia, line 493: For this one we'll need to add `$autoload` (used in the update_option function.

includes/utils.php
```diff
@@ -467,6 +467,37 @@ function get_language() {
...
+	if ( defined( 'EP_IS_NETWORK' ) && EP_IS_NETWORK ) {
+		return update_site_option( $key, $value );
+	} else {
```
felipeelia, line 496: This else can go away too.

includes/utils.php
```diff
@@ -467,6 +467,37 @@ function get_language() {
...
+		return update_option( $key, $value );
+	}
+}
+
```
felipeelia, line 500: As we are on it, can we add a delete_ep_option() function as well?

felipeelia (changes requested):

includes/classes/Elasticsearch.php
```diff
@@ -1028,6 +1050,43 @@ public function bulk_index( $index, $type, $body ) {
...
 	}
 
+	/**
+	 * Try to tune bulk setting after timing the request.
+	 *
+	 * @since 3.6
+	 *
+	 * @param float $exec_time Execution time of the request.
+	 */
+	private function try_tuning_bulk_setting( $exec_time ) {
```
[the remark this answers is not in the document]
dinhtungdu: Ah, this is my mistake, changed to `protected`.

includes/dashboard.php
dinhtungdu: Oh, we should do that actually! Thanks for pointing this out!

includes/utils.php
dinhtungdu: I think we should only add it when we need it. We can add it later when we refactor existing `get_site_option` and `get_option` calls to EP utils.

includes/classes/Indexable.php
```diff
@@ -311,7 +321,26 @@ public function bulk_index( $object_ids ) {
...
+
+			/**
+			 * If the body exceeds the max content length allow, stop preparation.
+			 */
+			if ( ( strlen( $body ) + strlen( $document_str ) ) > $max_content_length ) {
+				set_transient( 'ep_current_queue_indexed_documents', $count );
+				break;
+			} else {
+				$count++;
+			}
```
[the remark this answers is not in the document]
dinhtungdu: Good catch!!! I fix this issue in https://github.com/10up/ElasticPress/commit/2f6a8dbf3d0f4f46123304680c1778a1c67db54b

dinhtungdu:
@felipeelia Thank you so much for the super detailed review! I addressed all problems you pointed out! Please give this one another look!

PR #1955, opened 2020-11-05
Implement ID ranges instead of relying on OFFSET when doing pagination.

brandon-m-skinner:
### Description of the Change

We noticed that indexing speed gets much slower than expected as the data set grows. We found the issue to be a combination of factors, but this PR is focused on an issue present in both versions of ElasticPress.

Essentially the change from using offsets to using ID ranges provides a very noticeable speed boost as the data set grows. We've seen significant gains when object IDs are extremely high.

Props @WPprodigy

### Alternate Designs

We didn't have any. Just needed some speed enhancements in short order.

### Benefits

It makes indexing faster at high object IDs.

### Possible Drawbacks

Adds additional CLI params.

### Verification Process

We ran the indexing commands on large multisites in excess of 10 million posts and noticed drastic increases in indexing speed compared to previous indexing operations.

Queries run without the PR(fetch 500 rows):
```
Offset 2500: 0.35 secs
Offset 25000: 1.54 secs
Offset 250000: 23.42 secs
Offset 2500000: 35.66 sec
```

Queries run with the PR are around 0.28 secs consistently regardless of the ID(fetch 500 rows)

### Checklist:

### Changelog Entry

Added performance boost to querying posts from database. Props @WPprodigy

### DIFF
--- a/includes/classes/Indexable/Post/Post.php
+++ b/includes/classes/Indexable/Post/Post.php
@@ -63,6 +63,8 @@ class Post extends Indexable {
 			'ignore_sticky_posts' => true,
 			'orderby'             => 'ID',
 			'order'               => 'desc',
+			'no_found_rows'       => true,
+			'ep_indexing_advanced_pagination' => true,
 		];
 
 		if ( isset( $args['per_page'] ) ) {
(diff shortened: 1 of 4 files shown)


includes/classes/Indexable/Post/Post.php
```diff
@@ -84,25 +86,104 @@ public function query_db( $args ) {
...
+
+			// Sanitize. Abort if unexpected data at this point.
+			if ( ! is_numeric( $start_range_post_id ) || ! is_numeric( $requested_end_id ) ) {
+				return $where;
+			}
+
+			$range = [
+				'start' => "{$GLOBALS['wpdb']->posts}.ID <= {$start_range_post_id}",
+				'end'   => "{$GLOBALS['wpdb']->posts}.ID >= {$requested_end_id}",
+			];
```
felipeelia, line 150: Hi @brandon-m-skinner ! Any reason why the upper limit is the start and the lower number is the end?

I wonder if it wouldn't be easier for users to run
`wp elasticpress index --start-object-id=1100 --end-object-id=1200`
rather than
`wp elasticpress index --start-object-id=1200 --end-object-id=1100`

includes/classes/Indexable/Post/Post.php
```diff
@@ -84,25 +86,104 @@ public function query_db( $args ) {
 		 * @param  {array} $args Database arguments
 		 * @return  {array} New arguments
 		 */
-		$args = apply_filters( 'ep_post_query_db_args', wp_parse_args( $args, $defaults ) );
+		$args = apply_filters( 'ep_index_posts_args', apply_filters( 'ep_post_query_db_args', wp_parse_args( $args, $defaults ) ) );
```
felipeelia, line 89: Can we split this into two lines, so we can add the necessary documentation for both filters @brandon-m-skinner ?

includes/classes/Indexable/Post/Post.php
WPprodigy: Because the indexing works from latest to oldest, this matches that behavior since it's starting with the highest ID and then goes to the lowest. I do agree that it's a bit confusing for users that may not be aware of this.

If it's a concern, I think perhaps could support both methods? Doesn't matter which way it's input, behind the scenes it could be organized as needed?

brandwaffle: @felipeelia @WPprodigy let's keep the ordering (I think it makes sense from an indexing perspective to start with the most recent content) but document this heavily because it's somewhat arbitrary, not intuitive.

PR #1956, opened 2020-11-05
Added action to edited_term to sync posts when a term is edited

brandon-m-skinner:
### Description of the Change

When a term is edited, all related posts should be re-indexed. This is very important when it comes to term-based queries which happen pretty regularly on our platform.

Filters for skipping this process have also been added since at large affected-post counts, things tend to fail. We handle these occurrences in a separate queue system that asynchronously processes indexing jobs.

Props @nickdaugherty
cc @rinatkhaziev @pschoffer @netsuso @parkcityj

### Alternate Designs

We didn't consider any other designs.

### Benefits

Post data remains consistent with the database which keeps queries returning expected results.

### Possible Drawbacks

The default behaviour is problematic when there are a lot of posts affected. We only do this behaviour for 10k or less posts and offload any indexing with counts above that to be processed asynchronously.

### Verification Process

We noticed data inconsistencies when comparing posts from the DB against posts from ES when editing terms. After applying these changes, the index and the DB were consistent.

### Checklist:

### Changelog Entry

Added queuing of affected posts on term change. Props @nickdaugherty

### DIFF
--- a/includes/classes/Indexable/Post/Post.php
+++ b/includes/classes/Indexable/Post/Post.php
@@ -359,6 +359,51 @@ class Post extends Indexable {
 	}
 
 	/**
+	 * Get an array of taxonomies that are indexable for the given post
+	 *
+	 * @param WP_Post $post Post object
+	 * @return array Array of WP_Taxonomy objects that should be indexed
+	 */
+	public function get_indexable_post_taxonomies( $post ) {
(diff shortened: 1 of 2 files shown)


felipeelia (changes requested):

felipeelia:
Related to https://github.com/10up/ElasticPress/pull/1969#issuecomment-830251085

felipeelia:
Closing in favor of #2603

PR #1958, opened 2020-11-05
Updated prepare_date_terms to only call date_i18n once

brandon-m-skinner:
### Description of the Change

Calling `date_i18n` a lot gets pretty expensive. Only calling it once for each `prepare_date_terms` call is much more efficient and make indexing faster.

Props @WPprodigy

### Alternate Designs

N/A

### Benefits

Removes extraneous calls to `date_i18n` and makes indexing faster.

### Possible Drawbacks

N/A

### Verification Process

There are 4 calls to `prepare_date_terms()` per post being indexed. Each call resulted in 12 calls to `date_i18n()`, which ends up going down a bit of a rabbit hole and eventually adding up to a lot of time spent doing filters and alloption calls. So per page of 500 posts being indexed, that is 24,000 calls (500 x 12*4)- seen in cachegrind.

Before the changes, a test group of 500 posts took 7.91601 seconds running prepare_date(). After this PR, 0.85888 seconds.

### Checklist:

### Changelog Entry

Changed prepare_date_terms to only call date_i18n once. Props @WPprodigy

### DIFF
--- a/includes/classes/Indexable/Post/Post.php
+++ b/includes/classes/Indexable/Post/Post.php
@@ -340,21 +340,32 @@ class Post extends Indexable {
 	 * @since 0.1.4
 	 * @return array
 	 */
-	private function prepare_date_terms( $post_date_gmt ) {
-		$timestamp  = strtotime( $post_date_gmt );
-		$date_terms = array(
-			'year'          => (int) date_i18n( 'Y', $timestamp ),
-			'month'         => (int) date_i18n( 'm', $timestamp ),
-			'week'          => (int) date_i18n( 'W', $timestamp ),
-			'dayofyear'     => (int) date_i18n( 'z', $timestamp ),
-			'day'           => (int) date_i18n( 'd', $timestamp ),
-			'dayofweek'     => (int) date_i18n( 'w', $timestamp ),
-			'dayofweek_iso' => (int) date_i18n( 'N', $timestamp ),
-			'hour'          => (int) date_i18n( 'H', $timestamp ),
-			'minute'        => (int) date_i18n( 'i', $timestamp ),
-			'second'        => (int) date_i18n( 's', $timestamp ),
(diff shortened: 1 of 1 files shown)


includes/classes/Indexable/Post/Post.php
```diff
@@ -340,21 +340,32 @@ public function prepare_document( $post_id ) {
...
+		];
+
+		// Combine all the date term formats and perform one single call to date_i18n() for performance.
+		$date_format    = implode( '||', array_values( $terms_to_prepare ) );
+		$combined_dates = explode( '||', date_i18n( $date_format, strtotime( $date_to_prepare ) ) );
+
+		// Then split up the results for individual indexing.
+		$date_terms = [];
+		foreach ( $terms_to_prepare as $term_name => $date_format ) {
+			$index_in_combined_format = array_search( $term_name, array_keys( $terms_to_prepare ) );
```
tlovett1, line 365: This logic seems a bit strange. Can't you just use `$terms_to_prepare[ $term_name ]`? Also for `array_search` you should use the strict comparison parameter.

tlovett1 (changes requested):

includes/classes/Indexable/Post/Post.php
WPprodigy: It is a bit strange, but the tricky portion requiring this is that we need the index to access from `$combined_dates` on the next line. Given a data to prepare like "'2019-12-17 22:28:46'", `$combined_dates` looks like this:

```
Array
(
    [0] => 2019
    [1] => 12
    [2] => 51
    [3] => 350
    [4] => 17
    [5] => 2
    [6] => 2
    [7] => 22
    [8] => 28
    [9] => 46
    [10] => 201912
)
```

Rahmon:
Closing in favor of #2214

PR #1959, opened 2020-11-05
Added Elasticsearch::get_documents and Indexable::multi_get

brandon-m-skinner:
### Description of the Change

We needed an easy way to fetch multiple documents per API call, so we added `get_documents` to the `Elasticsearch` class and then added `multi_get` to the `Indexable` class for ease of use.

Props @nickdaugherty

### Alternate Designs

N/A

### Benefits

Allows you to get multiple documents per API call.

### Possible Drawbacks

N/A

### Verification Process

We tested in manually and then started using it in functionality for our search offering which has been running on large sites for months now.

### Checklist:

### Changelog Entry

Added Elasticsearch::get_documents and Indexable::multi_get. Props @nickdaugherty

### DIFF
--- a/includes/classes/Elasticsearch.php
+++ b/includes/classes/Elasticsearch.php
@@ -616,6 +616,51 @@ class Elasticsearch {
 	}
 
 	/**
+	 * Get multiple documents from Elasticsearch given an array of ids
+	 *
+	 * @param  string $index Index name.
+	 * @param  array  $document_ids Array of document ids to get.
+	 * @since  3.5
+	 * @return boolean|array
+	 */
+	public function get_documents( $index, $document_ids ) {
+		$path = $index . '/_doc/_mget';
+
+		$request_args = [
+			'method' => 'POST',
+			'body'   => wp_json_encode(
+				array(
+					'ids' => $document_ids,
+				)
+			),
+		];
+
+		$request = $this->remote_request( $path, $request_args, [], 'post' );
+
+		if ( is_wp_error( $request ) ) {
+			return false;
+		}
+
+		$response_body = wp_remote_retrieve_body( $request );
+
+		$response = json_decode( $response_body, true );
+
+		$docs = [];
+
+		if ( is_array( $response['docs'] ) ) {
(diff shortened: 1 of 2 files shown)


tlovett1:
@brandon-m-skinner will this only work with ES 7.0+?

brandon-m-skinner:
[5.0 and higher](https://www.elastic.co/guide/en/elasticsearch/reference/5.0/docs-multi-get.html)

includes/classes/Elasticsearch.php
```diff
@@ -615,6 +615,51 @@ public function delete_network_alias( $alias ) {
...
+	/**
+	 * Get multiple documents from Elasticsearch given an array of ids
+	 *
+	 * @param  string $index Index name.
+	 * @param  array  $document_ids Array of document ids to get.
+	 * @since  3.5
+	 * @return boolean|array
+	 */
+	public function get_documents( $index, $document_ids ) {
+		$path = $index . '/_doc/_mget';
```
felipeelia, line 627: Unfortunately, this doesn't work with Elasticsearch 5. To have it working on versions older than 7, we'll have to replicate the code we have [here](https://github.com/10up/ElasticPress/blob/develop/includes/classes/Elasticsearch.php#L87) and in some other parts of this same class. Something like this:

`if ( version_compare( $this->get_elasticsearch_version(), '7.0', '<' ) ) {
	$path = apply_filters( 'ep_index_' . $type . '_request_path', $index . '/' . $type . '/_mget', $document_ids, $type );
} else {
	$path = apply_filters( 'ep_index_' . $type . '_request_path', $index . '/_doc/_mget', $document_ids, $type );
}
`

It'll also be needed to pass the `$type` as a parameter to the `get_documents` method. Something like:

`return Elasticsearch::factory()->get_documents( $this->get_index_name(), $this->slug, $object_ids );`

Would it be possible to have these changes applied to this PR, @brandon-m-skinner? Thanks!

felipeelia (changes requested):

felipeelia:
Closing it in favor or #2210. Thanks!

PR #1968, opened 2020-11-05
Added bypass for insert / delete for machines

brandon-m-skinner:
### Description of the Change

Allows indexing / deletion while in cron / WP CLI.

Props @nickdaugherty 

### Alternate Designs

N/A

### Benefits

Allows indexing / deletion while in cron / WP CLI.

### Possible Drawbacks

Allows indexing / deletion while in cron / WP CLI.

### Verification Process

We needed a way to do indexing / deletion while in cron / WP CLI. After this change, it worked.

### Checklist:

### Changelog Entry

Added permission check bypass for indexing / deleting for cron and WP CLI. Props @nickdaugherty

### DIFF
--- a/includes/classes/Indexable/Post/SyncManager.php
+++ b/includes/classes/Indexable/Post/SyncManager.php
@@ -50,6 +50,31 @@ class SyncManager extends SyncManagerAbstract {
 		add_action( 'added_post_meta', array( $this, 'action_queue_meta_sync' ), 10, 4 );
 		add_action( 'deleted_post_meta', array( $this, 'action_queue_meta_sync' ), 10, 4 );
 		add_action( 'wp_initialize_site', array( $this, 'action_create_blog_index' ) );
+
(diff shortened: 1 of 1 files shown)


felipeelia:
The current call to `apply_filters( 'ep_sync_insert_permissions_bypass'`... )` already contains a check for `DOING_CRON`. Perhaps it would be worth removing it from there and having it only in one place.

This seems to be a better approach to this problem than #1814, as we'd be solving two problems with the same method.

brandwaffle:
@felipeelia let's go with this one and drop #1814.

felipeelia:
Hey @brandon-m-skinner and @nickdaugherty,

Would you be interested in pushing a commit to remove the check for DOING_CRON we have [here](https://github.com/10up/ElasticPress/blob/develop/includes/classes/Indexable/Post/SyncManager.php#L228) before we merge the PR? Thanks in advance!!

Issue #750, opened 2017-03-02
Re-indexing terms on update (ex: slug)


kallehauge:
# Re-indexing terms on update (ex: slug)

## Case
Hello ppl! A client changed a slug for a public term today and went to the new URL and no results showed up. What happens, like you probably know, is that the sync-manager only looks for changes on `posts` and not `terms`.

```
/**
 * Setup actions and filters
 *
 * @since 0.1.2
 */
public function setup() {
	add_action( 'wp_insert_post', array( $this, 'action_sync_on_update' ), 999, 3 );
	add_action( 'add_attachment', array( $this, 'action_sync_on_update' ), 999, 3 );
	add_action( 'edit_attachment', array( $this, 'action_sync_on_update' ), 999, 3 );
	add_action( 'delete_post', array( $this, 'action_delete_post' ) );
	add_action( 'delete_blog', array( $this, 'action_delete_blog_from_index') );
	add_action( 'make_spam_blog', array( $this, 'action_delete_blog_from_index') );
	add_action( 'archive_blog', array( $this, 'action_delete_blog_from_index') );
	add_action( 'deactivate_blog', array( $this, 'action_delete_blog_from_index') );
	add_action( 'updated_postmeta', array( $this, 'action_queue_meta_sync' ), 10, 4 );
	add_action( 'added_post_meta', array( $this, 'action_queue_meta_sync' ), 10, 4 );
	add_action( 'shutdown', array( $this, 'action_index_sync_queue' ) );
}
```

## Why is this a problem?
A lot of 3. party plugins use `slug`'s for `tax_query`'s and this would make it fail because Elasticsearch still holds the old slug.
Although it would be possible to hook into `pre_get_posts` and alter the `tax_query` to fetch with `term_id`'s. I feel like this solution would have a lot of edge-cases and it would make `ElasticPress` less "out of the box" ready – The worst case scenario is that archives/lists completely fail due to this.

I've looked in the the API and it's only possible to re-index entire posts which is extremely slow/heavy since it has to get and prepare all posts before and, in this example, there might be several thousand products for a term.

## The question
Is this a scenario you have discussed before? And/or is it something you wish to support? If yes, then I already have some ideas of how this could be done and I would happily supply some of my time to cooperate on a PR.

In the meantime: have any of you done a workaround for this, previously? If yes, then I would be very interested in hearing how you handled it :smile:

/ Best regards

tlovett1:
[...]

PR #1969, opened 2020-11-05
Added action_set_object_terms

brandon-m-skinner:
### Description of the Change

Queues posts for re-indexing when their terms are changed.

Props @nickdaugherty

### Alternate Designs

N/A

### Benefits

Post data remains consistent with the database which keeps queries returning expected results.

### Possible Drawbacks

The default behaviour is problematic when there are a lot of posts. We only do this behaviour for 10k or less posts and offload any indexing with counts above that to be processed asynchronously.

### Verification Process

We noticed data inconsistencies when comparing posts from the DB against posts from ES when editing terms. After applying these changes, the index and the DB were consistent.

### Checklist:

### Changelog Entry

Added queuing of affected posts on term change. Props @nickdaugherty

### DIFF
--- a/includes/classes/Indexable/Post/SyncManager.php
+++ b/includes/classes/Indexable/Post/SyncManager.php
@@ -50,6 +50,7 @@ class SyncManager extends SyncManagerAbstract {
 		add_action( 'added_post_meta', array( $this, 'action_queue_meta_sync' ), 10, 4 );
 		add_action( 'deleted_post_meta', array( $this, 'action_queue_meta_sync' ), 10, 4 );
 		add_action( 'wp_initialize_site', array( $this, 'action_create_blog_index' ) );
+		add_action( 'set_object_terms', array( $this, 'action_set_object_terms' ), 10, 6 );
 	}
 
 	/**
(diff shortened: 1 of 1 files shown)


includes/classes/Indexable/Post/SyncManager.php
```diff
@@ -107,6 +108,78 @@ public function action_queue_meta_sync( $meta_id, $object_id, $meta_key, $meta_v
...
+	 * @param int    $object_id  Object ID.
+	 * @param array  $terms      An array of object terms.
+	 * @param array  $tt_ids     An array of term taxonomy IDs.
+	 * @param string $taxonomy   Taxonomy slug.
+	 * @param bool   $append     Whether to append new terms to the old terms.
+	 * @param array  $old_tt_ids Old array of term taxonomy IDs.
+	 * @since  3.5
+	 */
+	public function action_set_object_terms( $object_id, $terms, $tt_ids, $taxonomy, $append, $old_tt_ids ) {
+		$post_id = $object_id;
```
rebeccahum, line 128: Missing a check for `$this->kill_sync()`:

```suggestion
		if ( $this->kill_sync() ) {
			return;
		}
		$post_id = $object_id;
```

includes/classes/Indexable/Post/SyncManager.php
```diff
@@ -107,6 +108,78 @@ public function action_queue_meta_sync( $meta_id, $object_id, $meta_key, $meta_v
...
+
+		$indexable = Indexables::factory()->get( 'post' );
```
rebeccahum, line 130: For code consistency, let's use `$this->indexable_slug` even though it will end up evaluating to `post`:

```suggestion
		$indexable = Indexables::factory()->get( $this->indexable_slug );
```

includes/classes/Indexable/Post/SyncManager.php
```diff
@@ -107,6 +108,78 @@ public function action_queue_meta_sync( $meta_id, $object_id, $meta_key, $meta_v
...
+		$post_type = get_post_type( $post_id );
+
+		if ( defined( 'DOING_AUTOSAVE' ) && DOING_AUTOSAVE ) {
+			// Bypass saving if doing autosave
+			return;
+		}
+
```
rebeccahum, line 137: We should add a filter for skipping a sync triggered by term changes

rebeccahum (changes requested):
Left some notes. 

But overall, I think this is will prove to be useful, especially if we have clients with larger term sets — this way, they won't have to run a re-index to get the accurate terms assigned to posts.

felipeelia:
This also closes #750

felipeelia:
Closing in favor of https://github.com/10up/ElasticPress/pull/2603

PR #1987, opened 2020-11-26
Add a way to change the ordering of the features on the Features page

### LINKED ISSUE #1983: Ordering of features
On the Features page, the features are listed in the order as they are registered and there is no way to change this order. I would like that the plugin has a way to change the ordering of the features and to chose if a feature should be on the left or right side.

**Additional context**
I'm developing a feature that registers other features (subfeatures) and I would like to keep the main feature on the left side and the subfeatures on the right side. So, preserving the order, the boxes will appear in a consistent way on the features screen.

Rahmon:
### Description of the Change

This PR adds two properties to the abstract class Feature that allow the developer to change the order of the features and chose if a feature should be on the left or right side when the features are listed on the Features page. Furthermore, changes are done in the `dashboard-page.php` to sort the array of features and to put on the left or right side.

### Benefits

Change the order of the features to prioritize the most important, most used, or to show the features in a different way. 

### Verification Process

This has been tested manually just to validate the implementation. 

### Checklist:

### Applicable Issues

Closes #1983 

### Changelog Entry

Added custom sorting to features on the Feature page.


includes/classes/Feature.php
```diff
@@ -52,6 +52,23 @@ abstract class Feature {
 	 */
 	public $requires_install_reindex;
 
+	/**
+	 * The order in the features screen
+	 *
+	 * @var int
+	 * @since  3.5.1
```
felipeelia, line 59: @Rahmon can we update these 2 @since tags before we merge it?

includes/partials/dashboard-page.php
```diff
@@ -26,7 +26,19 @@
 <div class="wrap">
 	<h2 class="ep-list-features"><?php esc_html_e( 'List of features', 'elasticpress' ); // We use this since WP inserts warnings after the first h2. This will be hidden. ?></h2>
 	<div class="ep-features metabox-holder">
-		<?php $features = Features::factory()->registered_features; ?>
+		<?php
+		$features = Features::factory()->registered_features;
+		uasort(
+			$features,
+			function( $feature_a, $feature_b ) {
+				if ( $feature_a->order == $feature_b->order ) {
```
felipeelia, line 34: PHPCS gives a warning about this line. Any chance we cast both values to integer at the beginning of the function?

felipeelia (changes requested):

includes/classes/Feature.php
Rahmon: Done.

includes/partials/dashboard-page.php
Rahmon: Refactored. Thanks for it.

felipeelia (approved):

PR #1990, opened 2020-11-30
Feature/autosuggest basic auth

zack-carlson:
### Requirements

By adding basic auth credentials - users can use different security methods, such as basic auth provided by Open Distro.  The advantage being, read only users would be available to be configured, while on the back end; read and write user could be used without exposing the writable features to autosuggest.

### Description of the Change

Allowing the format of https://user:password@elasticsearchserver.com to pass basic auth headers within autosuggest.js 

### Alternate Designs

This information could be stored in separate configuration values; this was the simplest way to include the information here

### Benefits

One additional authorization method using autosuggest.

### Possible Drawbacks

The code parses credentials from a string; it's unclear that exposing basic auth will instruct users that they must also include a "read only user" to connect to as well as a "read write user".. this might require additional updates in the documentation.

### Checklist:

- [ x ] My change requires a change to the documentation.

### Changelog Entry

Added basic auth parsing if applicable to autosuggest.


brandwaffle:
@zack-carlson if you include the username:password in the autosuggest URL string it will be passed to the DOM in the current version of the plugin. This enables you to set a separate password and pass it to JS. Am I missing a feature you're looking for?

zack-carlson:
> @zack-carlson if you include the username:password in the autosuggest URL string it will be passed to the DOM in the current version of the plugin. This enables you to set a separate password and pass it to JS. Am I missing a feature you're looking for?

I may be mistaken, but using window.fetch you cannot supply the credentials in the URL, only in the headers; which is what I ran into when attempting this, specifically in the esSearch function of autosuggest.js

zack-carlson:
@brandwaffle Not for autosuggest: 
Screen Shot 2021-01-25 at 9 29 03 AM copy

brandwaffle:
@zack-carlson ok, thanks for testing that out, sounds like it won't work.

In general, ElasticPress isn't going to support explicitly passing credentials that are client-viewable, but we could certainly add a filter to the epas options to support custom headers. This would then let you add a PHP filter to add in your credentials if you want, and it also supports adding any other necessary headers for, say, AWS Elasticsearch or another custom service.

Would that approach work for you?

brandwaffle:
@zack-carlson I'm not talking about routing your HTTP calls through WordPress, rather adding a filter to the `epas` object which is localized to JS via PHP (this is what sets your endpoint URL, query parameters, etc). This happens here https://github.com/10up/ElasticPress/blob/f6f381942261eefa6062ad5c4be14768f7a4b78e/includes/classes/Feature/Autosuggest/Autosuggest.php#L374

This would let you add the headers via PHP, and the JS would be changed to read in headers from the localized `epas` object and pass them along in the request.

The reason we wouldn't support "read only" credentials is that there are features in ElasticPress like Protected Content that expose private content, and just making queries read only won't do anything to protect them. [As advised by Elastic](https://www.elastic.co/guide/en/elasticsearch/reference/current/modules-scripting-security.html#_do_not_expose_elasticsearch_directly_to_users) Elasticsearch endpoints should _never_ be directly available to the internet, regardless of what type of security is attempted on the credential level.

zack-carlson:
@brandwaffle it seems like the headers would be base64 encoded using this technique - as in they might as well be exposed through a user:pass@endpoint.com paradigm anyway?

brandwaffle:
@zack-carlson correct, this would not improve security. I'm just saying your pull request as written is specific to parsing the URL out into auth headers, and we won't be adding anything specific to auth headers because we don't want anyone doing this.

Being able to add _any_ header via a PHP filter would let you add client-side username/password auth for your Elasticsearch instance if you really feel you need to go down that path. Again, we 100% do not support this and you probably shouldn't be doing it for all the reasons outlined in the Elastic article, but I can't stop you from doing it 😄 

The approach I outlined will also let other folks add their own needed client-side headers regardless of what header they want to add, so it would be more generally helpful to a wider range of ElasticPress users, while still enabling you to accomplish what you want with a simple filter.

I'm going to close this out and create a ticket referencing it to add the option to filter in headers. Thank you for raising the issue and contributing to the project--we should be able to add the header filter in the next version of the plugin.

Issue #1988, opened 2020-11-27
Facets are throwing errors when terms don't exist


shaneonabike:
**Describe the bug**
I came across a small coding issue whereby ElasticPress was throwing an error when no terms are being returned. It is a really simple fix in the Facets/Widget.php file. Essentially, the code is only verifying if no terms are returned, but never whether an error is returned.

```
               /**
                 * Get all the terms so we know if we should output the widget
                 */
                $terms = get_terms(
                        /**
                         * Filter arguments passed to get_terms() while getting all possible terms for the facet widget.
                         *
                         * @since  3.5.0
                         * @hook ep_facet_search_get_terms_args
                         * @param  {array} $query Weighting query
                         * @param  {string} $post_type Post type
                         * @param  {array} $args WP Query arguments
                         * @return  {array} New query
                         */
                        apply_filters(
                                'ep_facet_search_get_terms_args',
                                [
                                        'taxonomy'   => $taxonomy,
                                        'hide_empty' => true,
                                ],
                                $args,
                                $instance
                        )
                );

                /**
                 * No terms!
                 */
                if ( 0 === $terms) {
                        return;
                }
```
Should become:

```
               /**
                 * Get all the terms so we know if we should output the widget
                 */
                $terms = get_terms(
                        /**
                         * Filter arguments passed to get_terms() while getting all possible terms for the facet widget.
                         *
                         * @since  3.5.0
                         * @hook ep_facet_search_get_terms_args
                         * @param  {array} $query Weighting query
                         * @param  {string} $post_type Post type
                         * @param  {array} $args WP Query arguments
                         * @return  {array} New query
                         */
                        apply_filters(
                                'ep_facet_search_get_terms_args',
                                [
                                        'taxonomy'   => $taxonomy,
                                        'hide_empty' => true,
                                ],
                                $args,
                                $instance
                        )
                );

                /**
                 * No terms!
                 */
                if ( 0 === $terms || is_wp_error( $terms )) {
                        return;
                }
```

shaneonabike:
Adding that little extra check basically prevents an exception being thrown, and the search functionality to continue properly.

PR #1992, opened 2020-12-02
Bugfix/1938 facets get terms wp error

ciprianimike:
### Description of the Change

added is_wp_error() check to get_terms() result in the Widgets Facets.

Since [get_terms()](https://developer.wordpress.org/reference/functions/get_terms/
) may return a WP_Error object, adding the control will avoid exceptions in case the result of get_terms is an error.

### Verification Process
The bug was replicated, fixed and tested locally, using the Facets Widget.
It is a simple and "straightforward change".

### Checklist:

### Applicable Issues

https://github.com/10up/ElasticPress/issues/1988

### Changelog Entry

Bug fixes:
* Fix php exception when get_terms() returns a WP_Error object in the Facets Widget.

### DIFF
--- a/includes/classes/Feature/Facets/Widget.php
+++ b/includes/classes/Feature/Facets/Widget.php
@@ -99,9 +99,9 @@ class Widget extends WP_Widget {
 		);
 
 		/**
-		 * No terms!
+		 * Terms validity check
 		 */
-		if ( 0 === $terms || is_wp_error( $terms ) ) {
+		if ( is_wp_error( $terms ) || empty( $terms ) ) {
 			return;
 		}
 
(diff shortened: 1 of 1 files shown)


includes/classes/Feature/Facets/Widget.php
```diff
@@ -101,7 +101,7 @@ public function widget( $args, $instance ) {
 		/**
 		 * No terms!
 		 */
-		if ( 0 === $terms ) {
+		if ( 0 === $terms || is_wp_error( $terms ) ) {
```
bengreeley, line 104: @ciprianimike I'd recommending having `is_wp_error()` come before 0 === $terms to avoid any warnings trying to compare an integer with an object if we have an error.

tlovett1: The docs are a little confusing about the return value of `get_terms`. I'd suggest just checking if terms is `empty` or is a WP error object.

Issue #1971, opened 2020-11-10
Looking for more options to determine CLI indexing status


ChrisWiegman:
**Is your enhancement related to a problem? Please describe.**

In a hosting environment it is easy to kick off an indexing operation in a "headless" fashion while not connected to the session. This is necessary for most users yet makes determining the status of the index difficult. While new commands introduced in 3.5.1 make this possible for support personnel and engineers, they don't, unfortunately, provide enough for feedback to the user.

**Describe the solution you'd like**

We are looking for two changes to improve this experience:

1. Initiating an indexing operation via CLI should display the status in the ElasticPress dashboard in the same manner as initiating the index operation from the dashboard. In this case the progress bar should reflect the current status and controls for starting/stopping/pausing an indexing operation should reflect the nature of the ongoing CLI index operation.

2. In order to provide indexing status in the systems that initiated the operation, we request that the status information, in a similar fashion to the commands introduced in 3.5.1, also be available via the WordPress REST API.

**Designs**

We've discussed initial ideas with Vasken and will be happy to continue the conversation here or as otherwise appropriate.

**Describe alternatives you've considered**

The ideal would be to make this status available via the ElasticPress service API but we realize that, due to the stateless nature of indexing operations, such a change would be more difficult. We are open to other suggestions or ideas that would achieve the above goals via a different implementation.

**Additional context**
N/A

ChrisWiegman:
Do we have an estimate on when this will be available?

PR #1997, opened 2020-12-09
CLI indexing status in the ElasticPress dashboard

Rahmon:
### Description of the Change

Currently, when an indexing operation is started from WP-CLI, the dashboard doesn't provide enough feedback about the status of the index to the user. This change displays the progress bar when an indexing operation is running even when started from WP-CLI.

Some cases handled by this change: 

The user clicks on the start button but there is a WP CLI indexing operation running:

ep_cli_status

Page is reloaded and there is a WP CLI indexing operation running:

ep_cli_status_reload

It's no new case but just to show the browser and terminal side by side:
ep_cli_status_terminal

### Benefits

Users will have better feedback about indexing status.

### Possible Drawbacks

n/a

### Verification Process

This has been tested manually: (1) starting an indexing operation from the terminal; and (2) trying to start indexing from the dashboard page or reload the dashboard page.

### Checklist:

### Applicable Issues

Fixes: #1971 

### Changelog Entry

Improved the feedback about the status of indexing on the dashboard page.


brandwaffle:
@ChrisWiegman did you want to take a look at this?

ChrisWiegman:
This looks like it will go a long way to helping with the underlying issue of #1971. I would still love to see the status available via REST, however this will, at least, give us somewhere to link users to.

Rahmon:
Some updates:

I've added a way to stop the indexing operation started from CLI in the dashboard.

stop-cli-sync-from-dashboard

It's also possible to stop the indexing started from the dashboard in the WP-CLI using the command `wp elasticpress stop-indexing`.

stop-sync-from-wp-cli

Rahmon:
@ChrisWiegman I made the status available via REST API.

includes/classes/Command.php
```diff
@@ -1357,4 +1362,55 @@ function ( $prefix ) use ( $assoc_args ) {
...
+
+		$options = $wpdb->options;
+
+		$sql = "
+			SELECT option_value
+			FROM $options
+			WHERE option_name = '_transient_ep_wpcli_sync_interrupted'
+		";
+
+		$should_interrupt_sync = $wpdb->get_var( $sql ); // phpcs:ignore
```
felipeelia, line 1381: @Rahmon isn't is possible to use get_option or get_transient here? And are we accounting for multisite?

includes/dashboard.php
```diff
@@ -1138,3 +1171,48 @@ function action_wp_ajax_ep_site_admin() {
...
+	register_rest_route(
+		'elasticpress/v1',
+		'indexing_status',
+		[
+			'methods'             => 'GET',
+			'callback'            => __NAMESPACE__ . '\handle_indexing_status',
+			'permission_callback' => '__return_true',
+		]
+	);
+}
```
felipeelia, line 1188: I wonder if this shouldn't be inside a new class, @Rahmon, especially if we think this can receive new endpoints (authenticated users could start a reindex, perhaps?)
Also, worth mentioning we probably should create a function to keep this namespace constant... includes/classes/Feature/SearchOrdering/SearchOrdering.php also creates some endpoints, and if we decide to go to `elasticpress/v2`, we'll have several places to change.

includes/classes/Command.php
Rahmon: When I use `get_transient` in the WP-CLI env and after starting the sync, I can't retrieve the updated value to stop it. That is, I always get `false` after the command is running even when the value is updated.

includes/dashboard.php
Rahmon: I agree with you. But I think this should be handle in another PR. What do you think?

includes/classes/Command.php
Rahmon: In a multisite environment, I've tested two scenarios:

1. Plugin activated individually in each site. The command in the CLI needs the parameter `--url`
2. Plugin network activated. Only the network (wp-admin/network) has access to the ElasticPress dashboard.

includes/classes/Command.php
```diff
@@ -1357,4 +1371,79 @@ function ( $prefix ) use ( $assoc_args ) {
...
+			$sql = "
+				SELECT option_value
+				FROM $options
+				WHERE option_name = '_transient_{$transient}'
+			";
+
+			$should_interrupt_sync = $wpdb->get_var( $sql ); // phpcs:ignore
+		}
+
+		return (bool) $should_interrupt_sync;
```
tott, line 1446: You might want to make sure to return a non-false value here. Otherwise it will still fall back to the regular transient functionality here https://core.trac.wordpress.org/browser/tags/5.6/src/wp-includes/option.php#L787

includes/classes/Command.php
```diff
@@ -1357,4 +1371,79 @@ function ( $prefix ) use ( $assoc_args ) {
...
+
+		if ( wp_using_ext_object_cache() ) {
+			$should_interrupt_sync = wp_cache_get( $transient, 'transient' );
+		} else {
+			$options = $wpdb->options;
+
+			$sql = "
+				SELECT option_value
+				FROM $options
+				WHERE option_name = '_transient_{$transient}'
```
tott, line 1440: for good practice. let's add a "LIMIT 1" to the query and run it through $wpdb->prepare

Issue #2001, opened 2020-12-14
Weighting and WooCommerce 


mfs-mindsize:
The EP weight settings for WC products do not getting applied to searches does vie the WC Search widget, that is post_type=product.  

**Steps to Reproduce**

1. Install and activate Debug Bar, as well as the EP Debug Bar add-on.
2. Add the (Site) Search widget, as well as a Product Search widget somewhere (e.g., sidebar). 
3. Use the Site search and review the Query Body. You'll see weighting applied.
4. Use the Product Search and review the Query Body. No weighting.

Or do a Site Search and then add &post_type=product to the URL. 

**Expected behavior**

When &post_type=product the weights - as defined on the EP Search fields & Weighting page - will be used for a product search query. 

**Additional context**
This issue appears to be related to:

Issue 1722 [Weighting engine not working with Woocommerce search](https://github.com/10up/ElasticPress/issues/1722) 
PR 1801[ [WooCommerce] Add Fields to the Weighting Dashboard](https://github.com/10up/ElasticPress/pull/1801)

However, we found that we could fix the issue by:

[Weighting.php](https://github.com/10up/ElasticPress/blob/develop/includes/classes/Feature/Search/Weighting.php) > method: do_weighting() > 

```
		if ( isset( $args['search_fields'] ) && ! empty( $args['search_fields'] ) ) {
			return $formatted_args;
		}
```
If we remove this conditional (read: wrapped it in another conditional with a filter that returns a bool - so we could toggle it on and off) we were able to get weights add to product searches.

mgratch:
@tlovett1 @moraleida I thought it was worth pointing out that #1801 #1722 are only loosely related to this ticket. Where those tickets are related to weighting WC attributes via the dashboard. This ticket is related to making use of weighting when using any search directly associated with WC. i.e. If you use the WooCommerce included search widget or the default search included with the storefront theme search queries will automatically include `post_type=product` and therefore weighting will not be applied.

felipeelia:
Hi @mgratch , now that #1801 is merged, can you please try the code in the develop branch?

With the change made [here](https://github.com/felipeelia/ElasticPress/blob/b632170a9d8bd4a3c4a3a703dd61a263c4d004d1/includes/classes/Feature/WooCommerce/WooCommerce.php#L436), the WooCommerce feature is not setting `search_field` anymore, then that `if` you pointed out won't be true.

Thanks!

PR #2002, opened 2020-12-14
added conditional + filter to allow for additional control within do_weighting()

mfs-mindsize:
### Description of the Change

This PR is a proposed solution to the bug reported here:

Issue 2001 - [Weighting and WooCommerce](https://github.com/10up/ElasticPress/issues/2001)

### Alternate Designs

The proposed selected for its simplcity.

### Benefits

More control of when / when not the weighting will be used in the search query. Currently, for example, WC product searches are not weighted (even though there are weight settings under EP's Search Fields & Weighting). 

### Possible Drawbacks

Nothing obvious comes to mind. 

### Verification Process

Used the filter to pass over the conditional that was returning WC products (i.e., weights weren't used). 

### Checklist:

### Changelog Entry

Added filter 'ep_weighting_check_search_fields' in Weighting.php's method: do_weighting()

### DIFF
--- a/includes/classes/Feature/Search/Weighting.php
+++ b/includes/classes/Feature/Search/Weighting.php
@@ -519,11 +519,16 @@ class Weighting {
 	public function do_weighting( $formatted_args, $args ) {
 
 		/**
-		 * If search fields is set on the query, we should use those instead of the weighting, since the query was
-		 * overridden by some custom code
+		 * Force the weightings to be utilized but bypassing the inner conditional (which might cause a return)
 		 */
-		if ( isset( $args['search_fields'] ) && ! empty( $args['search_fields'] ) ) {
-			return $formatted_args;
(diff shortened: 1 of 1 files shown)


tlovett1:
This definitely works as a temporary solution but I feel like we need to address the real issue. When the WC feature sees a search and the post type is `product` it sets a bunch of search fields e.g. `_sku`. When the weighting feature sees search fields, it bails. We probably need to having the weighting engine attempt to cleverly merging in the weightings when search fields are set.

Feel free to submit a PR for that or I can try to do it for the next bug release.

Thanks!

mfs-mindsize:
@tlovett1 - Understood. Question: What's the purpose of having the product weighting "metabox" in the UI then? How and where do those weight settings get applied? I look as if (e.g.) those weights are applied to searches against posts. Why not products? How do those UI based settings for products matter?

felipeelia:
The root problem was solved when we merged #1801, so I'm closing this one out. Thanks!

Issue #1941, opened 2020-10-28
There was an error updating the synonym list.


tlovett1:
If no index has been created and I updated synonyms. I receive this error message:

```
There was an error updating the synonym list.
```

scooterlord:
I am also getting the same error, however, I already have an index.

embold-given:
I feel like this was merged and closed without being addressed. The task it was merged with only dealt with the notification message. It doesn't do anything about not being able to add synonyms.

I'm in the same boat as scooterlord. We had a list of synonyms, now it's gone, when we try to save any new ones, we get the message:
`There was an error storing you synoyms.`

We're on 3.5.2, we have an index, I've checked to make sure we have the "fix" from the linked task. The issue persists.

PR #2049, opened 2021-01-26
Fix synonyms list update

Rahmon:
### Description of the Change

The synonyms list is saved in a custom post type that is not accessible through the admin dashboard. However, when the synonyms list is updated, the post type is changed from `ep-synonym` to `post`. 

As a consequence, this post becomes available in the list of posts where can accidentally be deleted. If the post with the synonyms list is deleted, it's no possible to save a new one because the plugin references the deleted post and raises an error.

This change ensures that the custom post type doesn't change on the update and adds an upgrade routine to fix this problem.

### Benefits

Users can add/update the synonyms list.

### Verification Process

This has been tested manually.

### Checklist:

### Applicable Issues

Fixes: #1941 

### Changelog Entry

Fixed on update the synonyms list.

### DIFF
--- a/includes/classes/Upgrades.php
+++ b/includes/classes/Upgrades.php
@@ -45,6 +45,7 @@ class Upgrades {
 		 */
 		$routines = [
 			'3.5.2' => [ 'upgrade_3_5_2', 'init' ],
+			'3.5.3' => [ 'upgrade_3_5_3', 'init' ]
 		];
 
 		array_walk( $routines, [ $this, 'run_upgrade_routine' ] );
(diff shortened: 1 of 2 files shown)


christianc1 (approved):

includes/classes/Upgrades.php
```diff
@@ -45,6 +45,7 @@ public function setup() {
 		 */
 		$routines = [
 			'3.5.2' => [ 'upgrade_3_5_2', 'init' ],
+			'3.5.3' => [ 'upgrade_3_5_3', 'init' ]
```
felipeelia, line 48: @Rahmon Linter seems to be complaining about this line. Can you please add a comma at the end of the line? Thanks!

PR #2056, opened 2021-01-28
Remove elasticpress_synonyms_post_id option

Rahmon:
### Description of the Change

This change refactors the `get_synonym_post_id` function to remove the dependence of `elasticpress_synonyms_post_id` option.

### Verification Process

This has been tested manually.

### Checklist:

### DIFF
--- a/includes/classes/Feature/Search/Synonyms.php
+++ b/includes/classes/Feature/Search/Synonyms.php
@@ -769,7 +769,8 @@ class Synonyms {
 			[
 				'post_content' => $this->example_synonym_list(),
 				'post_type'    => self::POST_TYPE_NAME,
-			]
+			],
+			true
 		);
 	}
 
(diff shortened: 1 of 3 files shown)


includes/classes/Feature/Search/Synonyms.php
```diff
@@ -770,12 +760,38 @@ public static function prepare_localized_token( $token, $primary = false ) {
 	}
 
 	/**
-	 * Check if synonym post exists
+	 * Insert default synonym post
 	 *
-	 * @param int $id The post id.
-	 * @return boolean
+	 * @return int|WP_Error
```
felipeelia, line 765: @Rahmon looking at the docs, it seems `wp_insert_post()` will return a `WP_Error` only if the second parameter is set. Do you want to include that parameter?

felipeelia (changes requested):

includes/classes/Feature/Search/Synonyms.php
Rahmon: You're right. I'll remove `WP_Error` because now we don't need that parameter set to true.

felipeelia (approved):

Issue #2062, opened 2021-02-02
Allow for faceting via REST API


moraleida:
**Is your enhancement related to a problem? Please describe.**

Currently ElasticPress does not allow custom queries to use the Aggregations feature, instead relying only on main queries. This excludes all REST API queries, native or custom.

**Describe the solution you'd like**

Allow for Aggregations to be manually permitted in custom queries so they can be returned in REST API requests and others.

PR #2063, opened 2021-02-02
introduce a filter to allow custom queries to be facetable

moraleida:
### Description of the Change

Introduces a new filter (`ep_is_facetable`) to allow bypassing the checks in `Facets\is_facetable()`, effectively allowing any query to have facets handled natively.

### Benefits

Currently the `is_facetable()` checks only allow for queries that have `$query->is_main_query()` set to true. This excludes any and all REST API requests, either through the native `wp/v2/search` endpoint or via custom endpoints, because the REST API does not set a main query in core and forcing a query to be the main query is likely to cause other issues.

This change allows developers to determine that specific queries should have aggregations even if they are not the main query.

### Possible Drawbacks

This change should be invisible to current installs and won't have any effects unless the filter is used.

### Verification Process

Tested with both the native Search endpoint and a custom endpoint I can see `aggs` key being properly added to the ES query, as well as aggregation results being properly set to the `$wp_facet_aggs` global.

### Checklist:

### Applicable Issues

Implements the changes needed for #2062 

### Changelog Entry

Added a new filter to allow faceting on custom queries

### DIFF
--- a/includes/classes/Feature/Facets/Facets.php
+++ b/includes/classes/Feature/Facets/Facets.php
@@ -184,6 +184,19 @@ class Facets extends Feature {
 	 * @return bool
 	 */
 	public function is_facetable( $query ) {
+
+		/**
+		 * Bypass the standard checks and set a query to be facetable
+		 *
+		 * @hook ep_is_facetable
+		 * @param  {bool}     $bypass Defaults to false.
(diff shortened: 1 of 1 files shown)


felipeelia (approved):

tlovett1:
Looks good to me

tlovett1 (approved):

PR #2081, opened 2021-02-12
Add the `ep_highlight_should_add_clause` filter

### LINKED ISSUE #2051: Ajax search load-more doesn't highlight search word
**Describe the bug**
I'm building a custom search site with a load more button, where the user can add more search results over ajax. The inital result is build over the basic WordPress search.php template with the global loop result. Where everything works fine and the searched word is highlighted. If I click on the load-more, the posts load as expected, but the highlighting is missing.

**Expected behavior**
When I load content over ajax and I've highlighting enabled it should highlight the search word too.

**Screenshots**
The first post is form the search.php template, the second is loaded over ajax.
image

**Bug**
After doing a bit of research in the plugin I found the bug in the file `elasticpress/includes/classes/Feature/Search/Search.php`. Here in line 197. You do an early return and the highlighting is not added.
```
if ( is_admin() || ( defined( 'REST_REQUEST' ) && REST_REQUEST ) ) {
    return $formatted_args;
}
```
The problem is that the `is_admin` function returns true when doing ajax requests. So something like this would be better
```
if ( (is_admin() && !wp_doing_ajax()) || ( defined( 'REST_REQUEST' ) && REST_REQUEST ) ) {
    return $formatted_args;
}
```

The same thing is in line 246, here should it be
```
if ( ! (is_admin() && !wp_doing_ajax()) && ! empty( $args['s'] ) ) {
```

Or may you use the `ep_ajax_wp_query_integration` filter to check if ajax support is needed.

**Additional context**
I've tested the code with my two modifications and the highlighting works. I hove you can deploy this two little changes into the next version.

felipeelia:
### Description of the Change

This PR adds a new `ep_highlight_should_add_clause` filter to let developers decide where the `highlight` clause should be added to the ES query.

### Alternate Designs

n/a

### Benefits

Developers can use highlight in AJAX and REST requests, for example.

### Possible Drawbacks

n/a

### Verification Process

1. Added the code
2. Tested if everything worked as before
3. Added `add_filter( 'ep_highlight_should_add_clause', '__return_true' );` to a mu-plugin
4. Visited https://ep-352.test/wp-json/wp/v2/posts?search=lorem and saw <mark> tags

### Checklist:

### Applicable Issues

Fixes #2051

### Changelog Entry

Add a new filter to let developers decide where the `highlight` clause should be added to the ES query

### DIFF
--- a/includes/classes/Feature/Search/Search.php
+++ b/includes/classes/Feature/Search/Search.php
@@ -194,7 +194,30 @@ class Search extends Feature {
 			return $formatted_args;
 		}
 
-		if ( is_admin() || ( defined( 'REST_REQUEST' ) && REST_REQUEST ) ) {
+		/** This filter is documented in search_setup() method. */
+		$should_send_in_ajax = ( defined( 'DOING_AJAX' ) && DOING_AJAX ) && apply_filters( 'ep_ajax_wp_query_integration', false );
+
+		/**
+		 * Filter whether to add the `highlight` clause in the query or not.
+		 *
+		 * @since  3.6.0
+		 * @hook ep_highlight_should_add_clause
+		 * @param  {bool}  $add_highlight_clause True means the clause should be added.
+		 * @param  {array} $formatted_args  ep_formatted_args array
+		 * @param  {array} $args  WP query args
+		 * @return {bool}  New $add_highlight_clause value
+		 */
+		$add_highlight_clause = apply_filters(
+			'ep_highlight_should_add_clause',
+			(
+				( ! is_admin() || $should_send_in_ajax ) &&
+				( ! defined( 'REST_REQUEST' ) || ! REST_REQUEST )
+			),
+			$formatted_args,
+			$args
+		);
+
+		if ( ! $add_highlight_clause ) {
 			return $formatted_args;
 		}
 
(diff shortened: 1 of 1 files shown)


includes/classes/Feature/Search/Search.php
```diff
@@ -242,17 +265,15 @@ public function add_search_highlight_tags( $formatted_args, $args ) {
...
-		if ( ! is_admin() && ! empty( $args['s'] ) ) {
-			foreach ( $fields_to_highlight as $field ) {
-				$formatted_args['highlight']['fields'][ $field ] = [
-					'pre_tags'            => [ $opening_tag ],
-					'post_tags'           => [ $closing_tag ],
-					'type'                => 'plain',
-					'number_of_fragments' => 0,
-				];
-			}
+		foreach ( $fields_to_highlight as $field ) {
```
Rahmon, line 268: In line 234, the return of the filter is an array but the documentation said that is a string. Can you update that? :)

```php
/**
		 * Filter the fields that should be highlighted.
		 *
		 * @since 3.5.1
		 * @hook ep_highlighting_fields
		 * @param  {array} $fields Highlighting fields
		 * @param  {array} $formatted_args array
		 * @param  {array} $args WP_Query args
		 * @return  {string} New Highlighting fields
		 */
		$fields_to_highlight = apply_filters(
			'ep_highlighting_fields',
			[ 'post_title', 'post_content' ],
			$formatted_args,
			$args
		);
```

Rahmon (dismissed):
LGTM :+1:

includes/classes/Feature/Search/Search.php
felipeelia: Fixed in a84852e. Good catch @Rahmon

Rahmon (approved):

PR #2090, opened 2021-02-16
WP-CLI commands to get and set the search algorithm version

felipeelia:
### Description of the Change

This PR adds two new WP-CLI commands:

- `wp elasticpress set-algorithm-version [--version=<version>] [--default]`

- `wp elasticpress get-algorithm-version`

It also adds a PHP docblock for the `ep_search_algorithm_version` filter.

### Benefits

Users will be able to change the search algorithm version without PHP code.

### Changelog Entry

Add the new `set-algorithm-version` and `get-algorithm-version` WP-CLI commands.

### DIFF
--- a/includes/classes/Command.php
+++ b/includes/classes/Command.php
@@ -1434,6 +1434,61 @@ class Command extends WP_CLI_Command {
 	}
 
 	/**
+	 * Set the algorithm version.
+	 *
+	 * @synopsis [--version=<version>] [--default]
+	 * @subcommand set-algorithm-version
+	 *
+	 * @since       3.5.4
+	 * @param array $args Positional CLI args.
+	 * @param array $assoc_args Associative CLI args.
+	 */
+	public function set_search_algorithm_version( $args, $assoc_args ) {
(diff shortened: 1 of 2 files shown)


Rahmon:
@felipeelia Looks good to me.

I'm just wondering if you need the option `--version` as is. Because there is an `if` [that just check if the version is 3.5](https://github.com/10up/ElasticPress/blob/f7e39d45a4451e819c92748a37dad04b9b5d1530/includes/classes/Indexable/Post/Post.php#L1093)

image

Maybe the option could be just `--3.4`.

felipeelia:
I thought about it @Rahmon but I guess a `--version` parameter will keep things more future-proof, i.e., we won't have to revisit the WP-CLI command if a `3.6` algo version comes out AND I can imagine a scenario where it also can be something useful for the Labs plugin too. For example, we decide to test a completely new algorithm in that plugin, we could use that same parameter to it.

Makes sense?

Rahmon (dismissed):

PR #2094, opened 2021-02-19
Autosuggest: check if `epas.highlightingTag` is set before using it

### LINKED ISSUE #2093: Click on autosuggest item fails when highlighting is not active
**Describe the bug**

When the highlighting feature is not active, clicking on an autosuggest entry will fail.

**Steps to Reproduce**

1. Disable `Highlighting status` under `Post Search`
2. Enable the autosuggest
3. Search for something
4. Click on an entry
5. The page doesn't change

**Expected behavior**

For #5, the user should be redirected to the other page.

felipeelia:
### Description of the Change

This PR fixes the problem stated in #2093: Autosuggest won't behave properly when Highlighting is not active.

### Applicable Issues

Fixes #2093 

### Changelog Entry

Fix a problem in autosuggest when highlighting is not active.


assets/js/autosuggest.js
```diff
@@ -266,7 +266,7 @@ function updateAutosuggestBox(options, input) {
 	suggestList.addEventListener('click', (event) => {
 		event.preventDefault();
 		const target =
-			event.target.tagName === epas.highlightingTag.toUpperCase()
+			epas.highlightingTag && event.target.tagName === epas.highlightingTag.toUpperCase()
```
Rahmon, line 269: How about using [optional chaining](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining) here? 

```js
event.target.tagName === epas.highlightingTag?.toUpperCase()
```

Rahmon (approved):
LGTM

brandwaffle (dismissed):

PR #2105, opened 2021-02-25
Autosuggest: Helper function to delete cached query

### LINKED ISSUE #2104: Provide an easy way to clear the query cached for Autosuggest

Currently, we cache the query generated for Autosuggest for 5 minutes in [`\ElasticPress\Feature\Autosuggest::intercept_search_request()`
](https://github.com/10up/ElasticPress/blob/develop/includes/classes/Feature/Autosuggest/Autosuggest.php#L541). If we change the algorithm with the new `wp elasticpress set-algorithm-version` command, for example, that change will be reflected on Autosuggest only when that cache expires.

It would be good if we have a function clearing that cache and call it, for example, after that command is executed. Possibly, we'd have to run it on other scenarios as well (a full reindex after a remapping, perhaps?).

felipeelia:
### Description of the Change

This PR fixes the problem outlined in #2104. It also adds some new actions:

- ep_wp_cli_after_index
- ep_after_dashboard_index
- ep_cli_before_set_search_algorithm_version
- ep_cli_after_set_search_algorithm_version
- ep_after_update_feature
- ep_cli_before_clear_index
- ep_cli_after_clear_index

### Applicable Issues

Closes #2104 

### Changelog Entry

- Add new actions: ep_wp_cli_after_index, ep_after_dashboard_index, ep_cli_before_set_search_algorithm_version, ep_cli_after_set_search_algorithm_version, ep_after_update_feature, ep_cli_before_clear_index, and ep_cli_after_clear_index

### DIFF
--- a/includes/classes/Command.php
+++ b/includes/classes/Command.php
@@ -1282,6 +1282,15 @@ class Command extends WP_CLI_Command {
 	public function clear_index() {
 		$this->delete_transient();
 
+		/**
+		 * Fires after the CLI `clear-index` command is executed.
+		 *
+		 * @hook ep_cli_clear_index
+		 *
+		 * @since 3.5.5
+		 */
+		do_action( 'ep_cli_clear_index' );
+
 		WP_CLI::success( esc_html__( 'Index cleared.', 'elasticpress' ) );
 	}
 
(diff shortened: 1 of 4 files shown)


Rahmon (approved):
LGTM

PR #2136, opened 2021-03-15
Allow custom query on autosuggest based on input

### LINKED ISSUE #2135: Autosuggest: Allow multiple input search fields with different queries for post types

The problem I'm currently facing is that I was unable to find a way to have multiple search fields that are capable of querying specific/different post types in the search.

A way to be able to adjust the query before autosuggest fetch based on the input field information (like id, class, etc).

**Designs**

Add a check for a `window` function just like `epDataFilter` to allow query customization before fetching the results.

Any other implementation that requires less effort or that is already implemented and is just not on the sight atm would be highly considered.

**Additional context**

With this abstraction not only the post types would be customizable but pretty much anything within the query.

fagiani:
### Description of the Change
This proposed change aims to resolve issue #2135 where a custom query can be made based on the `input` element
before the search fetch happens.

The change shouldn't impact any current users and will allow any new implementations to use a javascript function called `window.epCustomQuery` to return a modified version of the `window.epas` on the fly.
 

### Alternate Designs
There were considered two alternatives to this PR:
- Implementing within the `esSearch` function on the very top was considered because it would eventually benefit of the `searchText` length but it looked a bit offset as there was already a place where the `queryJSON` is defined.
- Overriding `window.fetch()` to intercept the query but there would be no information about the `input` at that moment which would make it hard to achieve that goal.

### Benefits
It will enable users to search using different inputs with different queries, including post types and other parameters.

### Possible Drawbacks
A possible drawback is that a malformed query customization could lead to further errors down stream.

### Checklist:

### Applicable Issues
closes #2135

### Changelog Entry
Allow query customization for autosuggest before fetching the search based on the input field.

### DIFF
--- a/assets/js/autosuggest.js
+++ b/assets/js/autosuggest.js
@@ -128,15 +128,22 @@ function selectItem(input, element) {
  * Build the search query from the search text - the query is generated in PHP
  * and passed into the front end as window.epas = { "query...
  *
+ * @param {object} input - originating input field
  * @returns {string} json string
  */
-function getJsonQuery() {
+function getJsonQuery(input) {
 	if (typeof window.epas === 'undefined') {
 		const error = 'No epas object defined';
 
 		// eslint-disable-next-line no-console
 		console.warn(error);
 		return { error };
+	}
+
+	// allow for customized query before sending it to
+	// the search fetch endpoint based on the input
+	if (typeof window.epCustomQuery !== 'undefined') {
+		return window.epCustomQuery(window.epas, input);
 	}
 
 	return window.epas;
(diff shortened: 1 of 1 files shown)


fagiani:
@felipeelia I wasn't able to determine whether any of the errors reported on the checks are actually related to the code committed on this PR. I'd appreciate any insights on how to fix them. Thanks again!

felipeelia:
Instead of using a global variable like `window.epCustomQuery`, we should introduce a proper filter for that. As the long-term plan is to actually change the autosuggest approach entirely (using a proper API instead of sending full ES queries) and this PR is 4 years old already, I'm closing it now. If @fagiani (or anyone else) still wants to send this contribution, a new PR would be preferred. Thanks and sorry for the super delayed reply here.

Issue #2121, opened 2021-03-08
Network wide search may switch away from the site the search is performed on


gonzomir:
**Describe the bug**
When retrieving the posts in a network-wide search, a `switch_to_blog()` is called as part of `the_post` hook, but `restore_current_blog()` is not called immediately after that, it's called in the next `the_post` hook if the next post is from a different site. This works OK in the loop, but some plugins (Yoast for example) may trigger the hook outside the loop. In that case WordPress will be left switched to the blog the first result came from, and if it's not the blog the search was performed on, unexpected thing happen.

I didn't go that deep to investigate what exactly Yoast is doing, but the hook is triggered when it generates it's output in the `<head />`.

**Steps to Reproduce**
I'm sorry, but I can't share a public URL where the bug can be observed.

1. Create multisite WordPress installation with at least two blogs.
2. Install and activate Yoast and ElasticPress.
3. Do a search on the main blog that will return a post from the second site as a first result.
4. The site name in the header will be of the second blog, not the main one.

**Expected behavior**
The expected behaviour would be the site name not to change because of the returned results.

**Environment information**
 - WordPress version: 5.6
 - Plugins and version: Yoast 15.6.2
 - Theme and version: custom theme, N/A
 
**Additional context**

My workaround was the following:
```
add_action( 'the_post', __NAMESPACE__ . '\\maybe_restore_current_blog', 11, 1 );
function maybe_restore_current_blog( $post ) {
	if ( in_the_loop() ) {
		return;
	}
	if ( is_multisite() && ! empty( $post->site_id ) ) {
		restore_current_blog();
	}
}
```

Maybe at the end of `QueryIntegration::maybe_switch_to_blog()` a similar check should be made:

```
if ( ! in_the_loop() ) {
	restore_current_blog();
}
```

felipeelia:
Hi @gonzomir,

By default, ElasticPress won't search on other sites unless the `sites` parameter is added to the WP_Query arguments. Do you mind sharing how you are doing that?

FWIW, I've tried adding it with the following snippet and things seem to work as expected:

```
add_action(
	'pre_get_posts',
	function ( $query ) {
		if ( ! is_admin() && $query->is_main_query() && $query->is_search() ) {
			$query->set( 'sites', 'all' );
		}
	}
);
```

gonzomir:
Hi, @felipeelia, 

[...]

PR #2144, opened 2021-03-20
Restore current blog when the_post triggers outside the loop.

gonzomir:
### Description of the Change

As described in #2121, in multisite environment and searching the whole network when `the_post` is triggered outside the loop, the current site is not restored if the first result is from another blog. To fix the issue I'm checking if the `WP_Query` instance is in the loop and if not, calling `restore_current_blog()`. To do this properly I had to add second argument to `ElasticPress\Indexable\Post\QueryIntegration->maybe_switch_to_blog()`.

### Alternate Designs

In the project where I first discovered the issue I added another function to `the_post` to restore the current blog, but that is not necessary if the issue can be addressed at it's roots.

### Benefits

Fix the issue, described in #2121.

### Possible Drawbacks

I can't find any.

### Verification Process

I tested that the issue dissappears in my test setup and made sure current tests are passing.

### Checklist:

### Applicable Issues

Issue #2121.

### Changelog Entry

Fix: Restore current blog when the_post triggers outside the loop in multisite environment and the whole network is searched if the first result is from another blog.


gonzomir:
I stand corrected, as I didn't run all tests, obviously my changes affect more than I expected...

felipeelia:
A few things we have to do before merging this one:
1. Add a default `null` value to `$query`. if that is `null`, then use the global query (just to be safe people aren't relying on this function using just one parameter)
2. Add another `@since` there explaining the introduction of the parameter
3. Looking at that function, we noticed that it can have some performance issues on multi-sites with several different sites. We should add another paragraph to the docblock stating that.

@gonzomir is this something you'd like to do? Otherwise, we can take over the PR and implement it (obviously giving you the credits for the original code.)

Thanks in advance!

gonzomir:
@felipeelia Yes, I can do that next week.

felipeelia:
@gonzomir we went ahead and took over the PR just adding those final details. I'm closing this one now and hopefully, the code will be part of our next release. Thank you very much!

Issue #2143, opened 2021-03-19
HTML for autosuggest items should be filterable 


mckdemps:
**Is your enhancement related to a problem? Please describe.**

Right now it's impossible to change the autosuggest HTML (for example, to add a thumbnail image). 

**Describe the solution you'd like**

Make the HTML filterable on these lines: https://github.com/10up/ElasticPress/blob/1c325e8959da4034d40e2fd6947eccfdac106ec4/assets/js/autosuggest.js#L255-L259

**Designs**

**Describe alternatives you've considered**

**Additional context**

mckdemps:
Related to #2128

Issue #2128, opened 2021-03-12
Adding more fields to the autosuggest results


fagiani:
I was looking at the JS code that builds the autosuggest list and found that only the `post_title` is set to `text` field of the formatted hits as seen below:

https://github.com/10up/ElasticPress/blob/1c325e8959da4034d40e2fd6947eccfdac106ec4/assets/js/autosuggest.js#L427-L443

Is there any way to override this method somehow to add other fields like the post date? 

Also, is it something that you'd be considering abstracting for a future version? I'd be interested in contributing with a PR if I had some general guidance.

felipeelia:
Hi @fagiani ,

Thanks for bringing that up! We'll discuss it internally and reply back to you here in a few days.

FWIW, in addition to that method, we also have to keep in mind the `updateAutosuggestBox` function (the one which actually builds the autosuggest HTML code). We already filter the data received from the Elasticsearch server using:

```
if (typeof window.epDataFilter !== 'undefined') {
	return window.epDataFilter(data, searchTerm);
}
```

So probably we could use a similar approach here.

fagiani:
Interestingly I've found that exactly piece of code earlier and actually worked to add the date to the results! :)

Thanks a lot!

PR #2146, opened 2021-03-23
Allow filtering the HTML of Autosuggest items.

JakePT:
### Description of the Change

Applies a `epAutosuggestItemHTMLFilter()` function to autosuggest items to allow filtering the HTML of autosuggest results.

### Alternate Designs

In this implementation the filter callback only receives the option object as the second argument, and the option index as the third argument. These can be used to reconstruct the original HTML, or a variation of it, if desired. This means that a custom filter would need to implement any required escaping, as well as highlighting.

An alternative approach could be to also pass the pre-processed escaped text and highlighted results text as arguments to the callback. 

The appropriate approach would depend on how much hand holding we want to do.

### Benefits

Developers can provide their own HTML for autosuggest result items.

### Possible Drawbacks

The potential for autosuggest results to break if future changes are made to the HTML of results that are not reflected in the HTML that may have been applied by a filter.

### Verification Process

- Check out this branch.
- Run `npm run build`.
- In a theme or plugin, define a `window.epAutosuggestItemHTMLFilter()` JavaScript function that returns some HTML.
- Start typing in a search field. The autosuggest results should use the HTML returned by our `epAutosuggestItemHTMLFilter()` function.

### Checklist:

### Applicable Issues

#2143
#2128 

### Changelog Entry

Added the ability to filter the HTML of autosuggest results by defining a `window.epAutosuggestItemHTMLFilter()` function in JavaScript.

### DIFF
--- a/assets/js/autosuggest.js
+++ b/assets/js/autosuggest.js
@@ -265,7 +265,7 @@ function updateAutosuggestBox(options, input) {
 			</li>`;
 
 		if (typeof window.epAutosuggestItemHTMLFilter !== 'undefined') {
-			itemHTML = window.epAutosuggestItemHTMLFilter(itemHTML, options[i]);
+			itemHTML = window.epAutosuggestItemHTMLFilter(itemHTML, options[i], i);
 		}
 
 		itemString += itemHTML;
--- a/assets/js/autosuggest.js
+++ b/assets/js/autosuggest.js
@@ -242,7 +242,8 @@ function updateAutosuggestBox(options, input) {
 	// create markup for list items
 	// eslint-disable-next-line
 	for ( i = 0; resultsLimit > i; ++i ) {
-		const { text, url } = options[i];
+		const text = options[i]._source.post_title;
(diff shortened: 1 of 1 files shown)


JakePT:
This implements a filter that allows changing the HTML of autosuggest results, however I’m not sure it’s enough to be able to get thumbnails in, which has been mentioned. This is because the results that are passed into the function that creates that HTML have been put through the `formatSearchResults()` function which returns an object that only contains the URL and result text, but does not have a filter of its own. If we want to support developers adding thumbnails to autosuggest results we may need a filter for that function too.

felipeelia:
@JakePT you are absolutely right. Initially, I thought `window.epDataFilter` would cover that part but we'll strip out any additional data in `formatSearchResults()` indeed. Can please you add it too?

JakePT:
@felipeelia I've updated the PR to move the basic functionality of `formatSearchResults()` into the `updateAutosuggestBox()` function so that the unformatted option is passed to the new filter.

PR #2171, opened 2021-04-15
Implement ID ranges instead of relying on OFFSET when doing pagination.

felipeelia:
This is a "take over" of #1955 just to address some minor things.

### Description of the Change

We noticed that indexing speed gets much slower than expected as the data set grows. We found the issue to be a combination of factors, but this PR is focused on an issue present in both versions of ElasticPress.

Essentially the change from using offsets to using ID ranges provides a very noticeable speed boost as the data set grows. We've seen significant gains when object IDs are extremely high.

Props @WPprodigy

### Alternate Designs

We didn't have any. Just needed some speed enhancements in short order.

### Benefits

It makes indexing faster at high object IDs.

### Possible Drawbacks

Adds additional CLI params.

### Verification Process

We ran the indexing commands on large multisites in excess of 10 million posts and noticed drastic increases in indexing speed compared to previous indexing operations.

Queries run without the PR(fetch 500 rows):
```
Offset 2500: 0.35 secs
Offset 25000: 1.54 secs
Offset 250000: 23.42 secs
Offset 2500000: 35.66 sec
```

Queries run with the PR are around 0.28 secs consistently regardless of the ID(fetch 500 rows)

### Checklist:

### Applicable Issues

Closes #1955 

### Changelog Entry

Added performance boost to querying posts from database. Props @WPprodigy


Rahmon (commented):
I did some tests and it's working. However, the option `--nobulk` is not working as expected and looks like it is ignored.

Other than that, I think `--start-object-id` and `--end-object-id` need more documentation because it is still confusing.

felipeelia:
Thanks @Rahmon. Can you share what you have in mind for:

> Other than that, I think --start-object-id and --end-object-id need more documentation because it is still confusing.

felipeelia:
@Rahmon did you fix the problem with the --nobulk parameter? Also, can you please fix the conflict with the Command.php file? Thanks

oscarssanchez:
Looks good to me. I agree we should document it better, as it took a few reads for me to understand what it does.

Rahmon (approved):

Issue #2168, opened 2021-04-14
Ability to override the autosuggest selector, not just append to it


johnbillion:
**Is your enhancement related to a problem? Please describe.**

We're using the autosuggest feature on a site. The `autosuggest.js` file that handles this [contains a hardcoded list of selectors to target](https://github.com/10up/ElasticPress/blob/9dfe8331d8a69bee0d964fc6815aecdee4c3b38d/assets/js/autosuggest.js#L388-L390), with the `epas.selector` option appended to the end. We'd like to override this list of selectors completely because some pages contain a search input (`input type="search"`) which should not be targeted for autosuggest because it is used for another purpose.

**Describe the solution you'd like**

I'd like another option added to [the options](https://github.com/10up/ElasticPress/blob/aa1d42fa202f804b74bec710dab46eceb8309353/includes/classes/Feature/Autosuggest/Autosuggest.php#L380-L397) which can be filtered with [the `ep_autosuggest_options ` filter](https://github.com/10up/ElasticPress/blob/aa1d42fa202f804b74bec710dab46eceb8309353/includes/classes/Feature/Autosuggest/Autosuggest.php#L427-L437) which overrides the list of selectors completely for more fine grained control.

Maybe this option could be called `selectors` and if it's specified then it overrides the hardcoded list + selector.

fagianijunior:
You read this issue? It will help you.
https://github.com/10up/ElasticPress/issues/2135

PR #2181, opened 2021-04-21
Move default autosuggest selectors definition so they can be filtered.

JakePT:
### Description of the Change

The autosuggest feature is automatically added to these selectors `.ep-autosuggest, input[type="search"], .search-field`. These are defined in the JavaScript file, and cannot be changed by the user. This PR moves the definition of the default selectors into the script localization so that they can be changed by filtering the `defaultSelectors` option with the existing `ep_autosuggest_options` filter.

It also adds a `ep_autosuggest_default_selectors` filter just for the default selectors.

### Alternate Designs

In #2168 @johnbillion suggested a `selectors` option, which could take precedence over the `selector` option if it was present. I opted to just put the default selectors into a `defaultSelectors` option to avoid potential ambiguity between `selector` and `selectors`, and possible confusion due to what would be the absence of an existing `selectors` option in the values that are being filtered.

Another approach could have been to implement the filter in JavaScript, since that's where the original value was, but I opted for putting them inside the PHP filter since that PHP filter was already established as being available to change some selector behaviour.

### Benefits

The most obvious use case would be to stop the plugin adding Autosuggest to all search inputs, by removing the `input[type="search"]` selector using:

```
add_filter(
	'ep_autosuggest_options',
	function($options) {
		$options['defaultSelectors'] = '.ep-autosuggest, .search-field';

		return $options;
	}
);
```
or

```
add_filter(
	'ep_autosuggest_default_selectors',
	function($selectors) {
		$selectors = '.ep-autosuggest, .search-field';

		return $selectors;
	}
);
```

The existence of an additional filter specifically for selectors enables removing default selectors with just:

```
add_filter( 'ep_autosuggest_default_selectors', '__return_empty_string' );
```

### Possible Drawbacks

In the event that a user has used the `ep_autosuggest_options` filter in such a way that does not pass through the `defaultSelectors` option, autosuggest would no longer be added to the default selectors.

### Verification Process

1. Enable Autosuggest and add a search widget. Do not enter anything into the selector setting.
2. Confirm that autosuggest is working.
3. Add code like this to remove the default selectors:
```
add_filter( 'ep_autosuggest_default_selectors', '__return_empty_string' );
```
4. As long as the theme does not add an `ep-autosuggest` class itself, no autosuggest functionality should be present in the search widget.

### Checklist:

### Applicable Issues

Addresses #2168.

### Changelog Entry

Allows changing or removing the default selectors used for autosuggest by filtering the `defaultSelectors` option in the  `ep_autosuggest_options` filter.


johnbillion:
I think this makes more sense than my original suggestion of adding a new option 👍

includes/classes/Feature/Autosuggest/Autosuggest.php
```diff
@@ -386,6 +386,14 @@ public function enqueue_scripts() {
...
 			'endpointUrl'         => esc_url( untrailingslashit( $endpoint_url ) ),
 			'selector'            => empty( $settings['autosuggest_selector'] ) ? 'ep-autosuggest' : esc_html( $settings['autosuggest_selector'] ),
+			/**
+			 * Filter autosuggest default selectors.
+			 *
+			 * @hook ep_autosuggest_default_selectors
+			 * @param {string} $selectors Default selectors used to attach autosuggest.
+			 * @return {string} Selectors used to attach autosuggest.
+			 */
+			'defaultSelectors'    => apply_filters( 'ep_autosuggest_default_selectors', '.ep-autosuggest, input[type="search"], .search-field' ),
```
felipeelia, line 396: @JakePT Can you please add a `@since 3.6.0` here, please? Thanks!

assets/js/autosuggest.js
```diff
@@ -385,9 +385,13 @@ function setFormIsLoading(isLoading, input) {
  * init method called if the epas endpoint is defined
  */
 function init() {
-	const epInputNodes = document.querySelectorAll(
-		`.ep-autosuggest, input[type="search"], .search-field, ${epas.selector}`,
-	);
+	const selectors = [epas.selector, epas.defaultSelectors].filter(Boolean).join(',');
```
felipeelia, line 388: @JakePT just to be sure, the order here isn't important? Having default selectors before the custom ones seems to make more sense to me but that is just a thought. I can't think in a real scenario where it really could be important.

JakePT: It shouldn't matter, but I agree having the defaults makes more intuitive sense, so I've just changed the order.

includes/classes/Feature/Autosuggest/Autosuggest.php
JakePT: Done.

Rahmon (approved):
LGTM :+1:

PR #2184, opened 2021-04-21
Add cache to the GH Actions

### LINKED ISSUE #2178: Fix Acceptance Tests
The title says it all :)

felipeelia:
This is an attempt to fix #2178


tests/wpa/features/ProtectedContentTest.php
```diff
@@ -31,11 +33,17 @@ public function testTurnProtectedContentFeatureOn() {
 
 		$I->click( 'a.button[data-feature="protected_content"]' );
 
+		$I->waitUntilElementVisible( '.pause-sync' );
```
Rahmon, line 36: I'm not sure why it is necessary.

tests/wpa/inc/TestBase.php
```diff
@@ -290,38 +307,9 @@ protected function deactivatePlugin( $actor, $slug = 'elasticpress', $network =
...
-		$actor->typeInField( '#user_login', $data['user_login'] );
-
-		$actor->typeInField( '#email', $data['user_email'] );
-
-		$actor->checkOptions( '#noconfirmation' );
-
-		$actor->click( '#createusersub' );
-
-		$actor->waitUntilElementVisible( '#message' );
+	protected function isElasticPressIo( $actor = null ) {
```
Rahmon, line 310: We can remove the $actor parameter, right?

tests/wpa/features/ProtectedContentTest.php
felipeelia: @Rahmon as Protected Content fires a reindex, it is here just to make sure WPA waits until that happens, otherwise, the `$I->seeText( 'Sync complete', '.sync-status' );` test will fail. Some weirdness I've seen in the test runs before.

tests/wpa/inc/TestBase.php
felipeelia: Good catch, I've fixed that in 175397b. Thanks!

Rahmon (approved):
LGTM

Issue #1519, opened 2019-10-10
Filtering REST API queries while requesting with a slug parameter


jayhill90:
**Describe the bug**
When filtering REST API post type queries with `$args['ep_integrate'] = true;` for a post type, you can no longer make a request with `slug=post-name` as a query parameter. You get back the 10 most recent posts of that post type. 

Unfortunately, since this is only observed while querying the REST API, I haven't found a way to be able to inspect the queries being generated in EP or what that looks like on the ES side of things. 

**Steps to Reproduce**
Filter a specific post type query, `rest_post_query` or `rest_CUSTOMPOSTTYPE_query` and prior to returning `$args` set `$args['ep_integrate'] = true`.
Make a request to `/wp-json/wp/v2/POSTTYPE?slug=APOSTSLUG` You'll get back the latest 10 posts of that post type.
Remove `$args['ep_integrate'] = true` and make the same request. You should get back the specific requested data.

**Expected behavior**
The only post(s) we should receive back when passing in ?slug= to the REST API are the specific posts that correspond to the passed in post_name__in array. 

**Environment information**
 - Device: MacBook Pro
 - OS: OS X, Linux, Doesn't matter
 - Browser and version: Postman
 - Plugins and version: ElasticPress 3.1.4 
 - Theme and version: Custom Build
 - Other installed plugin(s) and version(s): N/A

**Additional context**

I tested this with a few different `rest_POSTTYPE_query` filters for different post types and it only works when I remove the code to integrate the query into ElasticPress. 

Maybe this isn't an actual bug on the EP side and I'm just lost in a rabbit hole and this is all for naught. 

The other thought i had was that maybe ElasticSearch just doesn't know what to do with the `post_name__in` argument and how to map that as there's no slug value in ES when querying through Kibana, but there is `post_name`. Is it maybe possible that maybe EPdoesn't have a way to associate `post_name__in` to `post_name` in ES? 

Thanks!

tlovett1:
Currently EP doesn't support `post_name__in` which is used when passing `slug`. We can probably add this to the next version.

tlovett1:
For the time being just don't ep_integrate if post_name__in is used.

jayhill90:
Thanks Taylor.

oscarssanchez:
Estimating 2 hours for this.

PR #2194, opened 2021-05-04
Add post__name_in support

oscarssanchez:
### Description of the Change

Hi @felipeelia , can you please review this PR implementing https://github.com/10up/ElasticPress/issues/1519#issuecomment-541488811 ?

### Benefits

Adds support for `post__name_in` 

### Verification Process

* Unit tests
* Tested it worked with a normal `WP_Query` object and using the `pre_get_posts` hook

### Checklist:


includes/classes/Indexable/Post/Post.php
```diff
@@ -870,6 +870,21 @@ function( $tax_query ) use ( $args ) {
...
 
+		/**
+		 * 'post_name__in' arg support.
+		 *
+		 * @since 3.6.0
+		 */
+		if ( ! empty( $args['post_name__in'] ) ) {
+			$filter['bool']['must'][]['bool']['must'] = array(
+				'term' => array(
+					'post_name.raw' => $args['post_name__in'],
```
felipeelia, line 881: @oscarssanchez I'm afraid that if `post_name__in` is an associative array, it will become an object (rather than an array) in JSON, and ES will fail. Can you please change
`'post_name.raw' => $args['post_name__in'],`
to
`'post_name.raw' => array_values( (array) $args['post_name__in'] ),`
Thanks!

felipeelia (changes requested):

includes/classes/Indexable/Post/Post.php
oscarssanchez: Makes sense @felipeelia !

tests/php/indexables/TestPost.php
```diff
@@ -3682,6 +3682,35 @@ public function testPostParentQuery() {
...
+			)
+		);
+		Functions\create_and_sync_post( array( 'post_content' => 'findme name in test 2' ) );
+		Functions\create_and_sync_post( array( 'post_content' => 'findme name in test 3' ) );
+
+		ElasticPress\Elasticsearch::factory()->refresh_indices();
+
+		$args = array(
+			's'             => 'findme name in',
+			'post_name__in' => 'findme-name-in',
```
Rahmon, line 3705: @oscarssanchez I would add a test passing an array in `post_name__in` since this parameter [expects an array](https://developer.wordpress.org/reference/classes/wp_query/#post-page-parameters). Just to make sure it's working in this case.

Rahmon (changes requested):

tests/php/indexables/TestPost.php
oscarssanchez: Thanks @Rahmon I have added tests for this.

Rahmon (approved):
LGTM :+1:

felipeelia (approved):

PR #2197, opened 2021-05-05
Filter to disable SQL fallback for searches

cbratschi:
### Description of the Change
So far ElasticPress fallbacks to SQL in case of ElasticSearch timeouts or errors. This fallback can slowdown a WordPress instance with lots of posts. Therefore a filter was added to prevent this fallback (opt-in).

### Benefits
Slow SQL search operations are never performed on the server if the filter is enabled.

### Possible Drawbacks
None, if the modified code is the only source for SQL fallbacks. During re-indexing the SWL fallback should not happen either.

### Verification Process
The code is inactive by default until the filter is being used. After using the filter all slow SQL calls no longer occured.

### Checklist:

### Changelog Entry
Added ep_fallback_to_mysql filter to dsable SQL search fallback.

### DIFF
--- a/includes/classes/Indexable/Post/QueryIntegration.php
+++ b/includes/classes/Indexable/Post/QueryIntegration.php
@@ -299,6 +299,25 @@ class QueryIntegration {
 			 */
 			if ( false === $ep_query ) {
 				$query->elasticsearch_success = false;
+
+				/**
+				 * Filter whether to fallback to MySQL or not
+				 *
+				 * @hook ep_fallback_to_mysql
+				 * @param  {boolean} $show True to fallback
+				 * @return {boolean} New value
+				 */
+				if ( apply_filters( 'ep_fallback_to_mysql', true ) ) {
+					//perform SQL operation
+					return null;
+				}
+
+				//never fallback to MySQL for full table searches
+				if ( ! empty( $query_vars['s'] ) ) {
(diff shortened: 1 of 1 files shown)


felipeelia:
Hi @cbratschi ,

Although we appreciate the Pull Request, after discussing it internally we realized that it can be achieved with the following snippet:

```
function ep_custom_avoid_fallback_mysql( $posts, $query ) {
	if ( isset( $query->elasticsearch_success ) && false === $query->elasticsearch_success ) {
		return [];
	}
	return $posts;
}
add_filter( 'posts_pre_query', 'ep_custom_avoid_fallback_mysql', 11, 2 );
```

As developers would need to add some code anyway to activate the bypass, that snippet should be enough for those trying to avoid falling back to MySQL.

With that said, I'm closing the PR but I want to thank you once more for opening it.

cbratschi:
@felipeelia thanks for the code snippet. Yes, this should do the same. Do you know if this covers the indexing case too?

felipeelia:
Sorry @cbratschi , what would be the fallback in the indexing case?

cbratschi:
@felipeelia also while indexing SQL search should not be used. I am not sure what ElasticPress does in such cases.

cbratschi:
After updating to the latest ElasticPress version we had to run re-indexing. I can confirm in such cases the SQL search was still used, this means this workaround is only for ElasticSearch timeouts but not other cases.

PR #2210, opened 2021-05-21
Added Elasticsearch::get_documents and Indexable::multi_get

felipeelia:
### Description of the Change

This is a "take over" of #1959

> We needed an easy way to fetch multiple documents per API call, so we added get_documents to the Elasticsearch class and then added multi_get to the Indexable class for ease of use.
> 
> Props @nickdaugherty
> cc: @rinatkhaziev @pschoffer @netsuso @parkcityj

### Alternate Designs

> N/A

### Benefits

> Allows you to get multiple documents per API call.

### Possible Drawbacks

> N/A

### Checklist:

### Changelog Entry

Added Elasticsearch::get_documents and Indexable::multi_get. Props @nickdaugherty and @felipeelia


includes/classes/Elasticsearch.php
```diff
@@ -650,6 +650,56 @@ public function delete_network_alias( $alias ) {
...
+
+		$docs = [];
+
+		if ( is_array( $response['docs'] ) ) {
+			foreach ( $response['docs'] as $doc ) {
+				if ( ! empty( $doc['exists'] ) || ! empty( $doc['found'] ) ) {
+					$docs[] = $doc['_source'];
+				} else {
+					$docs[] = null;
+				}
```
Rahmon, line 696: @felipeelia What do you think about using the ID as an index in the $docs[]?

Now if we don't find a doc, the returned array looks like `[ 0 => doc1, 1 => null, 2 => doc3 ]`. Using the ID as an index the array would like `[ 11 => doc1, 25 => doc 3 ]` and we can check if a doc was found checking the index value.

felipeelia: makes sense @Rahmon . I've also added a new filter there so people can change it before it is returned.

Rahmon (approved):
LGTM :+1:

felipeelia:
@oscarssanchez as all tests have passed, can you please give this a final review and merge it? Thanks!

oscarssanchez (approved):

PR #2214, opened 2021-05-24
Updated prepare_date_terms to only call date_i18n once

Rahmon:
This is a "take over" of #1958  just to address some minor things.

### Description of the Change

Calling `date_i18n` a lot gets pretty expensive. Only calling it once for each `prepare_date_terms` call is much more efficient and make indexing faster.

Props @WPprodigy

### Alternate Designs

N/A

### Benefits

Removes extraneous calls to `date_i18n` and makes indexing faster.

### Possible Drawbacks

N/A

### Verification Process

There are 4 calls to `prepare_date_terms()` per post being indexed. Each call resulted in 12 calls to `date_i18n()`, which ends up going down a bit of a rabbit hole and eventually adding up to a lot of time spent doing filters and alloption calls. So per page of 500 posts being indexed, that is 24,000 calls (500 x 12*4)- seen in cachegrind.

Before the changes, a test group of 500 posts took 7.91601 seconds running prepare_date(). After this PR, 0.85888 seconds.

### Checklist:

### Applicable Issues

Closes: #1958 

### Changelog Entry

Changed prepare_date_terms to only call date_i18n once. Props @WPprodigy


includes/classes/Indexable/Post/Post.php
```diff
@@ -336,25 +336,36 @@ public function prepare_document( $post_id ) {
...
-			'dayofyear'     => (int) date_i18n( 'z', $timestamp ),
-			'day'           => (int) date_i18n( 'd', $timestamp ),
-			'dayofweek'     => (int) date_i18n( 'w', $timestamp ),
-			'dayofweek_iso' => (int) date_i18n( 'N', $timestamp ),
-			'hour'          => (int) date_i18n( 'H', $timestamp ),
-			'minute'        => (int) date_i18n( 'i', $timestamp ),
-			'second'        => (int) date_i18n( 's', $timestamp ),
-			'm'             => (int) ( date_i18n( 'Y', $timestamp ) . date_i18n( 'm', $timestamp ) ), // yearmonth
-		);
+	private function prepare_date_terms( $date_to_prepare ) {
```
felipeelia, line 343: @Rahmon there is no reason to keep this method private. Let's make it public, as it doesn't change any instance attribute...

includes/classes/Indexable/Post/Post.php
```diff
@@ -336,25 +336,36 @@ public function prepare_document( $post_id ) {
...
+			'dayofweek_iso' => 'N',
+			'hour'          => 'H',
+			'minute'        => 'i',
+			'second'        => 's',
+			'm'             => 'Ym', // yearmonth
+		];
+
+		// Combine all the date term formats and perform one single call to date_i18n() for performance.
+		$date_format    = implode( '||', array_values( $terms_to_prepare ) );
+		$combined_dates = explode( '||', date_i18n( $date_format, strtotime( $date_to_prepare ) ) );
```
felipeelia, line 360: heh, this was clever!

tests/php/indexables/TestPost.php
```diff
@@ -65,6 +65,21 @@ public function tearDown() {
...
 	}
 
+	/**
+	 * Get private function to be tested.
+	 *
+	 * @param string $functionName
+	 * @param string $className
+	 * @return ReflectionMethod.
+	 */
+	protected function get_private_function( $functionName, $className = '\ElasticPress\Indexable\Post\Post' ) {
```
felipeelia, line 75: making `prepare_date_terms` public will be good to get rid of this as well. It is a nice piece of code, though.

felipeelia (changes requested):

felipeelia:
@Rahmon I forgot to mention but there are some warnings in PHPCS that should be addressed as well. Thanks!

includes/classes/Indexable/Post/Post.php
WPprodigy: I will note that private is safer long-term, since it ensures you can make changes to it in the future and only have to account for the usage in the class itself, and not track down if anything else is relying / extending on it. Once public, you have to permanently keep backwards compatibility.

tests/php/indexables/TestPost.php
```diff
@@ -5916,4 +5931,63 @@ public function testPostSyncQueuePermissions() {
...
+		$this->assertIsArray( $return_prepare_date_terms );
+
+		$this->assertArrayHasKey( 'year', $return_prepare_date_terms );
+		$this->assertEquals( '2021', $return_prepare_date_terms['year'] );
+
+		$this->assertArrayHasKey( 'month', $return_prepare_date_terms );
+		$this->assertEquals( '04', $return_prepare_date_terms['month'] );
+
+		$this->assertArrayHasKey( 'week', $return_prepare_date_terms );
+		$this->assertArrayHasKey( 'dayofyear', $return_prepare_date_terms );
```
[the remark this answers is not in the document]
Rahmon: Added

includes/classes/Indexable/Post/Post.php
felipeelia: Thanks for weighing in @WPprodigy. Although I partially agree with you:
1. This entire class is probably one of the most extended in custom solutions that involve new Indexables, so just because of that, we would have enough reasons to make it at least protected.
2. From an OOP perspective, this is just a "utility" function, receiving one input and returning an output without modifying the instance itself, making things _easier_ to track down.
3. If we decide to make any change in the way the plugin creates those date_terms, people would be relying on the indexed values anyway.

felipeelia (approved):

PR #2218, opened 2021-05-26
Facet filters: Don't remove spaces from value

### LINKED ISSUE #2217: BUG: Facet filters don't work with term names containing spaces
**Describe the bug**

As we are [removing white spaces from the value typed by the user](https://github.com/10up/ElasticPress/blob/develop/assets/js/facets.js#L13) in Facets filters, it won't match term names that contain spaces.

**Steps to Reproduce**
1. Create a category called "New York" and assign a post to it
2. Create a facets widget with more than 15 terms, so filtering is enabled
3. Type "New Y"
4. See the term won't be displayed

**Expected behavior**
The term should be displayed

**Screenshots**

facet-filter

felipeelia:
### Description of the Change

As we don't remove spaces from the term name while generating the data-term-name attribute (see [here](https://github.com/10up/ElasticPress/blob/3.5.6/includes/classes/Feature/Facets/Widget.php#L188)), there is no point in removing it from the user inputted value.

### Verification Process

1. Create a category called "New York" and assign a post to it
2. Create a facets widget with more than 15 terms, so filtering is enabled
3. Type "New Y"
4. See the term there

### Applicable Issues

Closes #2217

### Changelog Entry

- Fix an error in facets filters where terms wouldn't match if the user types a space


oscarssanchez (approved):
LGTM!

Issue #2206, opened 2021-05-16
BUG: When Plugin enabled, search in Google Web Stories plugin is broken


olegcarguy:
**Describe the bug**
When creating a new web story using the Google Web Stories plugin, the plugin has a search field where the admin can search the media library for photos.  The search breaks and returns zero results from the media library.

**Steps to Reproduce**

1. Enable ElasticSearch plugin
2. Create a new web story via this plugin: https://wordpress.org/plugins/web-stories/
3. Search for a photo to import from the media library using a search term
4. See no results

**Expected behavior**
Correct results from media library search

**Screenshots**
GIF showing problem:
![QdMKgh74tX](https://user-images.githubusercontent.com/29559821/118386614-54806580-b5de-11eb-8f95-f2e1c1a6bacb.gif)
GIF showing the same search term showing results without the ElasticSearch plugin enabled:
![PwtBzDstcO](https://user-images.githubusercontent.com/29559821/118386739-3cf5ac80-b5df-11eb-8a2a-36a082bb694a.gif)

**Environment information**
WordPress version: 5.6.1
 
 - ElasticPress version: 3.5.6
 - Elasticsearch version: 7.9.0
 - Where do you host your Elasticsearch server? self-hosted
 - What ElasticPress features do you have currently active? Post Search, Users, Terms

felipeelia:
There is a chance it is related to https://github.com/10up/ElasticPress/issues/1604

felipeelia:
Hi @olegcarguy, now that #2222 is merged, can you please check if you are still able to reproduce the problem while using the code in the `develop` branch? Thanks.

landbryo:
This issue still exists. In the case that users don't want to enable Protected Content as a workaround, this seems to do the trick:
```php
add_filter( 'ep_is_integrated_request', function ( $is_integrated ) {
	if ( ! empty( $_REQUEST['search'] ) && ! empty( $_REQUEST['_web_stories_envelope'] ) ) {
		$is_integrated = false;
	}

	return $is_integrated;
}, 20 );
```
The Web Stories media search request looks like:
`http://localhost/wp-json/web-stories/v1/media/?per_page=50&page=1&_web_stories_envelope=true&search=feature&_locale=user`

PR #2222, opened 2021-05-31
Fix post_mime_type argument

Rahmon:
### Description of the Change

This change updates the plugin to handle `post_mime_type` argument when it has an array with a MIME type without the subtype i.e `image` instead of `image/png`.

### Benefits

It's possible to search images when using the Gutenberg editor or Web Stories plugin.

### Verification Process

1. Enable document indexing
2. Enable admin and AJAX via `ep_admin_wp_query_integration` and `ep_ajax_wp_query_integration`
3. Edit a post with the block editor
4. Add a new image block to bring up the media modal
5. Use the search bar to search for an image
6. See the image

### Checklist:

### Applicable Issues

Closes: #1604 #2206

### Changelog Entry

Fixed post_mime_type argument issue when it has an array with a MIME type without the subtype

### DIFF
--- a/includes/classes/Indexable/Post/Post.php
+++ b/includes/classes/Indexable/Post/Post.php
@@ -998,7 +998,10 @@ class Post extends Indexable {
 				$args_post_mime_type = [];
 
 				foreach ( $args['post_mime_type'] as $mime_type ) {
-					// check if matches the MIME type pattern: type/subtype
+					/**
+					 * check if matches the MIME type pattern: type/subtype and
+					 * leave an empty string as posts, pages and CPTs don't have a MIME type
+					 */
 					if ( preg_match( '/^[-._a-z0-9]+\/[-._a-z0-9]+$/i', $mime_type ) || empty( $mime_type ) ) {
 						$args_post_mime_type[] = $mime_type;
 					} else {
--- a/includes/classes/Indexable/Post/Post.php
+++ b/includes/classes/Indexable/Post/Post.php
@@ -999,7 +999,7 @@ class Post extends Indexable {
 
 				foreach ( $args['post_mime_type'] as $mime_type ) {
 					// check if matches the MIME type pattern: type/subtype
-					if ( preg_match( '/^[-._a-z0-9]+\/[-._a-z0-9]+$/i', $mime_type ) ) {
+					if ( preg_match( '/^[-._a-z0-9]+\/[-._a-z0-9]+$/i', $mime_type ) || empty( $mime_type ) ) {
 						$args_post_mime_type[] = $mime_type;
 					} else {
 						$filtered_mime_type_by_type = wp_match_mime_types( $mime_type, wp_get_mime_types() );
--- a/includes/classes/Indexable/Post/Post.php
+++ b/includes/classes/Indexable/Post/Post.php
@@ -994,9 +994,23 @@ class Post extends Indexable {
 		 */
 		if ( ! empty( $args['post_mime_type'] ) ) {
 			if ( is_array( $args['post_mime_type'] ) ) {
+
+				$args_post_mime_type = [];
+
+				foreach ( $args['post_mime_type'] as $mime_type ) {
+					// check if matches the MIME type pattern: type/subtype
+					if ( preg_match( '/^[-._a-z0-9]+\/[-._a-z0-9]+$/i', $mime_type ) ) {
+						$args_post_mime_type[] = $mime_type;
+					} else {
+						$filtered_mime_type_by_type = wp_match_mime_types( $mime_type, wp_get_mime_types() );
+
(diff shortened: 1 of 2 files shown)


includes/classes/Indexable/Post/Post.php
```diff
@@ -994,9 +994,23 @@ function( $tax_query ) use ( $args ) {
 		 */
 		if ( ! empty( $args['post_mime_type'] ) ) {
 			if ( is_array( $args['post_mime_type'] ) ) {
+
+				$args_post_mime_type = [];
+
+				foreach ( $args['post_mime_type'] as $mime_type ) {
+					// check if matches the MIME type pattern: type/subtype
+					if ( preg_match( '/^[-._a-z0-9]+\/[-._a-z0-9]+$/i', $mime_type ) || empty( $mime_type ) ) {
```
felipeelia, line 1002: @Rahmon shouldn't we skip it if the `$mime_type` is empty?

tests/php/indexables/TestPost.php
```diff
@@ -3997,7 +4010,51 @@ public function testPostMimeTypeQuery() {
...
+		$this->assertContains( get_post_field( 'post_name', $attachment_id_1_jpeg ), $attachment_names );
+		$this->assertContains( get_post_field( 'post_name', $attachment_id_2_jpeg ), $attachment_names );
+		$this->assertContains( get_post_field( 'post_name', $attachment_id_3_pdf ), $attachment_names );
+		$this->assertContains( get_post_field( 'post_name', $attachment_id_4_png ), $attachment_names );
+
+		$args = array(
+			'ep_integrate'   => true,
+			'post_mime_type' => array(
+				'image/png',
+			),
```
felipeelia, line 4046: @Rahmon Can we also test when the dev sends a query with `post_mime_type` as a string?

includes/classes/Indexable/Post/Post.php
Rahmon: @felipeelia No. An empty MIME type is used when you're searching through posts and attachments. Hopefully, we have a test with this case because I didn't know either: https://github.com/10up/ElasticPress/blob/de9d2aeed1c77ec3996e36ef6809fcad3ae52d3e/tests/php/features/TestDocuments.php#L184

tests/php/indexables/TestPost.php
Rahmon: @felipeelia We already have a test for this case in the previous line: https://github.com/10up/ElasticPress/blob/de9d2aeed1c77ec3996e36ef6809fcad3ae52d3e/tests/php/indexables/TestPost.php#L3979

includes/classes/Indexable/Post/Post.php
felipeelia: @Rahmon sorry, I'm not sure I follow. How that `'post_type' => array( 'post', 'attachment' ),` is related to having something like this in the ES query:
```
'terms' => array(
	'post_mime_type' => [
		'image/jpeg',
		'',
	],
),
```

tests/php/indexables/TestPost.php
felipeelia: @Rahmon got it. I wonder if we shouldn't have `'post_mime_type' => 'image/jpeg', ` as well, just to keep track of how it behaves when the ES query becomes something like:
```
'regexp' => array(
	'post_mime_type' => 'image/jpeg.*',
),
```
(it seems to be what would happen [here](https://github.com/10up/ElasticPress/blob/c818a4769781ac5e0517166221b0a17439ac3a4e/includes/classes/Indexable/Post/Post.php#L1018)).

includes/classes/Indexable/Post/Post.php
Rahmon: @felipeelia Sorry for the inaccurate answer. When you have the Document feature enabled, the EP adds an empty string as a MIME type to allow querying non-attachments and attachments. So we need to keep the empty string. 

https://github.com/10up/ElasticPress/blob/de9d2aeed1c77ec3996e36ef6809fcad3ae52d3e/includes/classes/Feature/Documents/Documents.php#L169

felipeelia: Ah, now I got it, thanks @Rahmon! If you are touching that line again, do you mind adding a comment there?
```
// check if matches the MIME type pattern: type/subtype
// Leave an empty string as posts, pages, and CPTs don't have a post mime type
```
Something like that :)

Rahmon: Sure. Since this is counterintuitive is better to have a comment.

Thanks :pray:

Rahmon: @felipeelia comment updated.

tests/php/indexables/TestPost.php
Rahmon: @felipeelia I've added this case.

felipeelia (approved):

Issue #2225, opened 2021-06-01
BUG: the search fails when trying to find a user in a block


Rahmon:
**Describe the bug**

The search fails when trying to find a user in a block with the error:

> Fielddata is disabled on text fields by default. Set fielddata=true on [display_name] in order to load fielddata in memory by uninverting the inverted index. Note that this can however use significant memory. Alternatively use a keyword field instead.

To fix it we need to update to use `display_name.raw` instead of `display_name` to sort the result.

**Steps to Reproduce**

1. Enable Users feature
2. Go to a page/post and add a paragraph block
3. Type `@` and try to search a user e.g. `@admin`
4. The result will appear but the search is made using MySQL and not Elasticsearch.

**Expected behavior**
Using Elasticsearch to search for a user.

**Screenshots**

**Environment information**
 - Device: Dell
 - OS: Linux
 - Browser and version: Chrome
 - WordPress version: 5.7.2
 
 - ElasticPress version: 3.5.6
 - Elasticsearch version: 5.6.16
 - Where do you host your Elasticsearch server? Local
 - What ElasticPress features do you have currently active? Users
 - Plugins and version: 
 - Theme and version: 
 
 - <details><summary>Site Health Info:</summary>
   
   </details>

**Additional context**

PR #2226, opened 2021-06-01
Fix WP_Query_User when it tries to sort the results

Rahmon:
### Description of the Change

This changes updates to use keyword fields to sort in the WP_User_Query and adds tests to ensure we are using these fields.

### Benefits

Elasticsearch will not fail and fall back to MySQL.

### Verification Process

1. Enable Users feature
2. Go to a page/post and add a paragraph block
3. Type `@` and try to search a user e.g. `@admin`
4. Check if the results were provided by Elasticsearch.

### Checklist:

### Applicable Issues

Closes: #2225 

### Changelog Entry

Fixed sort for WP_User_Query


tests/php/indexables/TestUser.php
```diff
@@ -607,11 +612,259 @@ public function testUserQueryOrderbyDisplayName() {
...
+		$users_display_name_fetched = wp_list_pluck( $user_query->results, 'display_name' );
+
+		/**
+		 * Check if Zoey is the first user because Elasticsearch sort
+		 * is case-sensitive
+		 *
+		 * Ref: https://www.elastic.co/guide/en/elasticsearch/guide/2.x/sorting-collations.html#case-insensitive-sorting
+		 *
+		 */
+		$this->assertEquals( 'Zoey', $users_display_name_fetched[0] );
```
felipeelia, line 632: @Rahmon it seems to be counterintuitive to rely on a case-sensitive sorting while WP's is case insensitive. Also, for fields in the posts indexable we have [a "sortable" field](https://github.com/10up/ElasticPress/blob/develop/includes/mappings/post/7-0.php#L265) that we could replicate here. Do you mind taking care of that?

felipeelia (changes requested):

tests/php/indexables/TestUser.php
Rahmon: Sure. Thanks for that @felipeelia

felipeelia (dismissed):

felipeelia:
@Rahmon did you test it with Elasticsearch 5.6.16 (the version shipped with WP Local Docker)? I'm using the following snippet in theme's header.php:

```
new WP_User_Query(
	[
		'search'  => 'test',
		'orderby' => 'display_name',
	]
);
```

As is, I receive a 400 error saying:
> Fielddata is disabled on text fields by default. Set fielddata=true on [display_name.sortable] in order to load fielddata in memory by uninverting the inverted index. Note that this can however use significant memory. Alternatively use a keyword field instead.

If I change `wp-content/plugins/elasticpress/includes/mappings/user/initial.php` to look like this, it works:

```
'display_name'    => array(
	'type'   => 'text',
	'fields' => array(
		'raw'      => array(
			'type'                => 'keyword',
			'ignore_above' => 10922,
		),
		'sortable' => array(
			'type'         => 'keyword',
			'ignore_above' => 10922,
			'normalizer'   => 'lowerasciinormalizer',
		),
	),
),
```

Can you please check it? Thanks

Rahmon:
> @Rahmon did you test it with Elasticsearch 5.6.16 (the version shipped with WP Local Docker)? I'm using the following snippet in theme's header.php:
> 
> ```
> new WP_User_Query(
> 	[
> 		'search'  => 'test',
> 		'orderby' => 'display_name',
> 	]
> );
> ```
> 
> As is, I receive a 400 error saying:
> 
> > Fielddata is disabled on text fields by default. Set fielddata=true on [display_name.sortable] in order to load fielddata in memory by uninverting the inverted index. Note that this can however use significant memory. Alternatively use a keyword field instead.
> 
> If I change `wp-content/plugins/elasticpress/includes/mappings/user/initial.php` to look like this, it works:
> 
> ```
> 'display_name'    => array(
> 	'type'   => 'text',
> 	'fields' => array(
> 		'raw'      => array(
> 			'type'                => 'keyword',
> 			'ignore_above' => 10922,
> 		),
> 		'sortable' => array(
> 			'type'         => 'keyword',
> 			'ignore_above' => 10922,
> 			'normalizer'   => 'lowerasciinormalizer',
> 		),
> 	),
> ),
> ```
> 
> Can you please check it? Thanks

@felipeelia I've updated the mapping and updated some tests to check if the query uses the Elasticsearch to retrieve the users.

felipeelia (approved):

PR #2229, opened 2021-06-07
Fix/facets pagination

### LINKED ISSUE #2201: BUG: Facet Filter
I set up a few facet filters to filter product attributes, however they seem to be returning the wrong results. I filter for a certain manufacturer, and it displays many different manufacturers, including that filter. This happens when I'm on a Product Category page, and I select one attribute filter.

I have the facets set to "Show all content tagged to **any** selected terms".

If I were to set the facets to "Show all content tagged to **all** selected terms", then the facets will work properly, however I will not be able to "grow" the filtered content by selecting more than one option from a single facet.

One more issue, e.g. If I am on the 50th paginated page of a category, and then filter by an attribute that will only have 5 pages of products, I will get a 404, because the filter doesn't clear the pagination.

Any workarounds to these issues?

Thank you!

oscarssanchez:
### Description of the Change

Fixes #2201 

### Verification Process

1.- Go to an archive
2.- Facets should remove pagination on all scenarios

### Checklist:


Rahmon:
@oscarssanchez I've updated the PR and added some tests. Can you please take a look at the tests and add new ones if you think it's necessary?

oscarssanchez:
It seems the tests are failing because the latest commit will add a slash before the `?` sign @Rahmon , if you don't see any issues with it, I will adjust the test for it to pass.

Rahmon:
> It seems the tests are failing because the latest commit will add a slash before the `?` sign @Rahmon , if you don't see any issues with it, I will adjust the test for it to pass.

Yes, @oscarssanchez . Feel free to adjust and maybe add new tests :relaxed:

Rahmon (approved):

PR #2230, opened 2021-06-09
Adds check for term before capability check

msawicki:
### Description of the Change

The `action_sync_on_object_update` method that's hooked into `set_object_terms` needs to be able to handle an array of term slugs. The reason being is that `wp_set_object_terms()`'s second parameter can be `string|int|array`. See [docs](https://developer.wordpress.org/reference/functions/wp_set_object_terms). When a string is passed, `wp_set_object_terms` wraps it in an array, thus becoming an array of slugs that's later passed to `set_object_terms`. Using the `term_exists()` function here provides some flexibility since it accepts ID, slug, or name. Without this check, when a slug is passed into `wp_set_object_terms()` (Polylang does this when setting a language) it will throw a php Notice "Trying to get property 'term_id' of non-object" in SyncManager.php:110. This can be replicated with Polylang and ElasticPress installed with the Terms Feature enabled. From there, creating a new post will give you a php notice noted above. 

I don't think there is a current issue files for this but please let me know if I'm incorrect and I'll link the two.

### Alternate Designs

NA

### Benefits

No php Notices

### Possible Drawbacks

Extra database queries in the `term_exists()` function but I think this a cleaner approach than various type checks.

### Verification Process

I made this change in my local environment and ran several check/tests updating and adding new post content. 

### Checklist:

### Applicable Issues

NA

### Changelog Entry

Added a check for term exists to get the term id by slug, id or name before a capabilities check is done.


felipeelia:
Hi @msawicki, first of all, thank you very much for the Pull Request! To make the code comply with our standards, do you mind changing
`if( ! $term_info ) {`
to
`if ( ! $term_info ) {`
?
We've also merged a few things recently, so it would be great if you could update your branch as well.

Thanks!

msawicki:
@felipeelia Updated! Thanks for the feedback.

felipeelia (approved):

Issue #2262, opened 2021-07-12
Migrate to 10up Toolkit


felipeelia:
**Is your enhancement related to a problem? Please describe.**

Currently, our building scripts rely on several different parts: Babel, Webpack, Post CSS, etc. That makes things harder to keep node packages updated and compatible. 

**Describe the solution you'd like**

[10up Toolkit](https://github.com/10up/10up-toolkit) would allow us to simplify those scripts. It would also drastically reduce the number of packages we require on our package.json file. The final number of packages installed may or may not be lower, but transferring that responsibility to 10up-toolkit would make our lives easier.

PR #2279, opened 2021-07-27
Migrate to 10up Toolkit

Rahmon:
### Description of the Change

This change migrates to 10up Toolkit.

### Benefits

 Transferring the building responsibility to 10up-toolkit.

### Checklist:

### Applicable Issues

Closes: #2262


package.json
```diff
@@ -4,71 +4,32 @@
...
-    "eslint-config-airbnb-base": "^14.2.1",
-    "eslint-config-prettier": "^7.2.0",
-    "eslint-loader": "^2.1.2",
-    "eslint-plugin-import": "^2.22.1",
-    "eslint-plugin-jsdoc": "^31.6.1",
-    "eslint-plugin-jsx-a11y": "^6.4.1",
-    "eslint-plugin-prettier": "^3.3.1",
     "eslint-plugin-react": "^7.22.0",
     "eslint-plugin-react-hooks": "^1.7.0",
-    "jsdoc": "^3.6.7",
```
felipeelia, line 39: @Rahmon if you are removing `jsdoc` here, are you sure the `build:docs` command will continue to work? It is used here: https://github.com/10up/ElasticPress/blob/develop/.github/workflows/build-docs.yml

felipeelia (approved):

.eslintrc.js
```diff
@@ -0,0 +1,28 @@
+const defaultEslintrc = require('10up-toolkit/config/.eslintrc');
+
+module.exports = {
+	...defaultEslintrc,
+	extends: '@10up/eslint-config',
```
nicholasio, line 5: you should probably extend `@10up/eslint-config/react` if you need react linting.

.babelrc
```diff
@@ -1,8 +1,6 @@
 {
   "presets": [
```
nicholasio, line 2: I don't think this file would be required anymore unless the babel config needs to be shared with other tools. The default babel preset should work out of the box but is worth testing.

package.json
```diff
@@ -4,71 +4,33 @@
   "license": "GPL-2.0-or-later",
   "description": "A fast and flexible search and query engine for WordPress.",
   "devDependencies": {
-    "@10up/eslint-config": "^2.3.2",
-    "@babel/core": "^7.13.8",
-    "@babel/eslint-parser": "^7.13.8",
+    "10up-toolkit": "^1.0.10",
+    "@10up/babel-preset-default": "^1.2.0",
+    "@10up/eslint-config": "^2.3.6",
```
nicholasio, line 9: with 10up-toolkit you can remove all of the eslint deps and babel preset as it is included with 10up-toolkit.

.eslintrc.js
Rahmon: Yes. The idea is to use `@10up/eslint-config/react`. However, I noticed that we need to fix a lot of issues in the code. So, to not make this PR giant, I'll do this change in another PR.

.babelrc
Rahmon: @nicholasio Deleting this file I get this error `Can't resolve '@wordpress/element'`. Any idea what could be?

package.json
Rahmon: It will be removed in another PR: https://github.com/10up/ElasticPress/pull/2279#discussion_r680236178

.eslintrc.js
nicholasio: Make sense.

PR #2284, opened 2021-08-01
fix #2275

### LINKED ISSUE #2275: BUG: only the "post" indexable is acted upon when a site's status changes
**Describe the bug**
When a site in a multisite setup is updated so that it is archived, deleted or marked as spam, only the "post" indexable will be acted upon because it is the only indexable that adds actions for those items (https://github.com/10up/ElasticPress/blob/3.6.1/includes/classes/Indexable/Post/SyncManager.php#L53 vs https://github.com/10up/ElasticPress/blob/3.6.1/includes/classes/Indexable/Term/SyncManager.php#L37). This leaves data on the backend server.

**Steps to Reproduce**

1. Have a multisite
2. Create an additional site
3. Enable some other indexable in the dashboard, such as terms
4. Index all sites
5. Note the indexes created in the backend
6. Mark a site as deleted
7. Note the index that still exist in the backend will include indexable items other than post

**Expected behavior**

All indexes should be removed from the backend server/service

**Screenshots**

**Environment information**
 - Device: 
 - OS: 
 - Browser and version: 
 - WordPress version: 
 
 - ElasticPress version: 
 - Elasticsearch version: 
 - Where do you host your Elasticsearch server? 
 - Plugins and version: 
 - Theme and version: 
 
   </details>

**Additional context**

dustinrue:
### Description of the Change

Refactors the multi site handling so that when blogs are updated (marked as spam, archived or deleted), their indexes are also removed from the ES server. Also protects global indexes, like User, from being deleted.

### Benefits

Keeps backend servers clean of unwanted data

### Possible Drawbacks

It's possible for data meant to be kept to remove. Users should use `ep_keep_index` to keep the data.

### Verification Process

Refactored the code then tested to ensure updated sites had their data removed. One think I'm seeing is that activating a site again does _not_ cause it to be reindexed again. 

### Checklist:

### Applicable Issues

Resolved #2275 

### Changelog Entry

Fixed #2275 - ElasticPress will now remove data from Elasticsearch on a multisite network when a site is archived, deleted or marked as spam


oscarssanchez (commented):
The code looks good to me, the only thing I noticed when testing is if the site status changes to publish it won't attempt a reindex. Should we trigger this or add a notice for the specific site?

oscarssanchez (approved):

PR #2287, opened 2021-08-03
Synonyms Type

### LINKED ISSUE #2187: BUG: Synonyms + additional terms return resuls like OR instead of AND
Hello everyone!

So, I've been struggling the whole day with this and the only conclusion I reached is that this must be a bug.

I have created a list of synonyms like the following - I will use actual keywords for convenience:
synonyms: `gearbox, shifter`

When I search for the following:
'Toyota engine'

I get expected results like:
Toyota AND engine

However, when I search for something inside the synonyms list, any term NOT in the synonym list is ignored. So if I search for:
`Toyota gearbox`

I get EVERYTHING that includes 'gearbox' OR 'shifter' OR 'Toyota' while the expected result would be:
('gearbox' OR 'shifter') AND 'Toyota'

I would assume that the synonym filter takes any group of keywords in the provided array and combines it with any additional keywords using the same logic as the initial query.

Is there a way to retrieve the results I am expecting? I tried adding a custom filter and changing the order, but still had no luck.

felipeelia:
### Description of the Change

Change synonyms filter type from `synonym` to `synonym_graph`, so multiple words synonyms start working.

### Alternate Designs

n/a

### Benefits

Multiple words synonyms starting to work, as described in #2187

### Possible Drawbacks

n/a

### Applicable Issues

Closes #2187

### Changelog Entry

Fixed: Multiple words synonyms. Props @scooterlord, @jonasstrandqvist, and @felipeelia via #2287


mckdemps:
We won't be creating the default search analyzer.

oscarssanchez (approved):

Issue #2276, opened 2021-07-20
BUG: status and stats commands will fail in some multisite scenarios


dustinrue:
**Describe the bug**
There are multiple ways in which the "status" and "stats" commands will fail in a multisite setup. 

1. Neither the wp-cli command "status" or "stats" checks to see if a site is currently active prior to sending the command. This leads to an improper list of index names sent to the backend (see bullet 2)
2. When performing the _stats call to the backend, the request includes all possible index names as a comma separated set. If any one of the indexes in the set doesn't exist then all sites will show as "not indexed" even if some are. Bullet 1 leads to this situation because a site that was previously deleted in the admin will be included in the comma separated set. This is also reproduced when a site is added to a multisite but not indexed. Even if bullet 1 is resolved the command will fail because the index is include in the comma separated set but doesn't exist in the backend yet.

Suggested fix is to use a filtered _stats call to the backend (/_cat/stats/<prefix/indexname>*) and then parsing out the difference between what is indexed and what isn't. Another option is to pay attention to if the plugin is network activated and if not then only ask about indexes for the current site.

**Steps to Reproduce**

1. Have a multisite
2. Have multiple sites
3. Index only one of the sites
4. Issue wp elasticpress stats
5. See that all sites report as not indexed

**Expected behavior**

A proper listing of indexed and not indexed sites should be presented

**Screenshots**

**Environment information**
 - Device: 
 - OS: 
 - Browser and version: 
 - WordPress version: 
 
 - ElasticPress version: 
 - Elasticsearch version: 
 - Where do you host your Elasticsearch server? 
 - What ElasticPress features do you have currently active? 
 - Plugins and version: 
 - Theme and version: 
 
 - <details><summary>Site Health Info:</summary>
   
   </details>

**Additional context**

PR #2290, opened 2021-08-03
Fix command stats and status

Rahmon:
### Description of the Change

This change fixes the stats and status command in a multisite scenario when there are sites not indexed.

### Benefits

Correct status about indexes

### Verification Process

1. Have a multisite
2. Have multiple sites
3. Index only one of the sites
4. wp elasticpress stats / wp elasticpress status
5. See the correct information for sites indexed and "is not currently indexed" otherwise.

### Checklist:

### Applicable Issues

Closes: #2276 

### Changelog Entry

Fixed command stats and status in a multisite scenario.


includes/classes/Command.php
```diff
@@ -1110,6 +1110,32 @@ public function status() {
...
+				$indexes_from_cat_indices_api,
+				function( $carry, $item ) {
+					if ( ! empty( $item['index'] ) ) {
+						$carry[] = $item['index'];
+					}
+
+					return $carry;
+				},
+				[]
+			);
```
felipeelia, line 1132: @Rahmon testing it locally, it seems this entire array_reduce call could be replaced with `$indexes_from_cat_indices_api = wp_list_pluck( $indexes_from_cat_indices_api, 'index' );`. Can we change it to this simpler code?

Rahmon: Sweet. Thanks @felipeelia

felipeelia (approved):

Issue #2242, opened 2021-06-28
BUG: Wrong results while filtering by category using 'tax_query'


karols0:
**Describe the bug**
I have a pretty simple query - find all published posts that have any of given categories. Works fine without Elastipress(without `ep_integrate` flag).
```
  $args = [
    'post_type' => 'post',
    'post_status' => 'publish',
    'ep_integrate' => true,
    'tax_query' => array(
        array(
            'taxonomy' => 'category',
            'terms'    => array('2','6','1'),
            'field' => 'term_id',
            'operator'    => 'in',
        )
    )
  ];
 $ajaxposts = new WP_Query($args);
```

However when ElasticPress is on, the query generated for Elasticsearch seems to have some redundant conditions. It does search by `terms.category.term_id: ["2","6","1"]`, but it also adds `"terms.category.term_id": [2]` and `"terms.category.slug": [ "custom-category"]` for the reason uknown to me. ('custom-category' is the slug of the category with ID=2)
As a result not all of the desired posts are returned.

I'm attaching data from Debug Bar below:

Query Args:
```
array (size=66)
  'post_type' => string 'post' (length=4)
  'post_status' => string 'publish' (length=7)
  'ep_integrate' => boolean true
  'tax_query' => 
    array (size=1)
      0 => 
        array (size=4)
          'taxonomy' => string 'category' (length=8)
          'terms' => 
            array (size=3)
              ...
          'field' => string 'term_id' (length=7)
          'operator' => string 'in' (length=2)
  'error' => string '' (length=0)
  'm' => string '' (length=0)
  'p' => int 0
  'post_parent' => string '' (length=0)
  'subpost' => string '' (length=0)
  'subpost_id' => string '' (length=0)
  'attachment' => string '' (length=0)
  'attachment_id' => int 0
  'name' => string '' (length=0)
  'pagename' => string '' (length=0)
  'page_id' => int 0
  'second' => string '' (length=0)
  'minute' => string '' (length=0)
  'hour' => string '' (length=0)
  'day' => int 0
  'monthnum' => int 0
  'year' => int 0
  'w' => int 0
  'category_name' => string 'custom-category' (length=15)
  'tag' => string '' (length=0)
  'cat' => int 2
  'tag_id' => string '' (length=0)
  'author' => string '' (length=0)
  'author_name' => string '' (length=0)
  'feed' => string '' (length=0)
  'tb' => string '' (length=0)
  'paged' => int 0
[...]

PR #2299, opened 2021-08-10
Exclude category taxonomy from root tax check

oscarssanchez:
### Description of the Change

Fixes https://github.com/10up/ElasticPress/issues/2242

As WordPress core sets the `category_name` parameter when performing a Tax Query if you are doing an id's search, this conflicts with this check EP does here: https://github.com/10up/ElasticPress/blob/develop/includes/classes/Indexable/Post/Post.php#L1003

This problem only applies to the category taxonomy.

### Checklist:


Rahmon:
@oscarssanchez looks good.

Do you mind in add a test if the key `terms.category.slug` and `terms.category.term_id` are not present when we use the same args used in  #2242 ?

Something similar to https://github.com/10up/ElasticPress/blob/develop/tests/php/indexables/TestUser.php#L712

tests/php/indexables/TestPost.php
```diff
@@ -6196,4 +6197,42 @@ public function testPostPrepareDateTerms() {
...
 	}
+
+	/**
+	 * Test when we perform a Tax Query with Id's for category taxonomy they are used instead of a cat slug.
+	 *
+	 * @return void
+	 * @group  post
+	 */
+	public function testCatIdTaxQuery() {
+		$term = Functions\create_and_sync_term( 'test', 'Test category', 'The testing category', 'category' );
```
Rahmon, line 6208: the PHPUnit is getting an error in this line `Error: Call to a member function index() on bool`

tests/php/indexables/TestPost.php
```diff
@@ -6196,4 +6197,42 @@ public function testPostPrepareDateTerms() {
...
+
+		$args  = $post->format_args(
+			[
+				'cat_name'       => 'test',
+			],
+			$query
+		);
+
+		$this->assertContains( $term, $args['post_filter']['bool']['must'][0]['bool']['must'][0]['terms']['terms.category.term_id'] );
+		$this->assertNotContains( 'test', $args['post_filter']['bool']['must'][0]['bool']['must'][1]['terms']['terms.category.slug'] );
```
Rahmon, line 6236: We need to update the assertion because `['must'][1]['terms']['terms.category.slug']` doesn't exist and I think it'll always return true.

Maybe we need to check if the array `$args['post_filter']['bool']['must'][0]['bool']['must']` has only one item (assertCount) and if this only item `$args['post_filter']['bool']['must'][0]['bool']['must'][0]['terms']` has the key `terms.category.term_id` (assertArrayHasKey). 

Makes sense?

PR #2340, opened 2021-09-06
Use date_Create that does not throw exception or warning

pschoffer:
### Description of the Change
This is inspired by https://github.com/Automattic/ElasticPress/pull/113

It seems that `new \DateTime( $meta_value, \wp_timezone() );` throws not only exception but also a `E_WARNING` in certain cases.

There is the `date_create` alternative, that according to [manual](https://www.php.net/manual/en/function.date-create.php) is an alias. While testing it though, I found that, it returns false on failure and does **NOT** throw an exception or warning (as far as I can see).

There are already unittests covering this function that are passing and also testing on the local dev-environment results in good results.
### Checklist:

### Changelog Entry

Changed - Meta parsing tweak to avoid `E_WARNING`s

### DIFF
--- a/includes/classes/Indexable.php
+++ b/includes/classes/Indexable.php
@@ -477,9 +477,9 @@ abstract class Indexable {
 		$meta_types['datetime'] = '1970-01-01 00:00:01';
 		$meta_types['time']     = '00:00:01';
 
-		try {
-			// is this is a recognizable date format?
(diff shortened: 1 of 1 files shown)


pschoffer:
Thanks for that @sjinks . One day I will bother you to teach me your super-power of tracking that down (I tried but failed)

felipeelia (approved):

PR #2341, opened 2021-09-07
Fix post tag duplication in ES query

oscarssanchez:
### Description of the Change

This PR aims to fix three things. First, there's a duplication of `terms.post_tag.slug` similar to the problem that was happening in https://github.com/10up/ElasticPress/issues/2242 when you are making with similar args: 

```
$args = [
	'ep_integrate' => true,
	'tag' => array( 'slug1, slug2' ),
];
```

This is fixed here: https://github.com/10up/ElasticPress/compare/fix/post-tag-duplication?expand=1#diff-7004c1b9f26ece8d68d4d3b7b822331a1f1e0cdf34a5eda1aa146d930d21ad08R1007 adding `post_tag` and `category` as these are WP Core default taxonomies, and I have added the filter 'ep_post_tax_excluded_wp_query_root_check' in case this is needed to be extended by a developer.

Second, doing a query for more than one term in an array form does not work, as it will format the `tag` arg like a comma separated string.

Third, searching for multiple tags with comma separated tags does not work at all. This is allowed by core: https://developer.wordpress.org/reference/classes/wp_query/#tag-parameters

It looks like core will translate both the array and comma separated strings as comma separated, which is different than how it formats a tax query, therefore that part remains unaffected  

### Benefits

Expand `WP_Query`

### Checklist:


includes/classes/Indexable/Post/Post.php
```diff
@@ -938,9 +938,12 @@ public function format_args( $args, $wp_query ) {
 		}
 
 		if ( isset( $args['tag'] ) && ! empty( $args['tag'] ) ) {
+			if ( ! is_array( $args['tag'] ) && strpos( $args['tag'], ',' ) ) {
```
Rahmon, line 941: I'm a little concerned about `strpos` because it could return 0 and evaluate to `false`. Other than that, wouldn't be an issue with just one tag?

tests/php/indexables/TestPost.php
```diff
@@ -4975,6 +4975,46 @@ public function testNestedTaxQuery() {
...
+			'tag' => 'slug1,slug2',
+		];
+
+		$query2_args = [
+			's'   => 'findme',
+			'tag' => ['slug1', 'slug2'],
+		];
+
+		$query1 = new \WP_Query( $query1_args );
+		$query2 = new \WP_Query( $query2_args );
```
Rahmon, line 5010: After these WP_Query, can you check if the data was retrieved using elasticsearch?

```php
foreach ( $query1->results as $item ) {
	$this->assertTrue( $item->elasticsearch );
}
```

oscarssanchez: Thanks @Rahmon I have addressed this here: https://github.com/10up/ElasticPress/pull/2341/commits/bbf5f9b114ad9fd10ea66f0df57576fc2740bd80

includes/classes/Indexable/Post/Post.php
oscarssanchez: Good call @Rahmon , I have addressed it here https://github.com/10up/ElasticPress/pull/2341/commits/1d71d8a622aef3eec3c580ace6ae69cbe46785e0

For your other question, if it is a single tag the evaluation of this logic will be false. I have checked it works with one tag as well.

Rahmon (dismissed):

includes/classes/Indexable/Post/Post.php
```diff
@@ -949,8 +952,16 @@ function( $tax_query ) use ( $args ) {
...
-			// Exclude the category taxonomy from this check if we are performing a Tax Query as category_name will be set by core
-			if ( $tax->query_var && ! empty( $args[ $tax->query_var ] ) && 'category' !== $tax->name ) {
+			// Prevents duplication of core's default taxonomies post_tag and category in ES query.
+			apply_filters(
+				'ep_post_tax_excluded_wp_query_root_check',
+				$excluded_tax_from_root_check = [
+					'category',
+					'post_tag',
+				]
+			);
```
felipeelia, line 962: @oscarssanchez @Rahmon I'm pretty sure this should be like the following, right?
```
			$excluded_tax_from_root_check = apply_filters(
				'ep_post_tax_excluded_wp_query_root_check',
				[
					'category',
					'post_tag',
				]
			);
```
Also, we need docs for this.

felipeelia (changes requested):

includes/classes/Indexable/Post/Post.php
felipeelia: Actually, looking at the code, that variable should be set outside of the foreach, there seems to be no point in setting that variable with the same value on every iteration.

includes/classes/Indexable/Post/Post.php
```diff
@@ -948,9 +951,18 @@ function( $tax_query ) use ( $args ) {
 		 */
 		$taxonomies = get_taxonomies( array(), 'objects' );
 
+		// Prevents duplication of core's default taxonomies post_tag and category in ES query.
+		$excluded_tax_from_root_check = apply_filters(
+			'ep_post_tax_excluded_wp_query_root_check',
```
felipeelia, line 956: @oscarssanchez not sure if this is under your radar already but we still need docs for this.

felipeelia (changes requested):

felipeelia (approved):

PR #2344, opened 2021-09-10
Terms: Use $query_vars['parent'] even when equals 0

### LINKED ISSUE #2232: BUG: When using get_terms it doesn't obey parent = 0
On my site I have following:

$terms = get_terms( array( 'taxonomy' => $taxonomy, 'parent' => 0, 'hide_empty' => true,  ) );

Works perfectly, it gets the terms that don't have any parents. 
Now if I add  'ep_integrate' => true, it's doesn't obey the parent = 0 anymore. It gets all the possible terms.

$terms = get_terms( array( 'taxonomy' => $taxonomy, 'parent' => 0, 'hide_empty' => true,  'ep_integrate' => true, ) );

felipeelia:
### Description of the Change

Currently, we are using `if ( ! empty( $query_vars['parent'] ) ) {` which isn't true when `parent = 0`. This PR changes the test to use `isset()` instead.

### Applicable Issues

Closes #2232

### DIFF
--- a/includes/classes/Indexable/Term/Term.php
+++ b/includes/classes/Indexable/Term/Term.php
@@ -424,7 +424,7 @@ class Term extends Indexable {
 		/**
 		 * Support `parent` query var.
 		 */
-		if ( ! empty( $query_vars['parent'] ) ) {
+		if ( isset( $query_vars['parent'] ) ) {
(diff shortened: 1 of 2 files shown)


oscarssanchez (approved):
LGTM. Some WP acceptance tests are failing but seems unrelated to the change.

PR #2345, opened 2021-09-13
Changes in mapping version determination

### LINKED ISSUE #2331: Early returns in `Post::determine_mapping_version()`

The [check inside the new `Indexable/Post/Post::determine_mapping_version()`](https://github.com/10up/ElasticPress/blob/3.6.2/includes/classes/Indexable/Post/Post.php#L300) method has some room for improvement. Instead of using the `not-` strings, we could take it to a separate method and have early returns, in that way if someone has a custom mapping and the first test fails, it would check it for the next version.

This is especially important for the first test, where we check `$mapping[ $index ]['mappings']['post']['properties']['post_name']['fields']['raw']['ignore_above']` and not handle the case where it is not set. 

Props to @pschoffer

### LINKED ISSUE #2332: BUG: `index.max_shingle_diff` check in `Post::determine_mapping_version()`
**Describe the bug**

Inside the [new `Indexable/Post/Post::determine_mapping_version()`](https://github.com/10up/ElasticPress/blob/3.6.2/includes/classes/Indexable/Post/Post.php#L300) method the plugin checks for `$mapping[ $index ]['settings']['index.max_shingle_diff']`, as set [in the mapping](https://github.com/10up/ElasticPress/blob/3.6.2/includes/mappings/post/7-0.php#L62). In fact, Elasticsearch returns the mapping as a perfect array, so we should check `$mapping[ $index ]['settings']['index']['max_shingle_diff']` actually.

**Additional context**
Props to @pschoffer

felipeelia:
### Description of the Change

This PR splits `Post::determine_mapping_version()` in two different methods, so we can:
1. Early return if a version is already determined
2. Keep the `ep_post_mapping_version_determined` filter being executed at the end of the process.

It also fixes the check for ES 7.

### Applicable Issues

Closes #2332 
Closes #2331

### Changelog Entry

- Fixed: Mapping version determination does not send false positives for ES 7 anymore.

### DIFF
--- a/includes/classes/Indexable/Post/Post.php
+++ b/includes/classes/Indexable/Post/Post.php
@@ -288,58 +288,7 @@ class Post extends Indexable {
 			return false;
 		}
 
-		$version = 'unknown';
-
-		if ( isset( $mapping[ $index ]['mappings']['post']['_meta']['mapping_version'] ) ) {
-			$version = $mapping[ $index ]['mappings']['post']['_meta']['mapping_version'];
-		} elseif ( isset( $mapping[ $index ]['mappings']['_meta']['mapping_version'] ) ) {
-			$version = $mapping[ $index ]['mappings']['_meta']['mapping_version'];
-		}
-
-		// mapping does not have meta value set - use legacy detection
-		if ( 'unknown' === $version ) {
-
-			// check for pre-5-0 mapping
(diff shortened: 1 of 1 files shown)


oscarssanchez (approved):
LGM WP acceptance seem to fail but I assume the errors are unrelated based on the logs @felipeelia

PR #2347, opened 2021-09-14
Autosuggest: Major rewrite of `adjust_fuzzy_fields`

### LINKED ISSUE #2346: BUG: Add back `term_suggest` to search queries
**Describe the bug**

Although we index a field called "term_suggest" it's not been used. We should include it back in autosuggest search queries.

**Steps to Reproduce**
1. Make an autosuggest search
2. Check that each document in results has a "term_suggest" property
3. Note that field is not used in the search query

**Expected behavior**
That field should/could be used in autosuggest searches.

**Additional context**
We already add a field called `post_title.suggest` in autosuggest [here](https://github.com/10up/ElasticPress/blob/develop/includes/classes/Feature/Autosuggest/Autosuggest.php#L248).

felipeelia:
### Description of the Change

This PR addresses the problem outlined in #2346 and improves the readability of the `Autosuggest::adjust_fuzzy_fields()` method.

### Benefits

- Some early returns and more comments will (hopefully) make the method easier to understand
- `term_suggest` is now used for autosuggest searches

### Applicable Issues

Closes #2346 

### Changelog Entry

Fixed: In addition to post titles, now autosuggest also partially matches taxonomy terms.

### DIFF
--- a/includes/classes/Feature/Autosuggest/Autosuggest.php
+++ b/includes/classes/Feature/Autosuggest/Autosuggest.php
@@ -255,8 +255,8 @@ class Autosuggest extends Feature {
 		$ngram_fields = apply_filters(
 			'ep_autosuggest_ngram_fields',
 			[
-				'post_title'       => 'post_title.suggest',
-				'terms.(.*?).name' => 'term_suggest',
+				'post_title'      => 'post_title.suggest',
+				'terms.(.+).name' => 'term_suggest',
 			]
 		);
 
--- a/includes/classes/Feature/Autosuggest/Autosuggest.php
+++ b/includes/classes/Feature/Autosuggest/Autosuggest.php
@@ -237,57 +237,117 @@ class Autosuggest extends Feature {
 	 * @return array $query adjusted ES Query arguments
 	 */
 	public function adjust_fuzzy_fields( $query, $post_type, $args ) {
-		if ( Utils\is_integrated_request( $this->slug, [ 'public' ] ) && ! empty( $args['s'] ) ) {
-			/**
-			 * Filter autosuggest ngram fields
-			 *
-			 * @hook ep_autosuggest_ngram_fields
-			 * @param  {array} $fields Fields available to ngram
-			 * @return  {array} New fields array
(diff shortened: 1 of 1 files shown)


includes/classes/Feature/Autosuggest/Autosuggest.php
```diff
@@ -237,57 +237,117 @@ public function set_fuzziness( $fuzziness, $search_fields, $args ) {
...
+		 *
+		 * @hook ep_autosuggest_ngram_fields
+		 * @param  {array} $fields Fields available to ngram
+		 * @return  {array} New fields array
+		 */
+		$ngram_fields = apply_filters(
+			'ep_autosuggest_ngram_fields',
+			[
+				'post_title'       => 'post_title.suggest',
+				'terms.(.*?).name' => 'term_suggest',
```
Rahmon, line 259: My only concern is about this regex specifically this part: (.*?). Since `.` represents any character, it could match something like `terms..name`. So, I'd only suggest using a more specific regex.

felipeelia: Good point @Rahmon. I've changed that now, can you please give it another look? Thanks!

Rahmon: I noticed another thing, we should escape the `.` otherwise it could match `terms*test&name` because `.` match any character.

Other than that, I think we should use something like `terms\.([a-z0-9]+(?:-[a-z0-9]+)*)\.name`

felipeelia: @Rahmon do you mind explaining why the `(?:-[a-z0-9]+)*` part?

Rahmon: @felipeelia It indicates that we don't want to capture the group but it doesn't make sense in this scenario. So you can remove the `?:`: `terms\.([a-z0-9]+(-[a-z0-9]+)*)\.name`.

Rahmon (approved):

PR #2369, opened 2021-09-25
Update WP-CLI documentation

felipeelia:
### Description of the Change

This PR updates our WP-CLI commands documentation. It was generated with this [WP-CLI command](https://github.com/felipeelia/cli-command-docs) using:

```
wp cli-command-docs elasticpress --custom-order=index,activate-feature,deactivate-feature,list-features,get-algorithm-version,set-algorithm-version --remove=delete_transient_on_int,custom_get_transient --custom-intro='The following WP-CLI commands are supported by ElasticPress:' > docs/wp-cli.md
```

Another PR will follow this one to make this process automatic.

### Benefits

WP-CLI already relies on the DocBlock above each subcommand. Generating the documentation from it we will have both things always sync'd.

### Possible Drawbacks

None. If someone wants to contribute with WP-CLI documentation specifically, it should be only a matter of updating the docblock in Command.php.

### Changelog Entry

Changed: WP-CLI documentation was updated

PR #2370, opened 2021-09-25
Automatically update wp-cli.md

felipeelia:
### Description of the Change

Based on the work made for #2369, this PR is an attempt to automate the process.


.github/workflows/build-docs.yml
```diff
@@ -1,28 +1,80 @@
...
 on:
- push:
-   branches:
-    - master
+  push:
+    branches:
+     - master
+  pull_request:
+    branches:
+      - develop
```
jeffpaul, line 9: We purposely do not update the docs off changes to develop as those items are not yet in a released version and thus the docs would conflict with whats actually in the plugin.  You might consider removing this unless there's a precise reason to keep it?

jeffpaul (commented):
Main question is if you really want to update the docs off develop or not.

.github/workflows/build-docs.yml
felipeelia: The only reason why I added that was to generate the artifact after creating the PR, I don't think there is any reason why we should update these docs when things are merged to develop. I'll push a commit shortly with what I think should be the "final" version of the action.

felipeelia: Done @jeffpaul. 5aa65ab contains the version that I'd consider ideal to be merged. Thanks!

felipeelia: Yup @jeffpaul.

.github/workflows/build-docs.yml
```diff
@@ -1,28 +1,71 @@
...
+      with:
+        path: 'wp-content/plugins/elasticpress'
+
+    - name: Set PHP version
+      uses: shivammathur/setup-php@v2
+      with:
+        php-version: '7.4'
+        coverage: none
+        tools: prestissimo, composer:v1
+        ini-values: memory_limit=3G
```
dinhtungdu, line 23: I don't see the usage of `composer` in this workflow, and PHP 7.4 is shipped with `ubuntu-latest`. I'm wondering if we can remove this section?

felipeelia: Good points @dinhtungdu! 
1. Composer is used by the wp package install command: https://github.com/wp-cli/package-command
2. GH Actions have both PHP 7.4 and 8.0 available at [this point](https://github.com/actions/virtual-environments/blob/main/images/linux/Ubuntu2004-README.md#php). I _think_ it will already use 8.0 by default but even if it is still using 7.4 it is just a matter of time to have 8.0 as the default. I'd rather force 7.4 until we are all sure EP is fully compatible with that version than risking it breaking. Also, we need that `memory_limit=3G` anyway because the package install command is a [huge memory consumer](https://make.wordpress.org/cli/handbook/guides/common-issues/#php-fatal-error-allowed-memory-size-of-999999-bytes-exhausted-tried-to-allocate-99-bytes).

dinhtungdu: Thanks, then this is looking good to me! Note that I assume the `cli-command-docs` works as expected.

jeffpaul:
:shipit:

PR #2378, opened 2021-09-29
Add get-mapping WP-CLI Subcommand

tfrommen:
### Description of the Change

Using the `wp elasticpress` WP-CLI command, I am able to create and retrieve any or all indexes. However, I can **put** mappings only, but not retrieve.

This PR adds a new subcommand, `wp elasticpress get-mapping`, that allows for retrieving the current mappings:

```shell
wp elasticpress get-mapping
```

By passing an optional index name, the command will return the mapping for that index only:

```shell
wp elasticpress get-mapping --index-name=ep-examplecom-post-1
```

### Alternate Designs

If I have access to WP-CLI, but don't have this subcommand, I could do the following:

```shell
wp shell

wp> wp_remote_retrieve_body( ElasticPress\Elasticsearch::factory()->remote_request( '[<index>/]_mapping' ) );
```

That is unneccesarily complicated.

### Benefits

More complete API/feature set. Easier to use than `wp shell` command.

### Possible Drawbacks

I don't see any.

### Verification Process

I tested this locally, both with and without passing an index name.

### Checklist:

### Changelog Entry

```md
### Added
- New WP-CLI subcommand for retrieving mappings: `wp elasticpress get-mapping [--index-name]`. Props [@tfrommen](https://github.com/tfrommen) via [#2378](https://github.com/10up/ElasticPress/pull/2378).
```

### DIFF
--- a/includes/classes/Command.php
+++ b/includes/classes/Command.php
@@ -348,6 +348,32 @@ class Command extends WP_CLI_Command {
 	}
 
 	/**
+	 * Return the mapping as a JSON object. If an index is specified, return its mapping only.
+	 *
+	 * ## OPTIONS
+	 *
+	 * [--index-name]
+	 * : The name of the index for which to return the mapping. If not passed, all mappings will be returned
+	 *
+	 * @subcommand get-mapping
+	 * @since      3.7
+	 * @param array $args Positional CLI args.
+	 * @param array $assoc_args Associative CLI args.
+	 */
+	public function get_mapping( $args, $assoc_args ) {
+		$path = '_mapping';
+		if ( ! empty( $assoc_args['index-name'] ) ) {
+			$path = $assoc_args['index-name'] . '/' . $path;
+		}
+
+		$response = Elasticsearch::factory()->remote_request( $path );
+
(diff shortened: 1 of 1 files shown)


felipeelia:
Hey @tfrommen, first of all, thanks for the PR! I think this will be a nice addition to the plugin.

In cases where different clients are using a shared ES instance, we don't want to try to return all indexes in the server, though. 

Perhaps we could:
1. Move the logic we have to get all indexes in `get_indexes` to a new protected function
2. Call that in `get_indexes` and in this new `get_mapping` method

How does that sound?

ps.: Looking at that logic, it seems we should iterate over all registered indexables for each site, rather than just post.

tfrommen:
How does that look to you now, @felipeelia?

tfrommen:
Any updates, anyone? 🙂

felipeelia:
Hi @tfrommen! Overall it is looking pretty good, thanks for that! I've left some comments there, let me know if that is something you can help us with, okay? Thanks

felipeelia:
Closing this in favor of #2414.

Issue #2335, opened 2021-08-31
Check if all tests are using Elasticsearch


Rahmon:
**Is your enhancement related to a problem? Please describe.**

**Describe the solution you'd like**
We need to make sure that our tests that fetch data are using Elasticsearch. Example how we're checking it ([TestUser.php](https://github.com/10up/ElasticPress/blob/develop/tests/php/indexables/TestUser.php#L315)):

```php
foreach ( $user_query->results as $user ) {
	$this->assertTrue( $user->elasticsearch );
}
```

If we don't check it, ElasticPress could use the database and the tests will probably always pass.

Tests that need to be checked:

- [x] [testSearchAllowedMimeType](https://github.com/10up/ElasticPress/blob/develop/tests/php/features/TestDocuments.php#L94) (#2388)
- [x] [testSearchDisallowedMimeType](https://github.com/10up/ElasticPress/blob/develop/tests/php/features/TestDocuments.php#L138) (#2388)
- [x] [testSearchNormalPost](https://github.com/10up/ElasticPress/blob/develop/tests/php/features/TestDocuments.php#L187) (#2388)
- [x] [testProductsPostTypeQueryOn](https://github.com/10up/ElasticPress/blob/develop/tests/php/features/TestWooCommerce.php#L62) https://github.com/10up/ElasticPress/pull/2389
- [x] [testProductsPostTypeQueryProductCatTax](https://github.com/10up/ElasticPress/blob/develop/tests/php/features/TestWooCommerce.php#L116) https://github.com/10up/ElasticPress/pull/2389
- [x] [testSearchOnShopOrderAdmin](https://github.com/10up/ElasticPress/blob/develop/tests/php/features/TestWooCommerce.php#L148) https://github.com/10up/ElasticPress/pull/2389
- [x] [testSearchOnAllFrontEnd](https://github.com/10up/ElasticPress/blob/develop/tests/php/features/TestWooCommerce.php#L174) https://github.com/10up/ElasticPress/pull/2389
- [x] [testCommentQueryOrderCommentContent](https://github.com/10up/ElasticPress/blob/develop/tests/php/indexables/TestComment.php#L346) https://github.com/10up/ElasticPress/pull/2390
- [x] [testCommentQueryOrderCommentID](https://github.com/10up/ElasticPress/blob/develop/tests/php/indexables/TestComment.php#L403) https://github.com/10up/ElasticPress/pull/2390
[...]

PR #2400, opened 2021-10-12
Check if query is using Elasticsearch on testMultipleTests group

Rahmon:
### Checklist:

### Applicable Issues

https://github.com/10up/ElasticPress/issues/2335


Rahmon:
> @Rahmon overall it looks good but it seems for some tests we are checking if the `ep_wp_query_search` action was fired (`testWPQuerySearchContent()` and `testWPQuerySearchTitle()`) and for the others, we are relying on `$query->elasticsearch_success`. Is there any difference?

You should rely only on `$query->elasticsearch_success`. I've updated the PR.

PR #2417, opened 2021-10-18
Terms: Adjust fuzziness for admin term search queries

rebeccahum:
### Description of the Change

This fixes a bug when searching in the admin for post tags and the results are unexpected. To reproduce:

1) Enable feature terms `wp elasticpress activate-feature terms`
2) Run `wp elasticpress index --setup` 
3) Create three post tags:
- Test Post
- Test Ruth
- The Most Fun Thing
4) Search for "the most fun thing" and you will see that all terms are being returned

It doesn't make sense to have fuzzy search enabled for terms in the back-end, especially since users are most likely to be searching for a specific word in the tags..

### Alternate Designs

We could change the default to be 0 instead of 1 but I'm not sure how far reaching that would be for non-admin term queries: https://github.com/10up/ElasticPress/blob/a1c02b2c5b9706da824bf7a77270dad10fcbea3f/includes/classes/Indexable/Term/Term.php#L387

### Benefits

Results will be more accurate when searching for a tag with a term in it.

### Possible Drawbacks

For users who appreciate the fuzziness, they may be wondering why it is no longer there.

### Verification Process

1) Follow reproduction steps in "Description of the Change"
2) Check out PR
3) Do search again from step 4 and see just the one result: "The most fun thing"

### Checklist:

### Changelog Entry

Changed: Terms search query in admin will not be fuzzy


rebeccahum:
Thanks for looking @felipeelia. Version tag has been changed.

PR #2438, opened 2021-11-04
Add a "request" subcommand

joehoyle:
To help when debugging, or otherwise wanting to send arbitrary requests to ElasticSearch, this adds a new `request` sub-command, that is some what similar to the idea of `wp db query` or the like.

For example `wp elasticpress request /_cat/indices --method=GET`

### DIFF
--- a/includes/classes/Command.php
+++ b/includes/classes/Command.php
@@ -1852,4 +1852,45 @@ class Command extends WP_CLI_Command {
 	public static function skip_number_format_i18n( $formatted, $number, $decimals ) {
 		return number_format( $number, absint( $decimals ), '.', '' );
 	}
+
+	/**
+	 * Send a HTTP request to Elasticsearch
+	 *
+	 * @subcommand request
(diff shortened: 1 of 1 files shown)


felipeelia:
Hey @joehoyle,

Sorry for taking so long to get back to you (I was sure I've replied to you already!) The idea is great but we need some adjustments in the PR in order to merge it. Can you please:
1. Fix the lint problems found by the GitHub Action?
2. Merge `develop` into your branch again so PHPUnit tests run
3. Add a `@since` tag to the new method doc block. Let's add it to `4.0.0`
4. We are now following a new pattern for command arguments, which can be seen [here](https://github.com/10up/ElasticPress/blob/5399716601d07fb9cce65849fc54f6c7be4b1d76/includes/classes/Command.php#L177). Having them in that pattern helps to build the [docs](http://10up.github.io/ElasticPress/tutorial-wp-cli.html).

Thanks!

felipeelia:
Closing in favor of #2484

PR #2443, opened 2021-11-05
Add $type parameter to ep_do_intercept_request filter

rebeccahum:
### Description of the Change

It would be nice if we could pass in the `$type` parameter into the `ep_do_intercept_request` filter:

https://github.com/10up/ElasticPress/blob/5bb9d76ff4bfd85761e4416a892d731e62cdf33d/includes/classes/Elasticsearch.php#L1067-L1071

This way, we have the flexibility of intercepting the request based on the type of request.

### Alternate Designs

N/A.

### Benefits

We would be able to intercept the request based on the type of request.

### Possible Drawbacks

Those hooking onto the filter already may need to update their filter arguments

### Checklist:

### Changelog Entry

Changed: Add $type parameter to ep_do_intercept_request filter

### DIFF
--- a/includes/classes/Elasticsearch.php
+++ b/includes/classes/Elasticsearch.php
@@ -1151,6 +1151,8 @@ class Elasticsearch {
 				 * Filter intercepted request
 				 *
 				 * @hook ep_do_intercept_request
+				 * @since 3.2.2
+				 * @since 4.0.0 added $type
 				 * @param {array} $request New remote request response
 				 * @param  {array} $query Remote request arguments
 				 * @param  {args} $args Request arguments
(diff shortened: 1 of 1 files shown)


felipeelia:
hey @rebeccahum, thanks for the PR! Do you mind updating the doc block to the following? (just adding the `@since` info there)

```
/**
 * Filter intercepted request
 *
 * @hook ep_do_intercept_request
 * @since 3.2.2
 * @since 4.0.0 added $type
 * @param {array} $request New remote request response
 * @param  {array} $query Remote request arguments
 * @param  {args} $args Request arguments
 * @param  {int} $failures Number of failures
 * @param  {string} $type Type of request
 * @return {array} New request
 */
```

rebeccahum:
@felipeelia done

PR #2445, opened 2021-11-05
Cache the post mapping version

### LINKED ISSUE #2432: BUG: as an admin user, every page load in the dashboard results in a "get mapping" call
**Describe the bug**

If you are a full admin user, then every page load within the admin dashboard results in a remote call to Elasticsearch to fetch the mapping. This adds a significant amount of time to each page load.

**Steps to Reproduce**

1. Be a user with the Administrator role
2. Enable the plugin
3. Browse the admin
4. Suffer with longer page loads

**Expected behavior**

The mapping currently installed vs what is expected probably won't change often or ever until a plugin update is performed. 

My suggestion is to rewrite notices routines so they happen on wp-cron rather than inline with page requests. Use transients to store the results of the checks and display noticed based on what the transient says. Cache this information and don't refresh it unless necessary.

felipeelia:
### Description of the Change

This PR caches the detected Posts mapping version for a day. This way, we don't send get_mapping calls on every request but we are still able to detect if the ES server was upgraded in a timely manner. People wanting to change that period can do that via the new `ep_post_mapping_version_cache_expiration` filter.

### Applicable Issues

Closes #2432

### Changelog Entry

Changed: Cache the detect posts mapping version, avoiding get_mapping calls in all admin requests.

### DIFF
--- a/includes/classes/Indexable/Post/Post.php
+++ b/includes/classes/Indexable/Post/Post.php
@@ -277,18 +277,36 @@ class Post extends Indexable {
 	 * @return string|WP_Error|false $version
 	 */
 	public function determine_mapping_version() {
-		$index   = $this->get_index_name();
-		$mapping = Elasticsearch::factory()->get_mapping( $index );
+		$version = get_transient( 'ep_post_mapping_version' );
 
-		if ( empty( $mapping ) ) {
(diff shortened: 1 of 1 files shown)


oscarssanchez (approved):
LGTM!

PR #2448, opened 2021-11-09
Adjust visuals sync progress

Rahmon:
### Checklist:


assets/js/sync.js
```diff
@@ -1,23 +1,178 @@
...
+
+function deleteAndSync() {
+	activeBox = deleteAndSyncBox;
+
+	disableButtonsInSyncBox();
+	updateDisabledAttribute(deleteAndSyncButton, true);
+	showPauseStopButtons();
+	showProgress();
+
+	addLineToOutput(__('Deleting all data…', 'elasticpress'));
```
felipeelia, line 78: This isn't accurate (or shouldn't be), as we are not deleting everything at once.

includes/classes/IndexHelper.php
```diff
@@ -93,14 +93,17 @@ protected function build_index_meta() {
...
-			'network_alias' => [],
-			'start_time'    => microtime( true ),
-			'totals'        => [
+			'method'          => ! empty( $this->args['method'] ) ? $this->args['method'] : 'web',
+			'offset'          => ! empty( $this->args['offset'] ) ? absint( $this->args['offset'] ) : 0,
+			'start'           => true,
+			'sync_stack'      => [],
+			'network_alias'   => [],
+			'start_time'      => microtime( true ),
+			'start_date_time' => $start_date_time ? $start_date_time->format( 'D, F d, Y H:i' ) : false,
```
felipeelia, line 105: Not a blocker for Beta 1 but we should have this stored in another format, so people can translate that format later.

includes/partials/sync-page.php
```diff
@@ -2,35 +2,209 @@
...
+						</button>
+					</div>
+				</div>
+			</div>
+			<div class="ep-sync-box__progress-wrapper">
+				<div class="ep-sync-box__progress">
+					<div class="ep-sync-box__sync-in-progress">
+						<img width="36" height="36" src="<?php echo esc_url( plugins_url( '/images/sync-in-progress.png', dirname( __DIR__ ) ) ); ?>" />
+						<div class="ep-sync-box__sync-in-progress-info">
+							<div class="ep-sync-box__progress-info">Sync in progress</div>
```
felipeelia, line 85: can we wrap this text with a esc_html__() call?

includes/partials/sync-page.php
```diff
@@ -2,35 +2,209 @@
...
+						</span>
+					</button>
+				</div>
+			</div>
+			<div class="ep-sync-box__progress-wrapper">
+				<div class="ep-sync-box__progress">
+					<div class="ep-sync-box__sync-in-progress">
+						<img width="36" height="36" src="<?php echo esc_url( plugins_url( '/images/sync-in-progress.png', dirname( __DIR__ ) ) ); ?>" />
+						<div class="ep-sync-box__sync-in-progress-info">
+							<div class="ep-sync-box__progress-info">Sync in progress</div>
```
felipeelia, line 166: same as above

includes/partials/sync-page.php
```diff
@@ -2,35 +2,209 @@
...
+							<div class="ep-sync-box__start-time">Start time: <span class="ep-sync-box__start-time-date">Fri, September 30, 2021 16:20</span></div>
```
felipeelia, line 167: this one too :)

felipeelia (changes requested):
Some issues I found during my tests:

### General
- [x] Progress bar doesn't seem to be working. As soon as I click on Sync Now, it goes directly to 100% but if I click on Show log I can see things happening
- [x] If the user pauses the sync and reload the page, the pause/stop should be visible. Clicking on Sync Now will resume it but it is not clear it works like that.

### Errors count
In a site with content imported from WooCommerce sample data and tests/wpa/test-docs/content-example.xml, and also wit a script to cause the `Limit of total fields [5000] in index [...] has been exceeded` error
- [ ] 350 posts per page causes a "Request Entity Too Large" error. That is outputted in the Errors tab but the counter isn't updated
- [x] Full log isn't "full". Errors should be outputted in both places, or we need to change the first tab name
- [ ] Reducing the 350 to 100 I can see 24 lines of errors in that tab but the count says 123

includes/partials/sync-page.php
Rahmon: Done

includes/partials/sync-page.php
Rahmon: done

includes/partials/header.php
```diff
@@ -23,18 +23,10 @@
...
 			<a class="dashicons cancel-sync dashicons-no" title="<?php esc_attr_e( 'Cancel Sync', 'elasticpress' ); ?>" aria-label="<?php esc_attr_e( 'Cancel Sync', 'elasticpress' ); ?>"></a>
 		<?php endif; ?>
-		<?php if ( Elasticsearch::factory()->get_elasticsearch_version() && defined( 'EP_DASHBOARD_SYNC' ) && EP_DASHBOARD_SYNC ) : ?>
-			<a
-				class="dashicons start-sync dashicons-update"
-				title="<?php esc_attr_e( 'Start Sync', 'elasticpress' ); ?>"
-				aria-label="<?php esc_attr_e( 'Start Sync', 'elasticpress' ); ?>"
-				<?php echo ( $is_sync_page ) ? '' : 'href="' . esc_url( $base_url . 'elasticpress-sync' ) . '"'; ?>
-			></a>
-		<?php endif; ?>
```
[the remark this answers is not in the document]
Rahmon: I brought this button back. It makes sense to me to keep it there.

PR #2462, opened 2021-11-16
Ensure array query parameters do not contain empty items

roborourke:
### Description of the Change

In some cases query args like `post__in` that are expected to be array might contain invalid data such as an empty string. An array like `[ '' ]` will not appear as empty when checked with `empty( [ '' ] )` so this update ensures the arrays only contain real values. Database IDs all start from 1 so removing `0` is ok here.

This issue was observed with Yoast SEO adding subqueries for indexables to populate `post__in` parameters, however the index was incomplete. Can also happens in non-production environments where indexables aren't saved to the database.

### Benefits

Prevents ES query errors such as this one:

```
Warning[16-Nov-2021 09:00:26 UTC] PHP Warning:  Error in ElasticSearch request: {"error":{"root_cause":[{"type":"query_shard_exception","reason":"failed to create query: empty String","index_uuid":"nRExUs6-TSCUfwX3cBDodg","index":"examplecom-post-2"}],"type":"search_phase_execution_exception","reason":"all shards failed","phase":"query","grouped":true,"failed_shards":[{"shard":4,"index":"examplecom-post-2","node":"-ftiX3JgRtC_yi8bXpnR9A","reason":{"type":"query_shard_exception","reason":"failed to create query: empty String","index_uuid":"nRExUs6-TSCUfwX3cBDodg","index":"examplecom-post-2","caused_by":{"type":"number_format_exception","reason":"empty String"}}}]},"status":400} (400) in /usr/src/app/vendor/altis/cloud/inc/namespace.php on line 365
```

Note this part:

```
"caused_by":{"type":"number_format_exception","reason":"empty String"}
```

### Verification Process

I added this code to as a `mu-plugin` reproduce the WP Query args as a minimal test case:

```
add_action( 'pre_get_posts', function( \WP_Query $query ) : void {
	$query->set( 'post__not_in', [ '' ] );
}  );
```

Observe running a search query yields no results (or falls back to a db query).

The change is intentionally light in terms of sanitisation and I haven observed no regressions with valid values for the listed query arguments, even if provided as strings e.g. `(array) "some string" -> [ "some string" ]`.

### Checklist:

### DIFF
--- a/includes/classes/Indexable/Post/Post.php
+++ b/includes/classes/Indexable/Post/Post.php
@@ -887,6 +887,32 @@ class Post extends Indexable {
 			),
 		);
 		$use_filters = false;
+		
+		// Sanitize array query args. Elasticsearch will error if a terms query contains empty items like an
+		// empty string.
+		$keys_to_sanitize = [
+			'author__in',
+			'author__not_in',
+			'category__and',
+			'category__in',
+			'category__not_in',
+			'tag__and',
+			'tag__in',
+			'tag__not_in',
+			'tag_slug__and',
+			'tag_slug__in',
+			'post_parent__in',
+			'post_parent__not_in',
+			'post__in',
+			'post__not_in',
+			'post_name__in',
+		];
+		foreach ( $keys_to_sanitize as $key ) {
+			if ( ! isset( $args[ $key ] ) ) {
+				continue;
+			}
+			$args[ $key ] = array_filter( (array) $args[ $key ] );
+		}
 
 		/**
 		 * Tax Query support
@@ -1072,7 +1098,7 @@ class Post extends Indexable {
 		 *
 		 * @since 3.6.0
 		 */
-		if ( ! empty( $args['post_name__in'] ) ) {
+		if ( ! empty( array_filter( (array) $args['post_name__in'] ) ) ) {
 			$filter['bool']['must'][]['bool']['must'] = array(
 				'terms' => array(
 					'post_name.raw' => array_values( (array) $args['post_name__in'] ),
--- a/includes/classes/Indexable/Post/Post.php
+++ b/includes/classes/Indexable/Post/Post.php
@@ -887,7 +887,7 @@ class Post extends Indexable {
 			),
 		);
 		$use_filters = false;
-		
+
 		// Sanitize array query args. Elasticsearch will error if a terms query contains empty items like an
 		// empty string.
 		$keys_to_sanitize = [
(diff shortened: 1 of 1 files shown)


brandwaffle:
@roborourke thanks for submitting the PR! We'd like to include this in 3.6.5 which will be shipping next week, but that would require fixing the failing tests. Is this something you would have time for in the short term? If not, we can always slot this for a later release. Thanks in advance!

roborourke:
@brandwaffle I’ll have a crack at the tests tomorrow, some of the failures looked unrelated so I might need a bit of help

includes/classes/Indexable/Post/Post.php
```diff
@@ -887,6 +887,32 @@ public function format_args( $args, $wp_query ) {
 			),
 		);
 		$use_filters = false;
+
```
felipeelia, line 890: Hey @roborourke, it seems you have a white space or a tab in this line, do you mind removing it, so the lint check passes? Thanks!

includes/classes/Indexable/Post/Post.php
```diff
@@ -1072,7 +1098,7 @@ function( $tax_query ) use ( $args ) {
 		 *
 		 * @since 3.6.0
 		 */
-		if ( ! empty( $args['post_name__in'] ) ) {
+		if ( ! empty( array_filter( (array) $args['post_name__in'] ) ) ) {
```
felipeelia, line 1101: As `$args['post_name__in']` isn't always set, `array_filter( (array) $args['post_name__in'] )` will trigger that error we are seeing in the PHPUnit tests. Perhaps we could rewrite the entire section and make use of a new variable. Something like:
```
/**
 * 'post_name__in' arg support.
 *
 * @since 3.6.0
 */
$post_name__in = ( ! empty( $args['post_name__in'] ) ) ? array_filter( (array) $args['post_name__in'] ) ) : [];
if ( ! empty( $post_name__in ) ) {
	$filter['bool']['must'][]['bool']['must'] = array(
		'terms' => array(
			'post_name.raw' => array_values( $post_name__in ),
		),
	);

	$use_filters = true;
}
```

felipeelia (changes requested):

includes/classes/Indexable/Post/Post.php
roborourke: I was trying to avoid too much duplication of code, I should add an `isset()` check in the foreach loop I think. I'll try it, and see if that solves the problem first.

roborourke: Derp - I see, this was actually the initial approach I'd thought of before writing the generic loop! I should have just removed it.

PR #2481, opened 2021-12-02
Instant Results feature

JakePT:
### Description of the Change

WIP

### Alternate Designs

WIP

### Benefits

WIP

### Possible Drawbacks

WIP

### Verification Process

WIP

### Checklist:

- [x] I have read the [**CONTRIBUTING**](/CONTRIBUTING.md) document.
- [x] My code follows the code style of this project.
- [x] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [ ] I have added tests to cover my change.
- [ ] All new and existing tests passed.

### Applicable Issues

WIP

### Changelog Entry

WIP

PR #2484, opened 2021-12-03
Add a "request" subcommand

felipeelia:
### Description of the Change

This is a take over of #2481:

To help when debugging, or otherwise wanting to send arbitrary requests to ElasticSearch, this adds a new request sub-command, that is somewhat similar to the idea of `wp db query` or the like.

For example `wp elasticpress request _cat/indices --method=GET`

### Changelog Entry

Added: WP-CLI `request` command. Props @joehoyle and @felipeelia

### DIFF
--- a/includes/classes/Command.php
+++ b/includes/classes/Command.php
@@ -1336,4 +1336,45 @@ class Command extends WP_CLI_Command {
 		 */
 		do_action( 'ep_cli_put_mapping', $indexable, $this->args, $this->assoc_args );
 	}
+
+	/**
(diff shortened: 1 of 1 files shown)


joehoyle:
Nice one!

oscarssanchez (commented):
Hi @felipeelia I have just one suggestion: 

* Have a list of allowed methods and check if the method matches any in the list. If not øn the list, display error message. For example: `wp elasticpress request _cat/indices --method=TEST` doesn't give an error message.

PR #2536, opened 2021-12-20
Fix usage of requiresReindex

felipeelia:
### Description of the Change

In #2491 we've introduced a new confirmation prompt on features activation. This PR adjusts its implementation so the prompt only appears in features that really require a reindex.

### Changelog Entry

Append to the `Enabling features that require a reindex will now ask for confirmation` line


assets/js/dashboard.js
```diff
@@ -605,7 +606,7 @@ $features.on('change', '.js-toggle-feature', function (event) {
 	const { value } = event.target;
 	const { requiresReindex, wasActive } = event.currentTarget.dataset;
 
-	if (requiresReindex && wasActive === '0' && value === '1') {
+	if (requiresReindex === '1' && wasActive === '0' && value === '1') {
 		container.style.display = 'block';
 	} else {
 		container.style.display = null;
```
Rahmon, line 612: @felipeelia we need to check if there is an element on container because we are retrieving only if it has `.requirements-status-notice--reindex` and it is causing an error when we try to access `container.style.display`:

```
Cannot read properties of null (reading 'style')
```

Maybe we can add an early return:

```js
if (!container) return;
```

Rahmon (approved):

PR #2547, opened 2022-01-11
Evaluate 0 and false string as false booleans as well

### LINKED ISSUE #2528: Evaluate 'falsey' values for ep_integrate

Currently, in lines like [this](https://github.com/10up/ElasticPress/blob/e4194a6b1b2012d275af5a0148c53be7c43a4664/includes/classes/Indexable.php#L388) we only disable integration if `ep_integrate` is a boolean false, making it integrate even if the value is `0` or a string `'false'`:

`if ( isset( $query->query_vars['ep_integrate'] ) && false === $query->query_vars['ep_integrate'] ) {`

Let's change that test to get all 'falsey' values, including zeros and 'false' as a string. We can achieve that by using:

`if ( isset( $query->query_vars['ep_integrate'] ) && ! filter_var( $query->query_vars['ep_integrate'], FILTER_VALIDATE_BOOL ) ) {`

oscarssanchez:
### Description of the Change

Fixes #2528 

### Checklist:

### DIFF
--- a/includes/classes/Feature/Comments/Comments.php
+++ b/includes/classes/Feature/Comments/Comments.php
@@ -110,7 +110,7 @@ class Comments extends Feature {
 			return $enabled;
 		}
 
-		if ( isset( $query->query_vars['ep_integrate'] ) && false === $query->query_vars['ep_integrate'] ) {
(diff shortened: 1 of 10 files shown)


Rahmon (commented):
What do you think about adding tests for `'0'` and `'false'`? E.g.: https://github.com/10up/ElasticPress/blob/develop/tests/php/features/TestComments.php#L119 

Other than that, looks good @oscarssanchez

Rahmon (approved):

PR #2560, opened 2022-01-18
Fix/facets tax query

### LINKED ISSUE #2502: BUG: Incompatibility with WooCommerce Brands
**Describe the bug**
When filtering on category pages with WooCommerce brands layered nav widget, nothing is shown because elasticpress adds a second tax query with field = slug and operator = and / or. While woocommerce brand adds/uses term_id as field and IN as operator.

Skärmavbild 2021-12-08 kl  08 57 58

The issue is in Facet.php and row 326 - 333. You add a tax query with field = slug and operator = AND / OR

```php
foreach ( $selected_filters['taxonomies'] as $taxonomy => $filter ) {
	$tax_query[] = [
		'taxonomy' => isset( $attribute_taxonomies[ $taxonomy ] ) ? $attribute_taxonomies[ $taxonomy ] : $taxonomy,
		'field'    => 'slug',
		'terms'    => array_keys( $filter['terms'] ),
		'operator' => ( 'any' === $settings['match_type'] ) ? 'or' : 'and',
	];
}
```

**Steps to Reproduce**
1. Add a brand
2. Add products to the newly added brand
3. Add the woocommerce brand layered nav widget in sidebar
4. Go to the category page and filter
5. Nothing is shown

**Expected behavior**
Don't add another tax query if a tax query for the same taxonomy already exists.

oscarssanchez:
### Description of the Change

This PR fixes #2502 by checking if we have a pre existing tax query done by external code (in this case, the Woocommerce brands plugin). If we do, there's no need to add another tax query with EP if using facets.

### Verification Process

1.- Have some brands and products associated with them
2.- Have a layared brand nav widget set
3.- Select one of the brands in the widget in the frontend
4.- Associated products should appear

### Checklist:

### Credits

Props @

### DIFF
--- a/includes/classes/Feature/Facets/Facets.php
+++ b/includes/classes/Feature/Facets/Facets.php
@@ -324,12 +324,15 @@ class Facets extends Feature {
 		}
 
 		foreach ( $selected_filters['taxonomies'] as $taxonomy => $filter ) {
-			$tax_query[] = [
-				'taxonomy' => isset( $attribute_taxonomies[ $taxonomy ] ) ? $attribute_taxonomies[ $taxonomy ] : $taxonomy,
-				'field'    => 'slug',
-				'terms'    => array_keys( $filter['terms'] ),
-				'operator' => ( 'any' === $settings['match_type'] ) ? 'or' : 'and',
-			];
+			$tax = isset( $attribute_taxonomies[ $taxonomy ] ) ? $attribute_taxonomies[ $taxonomy ] : $taxonomy;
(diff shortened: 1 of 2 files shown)


felipeelia:
@oscarssanchez talked to @brandwaffle about this one and it would probably be better to change our "filter_" to "ep_filter_", so we avoid any future conflict that may arise. Obviously, that would be a filterable string, so people could go back to "filter_" if they want. What do you think?

oscarssanchez:
Assigning for review @felipeelia

PR #2563, opened 2022-01-20
Displaying results counts for facet options in Instant Results

### LINKED ISSUE #2526: Support displaying results counts for facet options in Instant Results
Currently Instant Results does not support displaying the number of results next to each facet option. The ability to display these numbers should be added to the feature as a setting.

A simple on/off switch for displaying result counts on facets in the feature settings.

The data should already be present in the response, but the facet component either needs to add the number in parentheses to the option labels or the checkbox list component could be updated to support supplementary text for each option that could be output and styled separately, a la the issue and pull request counts on GitHub, above.

JakePT:
### Description of the Change

Adds result counts for facet options in Instant Results.

Closes #2526.

### Alternate Designs

Design-wise there could be any number of variations, but as the default I stuck with simple numbers in parentheses. To support users and themes styling the counts differently, such as in a bubble or similar, the counts have their own element, and parentheses are added with CSS so that they can be removed with CSS. From testing with VoiceOver there was no difference in accessibility in using CSS for the parentheses.

With this PR the counts are enabled by default, with no setting to choose to enable or disable them, however they occupy their own element so they can be easily hidden with a single line of CSS.

### Possible Drawbacks

N/A

### Verification Process

With this PR checked out the counts should be visible next to filter options in Instant Results. Selecting filter options should narrow down the number of results to the number next to each option, and the counts next to any remaining options should be updated to reflect the number of matching results in the current result set.

### Checklist:

### Changelog Entry

Displaying results counts for facet options in Instant Results.

### Credits

Props @JakePT

### DIFF
--- a/assets/css/instant-results.css
+++ b/assets/css/instant-results.css
@@ -1,4 +1,5 @@
 @import "instant-results/utilities.css";
+@import "instant-results/checkbox.css";
 @import "instant-results/input.css";
 @import "instant-results/modal.css";
 @import "instant-results/options-list.css";
--- a/assets/css/instant-results/checkbox.css
+++ b/assets/css/instant-results/checkbox.css
@@ -1,0 +1,7 @@
+.ep-search-checkbox__count::before {
+	content: "(";
+}
+
+.ep-search-checkbox__count::after {
+	content: ")";
+}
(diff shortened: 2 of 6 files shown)


felipeelia:
@JakePT , the code overall looks good but the linter is giving a warning: `Expected JSDoc block lines to be aligned`. Do you mind fixing that one?

JakePT:
@felipeelia I've pushed fixes for the JSDoc warnings.

PR #2564, opened 2022-01-21
Show WP-CLI progress on sync page

Rahmon:
### Description of the Change

#2489 

### Checklist:

### Credits

Props @


Rahmon:
@felipeelia This PR changes the code to show the WP-CLI sync progress on the Sync page, however, it's necessary to handle the case when the WP-CLI sync is interrupted by Ctrl + C (for example). Currently, there is no difference compared to when the sync is finished. It's necessary to send a different response from AJAX to be possible to know when the sync finished or was interrupted. 

Other than that, it's not possible to stop the WP-CLI from the Sync page. Maybe this case could be done in another PR.

PR #2571, opened 2022-01-26
Add helper get_total_objects_for_query_from_db() which counts directly from the DB

rebeccahum:
### Description of the Change

We're noticing that on very large sites, the query to count the total objects in `get_total_objects_for_query()` dies. This PR adds an abstracted helper method `get_total_objects_for_query_from_db()` that breaks it down using `wp_count_posts` to count the posts directly from the DB.

### Alternate Designs

N/A.

### Possible Drawbacks

A bit more processing during indexing if the count query has died, but at least the user will have an idea of the number of posts, rather than "0".

### Verification Process

On a verrrry large site where it says `0` as the total objects, i.e. `Processed x/0`, apply the PR and expect it to return a number.

### Checklist:

### Changelog Entry

Fixed: Add an alternative way to count total posts on larger DBs during indexing

### Credits

Props @rebeccahum

### DIFF
--- a/includes/classes/Indexable/Post/Post.php
+++ b/includes/classes/Indexable/Post/Post.php
@@ -189,7 +189,42 @@ class Post extends Indexable {
 			$object_counts[ $cache_key ] = ( new WP_Query( $normalized_query_args ) )->found_posts;
 		}
 
+		if ( 0 === $object_counts[ $cache_key ] ) {
+			// Do a DB count to make sure the query didn't just die and return 0.
+			$db_post_count = $this->get_total_objects_for_query_from_db( $normalized_query_args );
+
+			if ( $db_post_count !== $object_counts[ $cache_key ] ) {
+				$object_counts[ $cache_key ] = $db_post_count;
+			}
+		}
+
 		return $object_counts[ $cache_key ];
+	}
+
+	/**
+	 * Get total posts from DB for a specific query based on it's args.
+	 *
+	 * @param array $query_args The query args.
+	 * @return int The total posts.
+	 */
+	public function get_total_objects_for_query_from_db( $query_args ) {
+		$post_count = 0;
+
+		if ( ! isset( $query_args['post_type'] ) ) {
+			return $post_count;
+		}
+
+		foreach ( $query_args['post_type'] as $post_type ) {
+			$post_counts_by_post_status = wp_count_posts( $post_type );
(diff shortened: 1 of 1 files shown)


includes/classes/Indexable/Post/Post.php
```diff
@@ -189,9 +189,44 @@ private function get_total_objects_for_query( $query_args ) {
...
+		$post_count = 0;
+
+		if ( ! isset( $query_args['post_type'] ) ) {
+			return $post_count;
+		}
+
+		foreach ( $query_args['post_type'] as $post_type ) {
+			$post_counts_by_post_status = wp_count_posts( $post_type );
+			foreach ( $post_counts_by_post_status as $post_status => $post_status_count ) {
+				if ( ! in_array( $post_status, $query_args['post_status'] ) ) {
```
felipeelia, line 220: Do you mind changing this to 
`in_array( $post_status, $query_args['post_status'], true )` @rebeccahum? As is it triggers a warning in PHPCS and as we are checking strings, I don't think it will be a problem.

includes/classes/Indexable/Post/Post.php
```diff
@@ -189,9 +189,44 @@ private function get_total_objects_for_query( $query_args ) {
...
+			}
+		}
+
 		return $object_counts[ $cache_key ];
 	}
 
+	/**
+	 * Get total posts from DB for a specific query based on it's args.
+	 *
+	 * @param array $query_args The query args.
```
felipeelia, line 207: Can you please add a `@since 4.0.0` in the doc block? Thanks!

includes/classes/Indexable/Post/Post.php
```diff
@@ -189,9 +189,44 @@ private function get_total_objects_for_query( $query_args ) {
...
+	 * @return int The total posts.
+	 */
+	public function get_total_objects_for_query_from_db( $query_args ) {
```
felipeelia, line 210: As it is only called by a private method (`get_total_objects_for_query()`), I guess we can change the method visibility to be also private.
I'd prefer to have them both as `protected` so people can extend the class without having to copy/paste code, though. Right now, anyone extending this class and using the `query_db` method will have to make a copy of `get_total_objects_for_query()` to have it working.
Thoughts?

felipeelia (changes requested):

rebeccahum:
@felipeelia Thanks for your feedback! All have been addressed now.

felipeelia (approved):

PR #2572, opened 2022-01-26
Features selection during install

### LINKED ISSUE #2570: Features selection BEFORE initial sync

Right now, the user installing the plugin will set the Elasticsearch server, run a complete sync, and only then will be able to select which features should be used, what potentially triggers a second complete reindex (autosuggest is one case, for example.)

The user should be able to select the features *before* that initial sync, so it already sends the correct mapping, fields, and contents.

**Designs**

TBD

felipeelia:
### Description of the Change

This PR introduces a new step in the installation process, so users can select features they want to install before the initial sync.

Closes #2570


felipeelia:
@JakePT The build process seems to be failing to this branch, and it _seems_ that #2631 would fix it. Do you mind giving a look? Thanks!

JakePT (dismissed):
Build completing without linting errors after merging #2631 and one small fix.

JakePT (approved):

PR #2577, opened 2022-01-28
Remove unnecessary order adjustment for WC queries

### LINKED ISSUE #2473: QUESTION: woocommerce product search results sort by stock_status
Is there a way to use _stock_status metadata in a woocommerce product to sort search results such that it shows all out-of-stock products last? 

When ElasticPress is not active, the following filter works correctly:
```php
add_filter( 'woocommerce_get_catalog_ordering_args', 'show_instock_first', 9999 );
function show_instock_first( $args ) {
   $args['orderby'] = 'meta_value';
   $args['order'] = 'ASC';
   $args['meta_key'] = '_stock_status';
   return $args;
}
```

But ElaticPress seems to ignore that. 

If I step on the function get_es_posts in includes/classes/Indexable/Post/QueryIntegration.php I can see that the $query is setup correctly, and it has the right _stock_status meta_query set. However, it seems that $query->query_vars has 'orderby' set to false and elastic press never picks up the _stock_status meta query from the query. 

Any idea how to make this work? 
Thanks

felipeelia:
### Description of the Change

This PR removes an adjustment in the `orderby` parameter that EP was doing on WooCommerce queries and is no longer needed. 

In older versions of WC (apparently until [WC 3.0.9](https://github.com/woocommerce/woocommerce/releases/tag/3.0.9), released in June 2017), it was defaulting to `menu_order title => ASC`: https://github.com/woocommerce/woocommerce/blob/3.0.9/includes/class-wc-query.php#L449

Newer versions are already ordering by relevance DESC, as ElasticPress does.

Closes #2473

### Changelog Entry

Changed: to comply with modern WooCommerce behavior, ElasticPress does not change the orderby parameter anymore.

### Credits

Props @felipeelia and @beazuadmin

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,11 +34,19 @@ ElasticPress has an in depth documentation site. [Visit the docs ☞](http://10u
 
 ElasticPress FAQs and tutorials can be found on our support site. [Visit the support site ☞](https://elasticpress.zendesk.com/hc/en-us)
 
-## Requirements
+## Requirements and Compatibility
+
+### Requirements
+
+ElasticPress requires these software with the following versions:
 
 * [Elasticsearch](https://www.elastic.co) 5.0+ **ElasticSearch max version supported: 7.9**
 * [WordPress](http://wordpress.org) 3.7.1+
 * [PHP](https://php.net/) 5.6+
+
+### Compatibility
+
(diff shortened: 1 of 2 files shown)


felipeelia:
Add to readme WC compatibility (last 2 major versions)

PR #2585, opened 2022-02-04
Dynamic Indexing (for 4.0)

### LINKED ISSUE #2519: Dynamic number of documents sent to Elasticsearch

Currently, we send a fixed number of posts/users/terms/comments (documents) to Elasticsearch. That fixed number can lead us to two different problems:
1. Sending too many documents can return a "Request Entity Too Large", i.e., EP tried to send too many documents in the same _bulk request OR
2. The opposite: sending just 10 documents when we could be sending 100.

We can fix that by creating a "buffer". Keeping in mind that Elasticsearch defaults to 100MB as max_content_length, we would
1. Create an empty string (the request body), let's say `$request_body`
2. For each document check if its size + `strlen( $request_body )` exceeds max_content_length
3. If it does not exceed, add it to $request_body and move to the next one
4. If it does exceed, send the request, empty the buffer and add posts to it.

We should also account for timed-out requests. If it happens, we can split the request in two and try again.

That way we would be fixing both scenarios, as we wouldn't send a request too big nor underuse a request that could carry more documents.

**Additional Context**

This is a rethink of #1949

felipeelia:
### Description of the Change

TBD

Closes #2519

### Changelog Entry

TBD

### Credits

TBD


felipeelia:
This is the log of a sync using the code in 1395043: https://pastebin.com/DAbD7G5B

PR #2587, opened 2022-02-08
Fix/update ep version

mahammedzkhan:
### Description of the Change

This update query to update the ElasticPress version kept running all the time on our sites, it also doesn't make any sense to always run it even when you're running the latest version. It caused a lot of `PHP Warning: mysqli_real_connect(): (HY000/1040): Too many connections`

### Alternate Designs

None

### Possible Drawbacks

Don't see any drawbacks of this optimization.

### Verification Process

Implemented this change and then snooped around the website & wp-admin, no more faulty update queries.

### Checklist:

- [X] I have updated the documentation accordingly.

### Changelog Entry

Changed - only update ElasticPress when older version is detected in DB.

### Credits

Props @

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,11 @@ All notable changes to this project will be documented in this file, per [the Ke
 
 ## [Unreleased]
 
+## [3.6.7] - 2021-02-08
+
+## Changed
+- Only update ElasticPress when older version is detected in DB.
+
 ## [3.6.6] - 2021-12-20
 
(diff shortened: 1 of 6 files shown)


felipeelia:
Hi @mahammedzkhan,

What version of WordPress are you using? Looking at [update_option source code](https://developer.wordpress.org/reference/functions/update_option/#source) it has a `$value === $old_value` check that would prevent the update in the database if the values are equal.

PR #2601, opened 2022-02-18
#2599: Limit users to the current site

### LINKED ISSUE #2599: BUG: User search request yields users who do not exist in the current site
**Describe the bug**
When searching for users, the ElasticPress query may yield users who do not exist in the current site.

This is not obvious by default, but when using some more involved configuration, for example, using autosuggest/ngram.

**Steps to Reproduce**

1. Have a multisite installation.
1. Create a user with login `test`.
1. Do not add that user to any site.
1. On some site, navigate to the Users page, `/wp-admin/users.php`.
1. Search for "test".
1. You will see the user appear as a result. This is wrong!

In order for this to fail, you will need something like the following filter callback running in your site:

```php
add_filter( 'ep_user_formatted_args_query', function ( $query, $args ) {

	$query['bool']['should'][] = [
		'multi_match' => [
			'query' => $args['s'] ?? $args['search'],
			'fields' => [
				'user_login.suggest',
			],
		],
	];

	return $query;
}, 10, 2 );
```

**Expected behavior**

The query should only yield users who exist in the current site.

**Screenshots**

**Environment information**
This bug is independent of the environment.

**Additional context**

tfrommen:
### Description of the Change

This PR adds a conditional to the User indexable class to ensure that only users that exist in the current site will be included in the search results.

This change is similar to this [section in the `WP_User_Query` class](https://github.com/WordPress/wordpress-develop/blob/c98f4c04db13a1c3db14345db4cff116dd01bf51/src/wp-includes/class-wp-user-query.php#L566-L572), which also captures the case that no user roles have been specified:

```php
			// If there are no specific roles named, make sure the user is a member of the site.
			if ( empty( $role_queries ) ) {
				$role_queries[] = array(
					'key'     => $wpdb->get_blog_prefix( $blog_id ) . 'capabilities',
					'compare' => 'EXISTS',
				);
			}
```

The difference to ElasticPress is that `capabilities.<ID>.roles` will always exist. If the user does not have any roles on the site (i.e., they have not been added to the site), the field includes the `0` term only.

Closes #2599

### Alternate Designs

No idea.

### Possible Drawbacks

I can't think of any.

### Verification Process

Following the steps given in #2599, with this change in place, you will no longer see the `test` user as result.

### Checklist:

### Changelog Entry

Fixed user search results include users who do not exist in the current site.

### Credits

Props @tfrommen

### DIFF
--- a/includes/classes/Indexable/User/User.php
+++ b/includes/classes/Indexable/User/User.php
@@ -173,6 +173,17 @@ class User extends Indexable {
 
 					$use_filters = true;
 				}
+
+				// If there are no specific roles named, make sure the user is a member of the site.
+				if ( ! $use_filters ) {
+					$filter['bool']['must_not'][] = array(
+						'term' => array(
+							'capabilities.' . $blog_id . '.roles' => 0,
+						),
+					);
+
+					$use_filters = true;
+				}
 			}
 		}
 


felipeelia:
@tfrommen just a heads up that I'm closing this PR in favor of #2670 that has your code + some tests and also some code that prevents the array from being created with a wrong `0` in it. If you have the time, it would be great if you could give it a look and test. Thanks!

Issue #2613, opened 2022-02-25
Migrate all React functionality to use @wordpress/element with dependency extraction


JakePT:
**Is your enhancement related to a problem? Please describe.**
Instant Results, introduced in 4.x, is powered by the version of React included in WordPress by importing `@wordpress/element` with the WordPress dependency extraction plugin enabled. However the Synonyms editor uses a (now) older version of React that is included as an NPM dependency of the plugin.

During the development of Instant Results several React related bugs would occur when building assets with `npm run watch`. A possible cause of this issue is a mismatch between the version of React in WordPress and the version used as a dependency of other features. For this reason, as well as consistency and reducing the number of dependencies, I propose migrating all React-based features in the plugin to using `@wordpress/element` with dependency extraction.

**Describe the solution you'd like**
- Update the Synonyms feature to use  `@wordpress/element` with dependency extraction instead of a separate React dependency.
- Remove unused React dependencies from the plugin.

Best case scenario this will be a handful of lines worth of changes to the import statements and package.json dependencies, but the results will need to be tested.

**Describe alternatives you've considered**
A solution could be to go in the other direction, and use our own React dependency for these features instead of what's included in WordPress, but the trouble with that approach is that with its current implementation Instant Results uses other WordPress libraries built on top of `@wordpress/element`, such as SlotFills, that will require whichever version of React is included in WordPress.

JakePT:
As noticed on #2631, custom search results will need to replace `react-beautiful-dnd` before it can be updated to use the version of React in WordPress, which is no longer compatible as of 5.9.

PR #2631, opened 2022-03-03
Node packages update

felipeelia:
Mainly updating 10up-toolkit and adjusting things around it

### Changelog Entry

Security - Node packages update

### Credits

Props @felipeelia


JakePT:
@felipeelia It looks like the drag and drop sorting of Custom Results no longer works (although it's not actually throwing an error). I notice that it's using the `react-beautiful-dnd` library. This has been the cause of a couple of sites not being able to update to 5.9 as it's not compatible with React 17.x which is used by WordPress 5.9. So I think updating the version of React here to 17.x might've broken it. We may need to downgrade the version used by these EP features to 16. This raises an issue with #2613 that will need to be solved before that can be done.

Everything else seemed to work ok though, and I agree with the change you made to the strict value rule, at least while we're dealing with lots of relatively old CSS. Maybe after a refactor or redesign we could revisit the rule.

JakePT (approved):
Everything seems to be working fine now.

PR #2637, opened 2022-03-07
Add generate_mapping to Post and Comment Indexable

rebeccahum:
### Description of the Change

The User and Term Indexable classes have the `generate_mapping()` function: 
https://github.com/10up/ElasticPress/blob/9a04a23c3f4e85c7f8c67957df5777ef688ce06e/includes/classes/Indexable/User/User.php#L753
https://github.com/10up/ElasticPress/blob/424ba22ff8e2af756ece0ec7654a6486af0f3487/includes/classes/Indexable/Term/Term.php#L587

It would be nice if the Post and Comment Indexable could have that same function, so we can piggyback onto it with some code via `$indexable->generate_mapping()`.

### Alternate Designs

N/A.

### Possible Drawbacks

N/A.

### Checklist:

### Changelog Entry

Added - generate_mapping() to post and comment indexable

### Credits

Props @rebasaurus


felipeelia:
@rebeccahum I guess if we are aiming to make that a pattern across all Indexables, we should change the parent Indexable class, perhaps with a check to keep backward compatibility. What do you think?

rebeccahum:
@felipeelia Let me know if this is what you had in mind.

PR #2645, opened 2022-03-08
Adjust search_template type and initial value + move save to end of sync

felipeelia:
### Changelog Entry

Changed: Instant Results: type and initial value of search template and move save to the end of sync.

### Credits

Props @felipeelia


JakePT (approved):
lgtm

PR #2670, opened 2022-03-21
Limit users to the current site

### LINKED ISSUE #2599: BUG: User search request yields users who do not exist in the current site

felipeelia:
### Description of the Change

This PR is an alternative version of #2601, fixing the problem outlined in #2599 and also preventing an empty `roles` field from being added.

Unit and e2e tests should be added here before merging.

Closes #2599

### Changelog Entry

Fixed: User search results include users who do not exist in the current site.

### Credits

Props @tfrommen and @felipeelia


includes/classes/Indexable/User/User.php
```diff
@@ -173,6 +173,22 @@ public function format_args( $query_vars, $query ) {
 
 					$use_filters = true;
 				}
+
+				// If there are no specific roles named, make sure the user is a member of the site.
+				if ( ! $use_filters ) {
```
tfrommen, line 178: After I submitted my patch (see #2601), I was wondering if relying on `$use_filters` was a good idea. Right now, this works fine, just because there are no conditions that would set `$use_filters` to `true` except for the ones related to roles. However, if there ever was some new code to be added after the inital `$use_filters = false;` and the roles conditions, this would no longer be true.

Instead of checking `$use_filters`, one could do it like so (which is a little more verbose, but absolutely specific to what we're after):

```suggestion
				if ( empty( $query_vars['role'] ) && empty( $query_vars['role__in'] ) && empty( $query_vars['role__not_in'] ) ) {
```

This should work just fine with your tests as well.

felipeelia: You have a good point @tfrommen. I went ahead and used that to do a small refactor in the entire if block. Hopefully, it'll make the code more future-proof and also easier to be understood by new programmers. Thanks!

oscarssanchez (approved):

Issue #1440, opened 2019-08-13
Add meta/range capabilities to facet feature.


tlovett1:
Right now facet feature just supports taxonomies. Let's add support for meta and ranges.

entttom:
Thats an awesome idea!

razorfrog:
+1

shmaltz:
+1

felipeelia:
As support to meta was recently introduced, I'm breaking up the range part into a new issue: #3118.

PR #2682, opened 2022-03-27
Add support for multiple aggregations

### LINKED ISSUE #636: Support multiple aggregations
It would be nice to add a support for multiple aggregations.

felipeelia:
### Description of the Change

Currently, the plugin only supports a single array as the `aggs` parameter of WP_Query. This PR changes that to also support an array of aggregations. This will be needed to implement #1440 in a cleaner way.

Closes #636

### Changelog Entry

Added: Support for an array of aggregations in the `aggs` parameter of WP_Query.

### Credits

Props @felipeelia

### DIFF
--- a/includes/classes/Indexable/Post/Post.php
+++ b/includes/classes/Indexable/Post/Post.php
@@ -1800,9 +1800,13 @@ class Post extends Indexable {
 		/**
 		 * Aggregations
 		 */
-		if ( isset( $args['aggs'] ) ) {
-			// An array of aggregations.
-			if ( isset( $args['aggs'][0] ) ) {
+		if ( ! empty( $args['aggs'] ) && is_array( $args['aggs'] ) ) {
+			// Check if the array indexes are all numeric.
+			$agg_keys          = array_keys( $args['aggs'] );
+			$agg_num_keys      = array_filter( $agg_keys, 'is_int' );
+			$has_only_num_keys = count( $agg_num_keys ) === count( $args['aggs'] );
+
+			if ( $has_only_num_keys ) {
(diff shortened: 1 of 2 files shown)


includes/classes/Indexable/Post/Post.php
```diff
@@ -1800,24 +1800,15 @@ public function format_args( $args, $wp_query ) {
...
-
-			// Add/use the filter if warranted
-			if ( isset( $agg_obj['use-filter'] ) && false !== $agg_obj['use-filter'] && $use_filters ) {
-
-				// If a filter is being used, use it on the aggregation as well to receive relevant information to the query
-				$formatted_args['aggs'][ $agg_name ]['filter'] = $filter;
-				$formatted_args['aggs'][ $agg_name ]['aggs']   = $agg_obj['aggs'];
+		if ( isset( $args['aggs'] ) ) {
+			// An array of aggregations.
+			if ( isset( $args['aggs'][0] ) ) {
```
oscarssanchez, line 1805: Would it be better to check if this is an array with `is_array()` ? Right now I don't think it would do any harm, but i'm thinking if there's ever a case where we offer an aggregations filter for example, and people can manipulate the array indexes. If there's not `$args['aggs'][0] `but there's `$args['aggs'][1]` this code will not run.

oscarssanchez (changes requested):

includes/classes/Indexable/Post/Post.php
felipeelia: I've made the changes we discussed in slack @oscarssanchez . Do you mind giving yet another look at it? Thanks!

oscarssanchez (approved):

PR #2683, opened 2022-03-27
[WIP] Facet by meta

felipeelia:
### Description of the Change

This PR adds support for meta fields in the Facet widget.

This is currently a Work In Progress and should *not* be merged.

Related to #1440 (it does not close the issue because the range implementation should be done in another PR)

### Changelog Entry

Added: Support for meta fields in Facet widgets.

### Credits

Props @felipeelia

### DIFF
--- a/includes/classes/Feature/Facets/Facets.php
+++ b/includes/classes/Feature/Facets/Facets.php
@@ -250,98 +250,8 @@ class Facets extends Feature {
 			return;
 		}
 
-		$taxonomies = get_taxonomies( array( 'public' => true ), 'object' );
-
-		/**
-		 * Filter taxonomies made available for faceting
-		 *
-		 * @hook ep_facet_include_taxonomies
-		 * @param  {array} $taxonomies Taxonomies
-		 * @return  {array} New taxonomies
-		 */
-		$taxonomies = apply_filters( 'ep_facet_include_taxonomies', $taxonomies );
-
-		if ( empty( $taxonomies ) ) {
-			return;
-		}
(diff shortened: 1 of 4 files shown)


Zyles:
I tried this update with Woocommerce and there are some weird results.

A boolean ACF field creates a facet like this:

image


Where I would rather it be a checkbox like:
`[ ] Meta field name`

It also did not work clicking on any of the boxes. They just added `#` to the URL.

Second test was a text field ACF type. It did spit out the options. But one value was null (no text), none were clickable and just added # to URL.

The choices were also not conditional in that if I go into a category with a product that does not have any of these values, the facet would still show. If there are no options matching the current query the options should be hidden.

The last test was a select field type. Only 4 names were shown even if the choices are over 100. And the first choice was null. It also had the same problem if I went into a category with no products using this meta at all, they would still show up in the widget.

Maybe make the meta facet widget into its own widget instead of using the current widget? When checking meta in the widget all the other options are not used and could just cause confusion.

The meta widget could use more options such as selecting if the facet should be dropdown, checkboxes, have a search window etc. As seen in [YITH Ajax Search](https://yithemes.com/themes/plugins/yith-woocommerce-ajax-product-filter/):
image

I think some inspiration from this plugin could be good.

felipeelia:
@Zyles, thanks for giving this one a try but as the title says, it is still a Work In Progress. We are aware of the fact that its current status is far from being testable :)

felipeelia:
Closing this in favor of #2919.

PR #2722, opened 2022-04-19
Facet block

### LINKED ISSUE #2689: Facets need to become a WP Block
Having in mind that we also plan to expand Facets to handle meta fields and have a range type of widget (see #1440), we can take this opportunity to better structure the code of this feature. Here are (some of) the steps:

- [x] Create a common code to handle both renderings: blocks and widgets. That way, if we want to make a change in the outputted HTML, we'd have to change it in only one place. #2712
- [x] Create a proper WP Block, with the same settings we have in the widgets. If possible, it'd be nice if we could keep that list of fields in just one place as well.

felipeelia:
### Description of the Change

This PR adds a block version of the Facet widget.

Closes #2689

### Changelog Entry

Added: Facet block (previously only available as a widget.)

### Credits

Props @felipeelia


includes/classes/Feature/Facets/Facets.php
```diff
@@ -526,4 +537,84 @@ protected function get_filter_name() {
...
+			]
+		);
+	}
+
+	/**
+	 * Check permissions of the /facets/taxonomies REST endpoint.
+	 *
+	 * @return WP_Error|true
+	 */
+	public function check_facets_taxonomies_rest_permission() {
```
JakePT: @felipeelia With this becoming a block I'd imagine users would be able to add it to posts and pages, but those users might not have this capability. Would this just need to be open to all logged in users? Not sure how to narrow it down to users that can access blocks.

felipeelia: Good point @JakePT. I've opened it up to all logged-in users now.

JakePT:
@felipeelia LGTM. By the looks of it I think it just needs an ignore/disabled for the linting rule on the `dangerouslySetInnerHtml` line, and to be up to date with develop.

JakePT (dismissed):
@felipeelia LGTM. By the looks of it I think it just needs an ignore/disabled for the linting rule on the `dangerouslySetInnerHtml` line, and to be up to date with develop.

felipeelia:
@JakePT I've pushed a new commit improving the preview a bit. Do you mind giving it a quick look? Thanks!

JakePT (commented):
@felipeelia I wonder if the message in the preview could explain a little bit more about why the preview isn't available? Something like:

>No categories found for available results.

Just a thought. Otherwise looks good.

JakePT (dismissed):

Issue #2706, opened 2022-04-08
BUG: Installation Workflow and initial sync


felipeelia:
**Describe the bug**

There are actually two:

- [x] During the installation process, if the user does not select any feature during step 3, the "Save Feature" button only reloads the page. 
- [x] During the initial sync (the one automatically made after clicking on the step 4 button), the screen won't be properly set as done, i.e., make the progress bar green and hide the pause/stop buttons. (There is a chance this is fixed by #2704)

**Additional context**

This can be reproduced in an easier way using code from #2705.

**Credits**

Props @jakemgold.

brandwaffle:
@tlovett1 are you thinking if someone checks none of the boxes we should warn them that nothing will be active and then move past install, or should we be indicating we're requiring they check at least one box?

JakePT:
@felipeelia As discussed with @brandwaffle we will address the first point by displaying a standard browser alert that reads:

>It looks like you’re trying to use ElasticPress’s advanced features only. If you’d like to activate basic search, please select Cancel and activate the Post Search Feature. Otherwise, please click Ok to configure advanced features.

And continuing will proceed to the Features page without starting a sync.

PR #2734, opened 2022-05-02
Fix issues with feature activation during install

JakePT:
### Description of the Change

This change fixes two issues:

1. If no features were selected installation would not proceed, but without any feedback (#2706).

The fix is that if no features are selected the user is asked to confirm, and if they do install is skipped with the same behaviour as following the _Skip Install_ link. Additionally, when skipping installation any features that are activated automatically are deactivated so that skipping installation results in no active features.

2. If a feature that is activated by default is deselected during installation it would be activated anyway.

This is fixed by deactivating any features that can be selected during install if they are not selected. This is necessary as certain features are activated by default before installation has even started.

### Verification Process

Before each test clear all ElasticPress settings from the database for a fresh install. Each test should also be repeated for a multisite network activation.

1. When prompted to enter host settings click "Skip Install". You should be taken to the _Features_ page where no Features are active but enabling them is disabled until host settings are entered.

---

1. When prompted to enter host settings, click the button to go to Settings and enter the appropriate settings. 
2. When prompted to select features click "Skip Install". You should be taken to the _Features_ page where no Features are active.

---

1. When prompted to enter host settings, click the button to go to Settings and enter the appropriate settings. 
2. When prompted to select features deselect all features and click "Save Features". You should be prompted with a dialog box asking you to confirm.
3. On cancelling the dialog box should close and nothing else should happen.
4. On confirming that you want to proceed you should be taken to the _Features_ page where no Features are active.

---

1. When prompted to enter host settings, click the button to go to Settings and enter the appropriate settings. 
2. When prompted to select features, deselect one of the features that are already selected and select a feature that isn't already selected. Click "Save Features". You should be prompted to index content.
3. When prompted to index content, click "Skip Install". You should  You should be taken to the _Features_ page where he selected features should match what was selected during install i.e. the selected feature should be active, and the deselected feature shouldn't be.

---

1. When prompted to enter host settings, click the button to go to Settings and enter the appropriate settings. 
2. When prompted to select features, deselect one of the features that are already selected and select a feature that isn't already selected. Click "Save Features". You should be prompted to index content.
3. On indexing content you should be taken to the Sync page and a Sync should start.
4. After sync is complete, visit the Features page. The selected features should match what was selected during install.

### Checklist:

### Changelog Entry

- Fixed - An issue where selecting no features during install would just cause the install page to reload without any feedback.
- Fixed - An issue where deselecting a feature during install would not stop that feature from being activated.

### Credits

Props @JakePT

### DIFF
--- a/assets/js/dashboard.js
+++ b/assets/js/dashboard.js
@@ -13,7 +13,7 @@ import { __ } from '@wordpress/i18n';
  */
 const {
 	ajaxurl,
-	epDash: { syncUrl },
+	epDash: { skipUrl, syncUrl },
 } = window;
 
 /**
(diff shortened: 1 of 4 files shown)


JakePT:
Note that this addresses one of the issues in #2706, but not the other one, so it does not close that issue.

felipeelia:
@JakePT do you think it would be possible to add some tests for this?

PR #2740, opened 2022-05-05
Fix type error in console when searching

### LINKED ISSUE #2725: BUG: Type error in console when searching
**Describe the bug**
When you have autosuggest enabled and press enter to search for something, a type error briefly appears before being redirected to the results page.

**Steps to Reproduce**

1. Go to a site with an elastic press search engine with autosuggest working.
2. Put something random in the search bar.
3. Without using your arrow keys, press enter.

**Expected behavior**
A type error briefly appearing in the console. For me it just said: 'uncaught' but in another environment it gave Uncaught TypeError: n[o] is undefined

**Additional context**
The solution to this problem is in autosuggest.js between line 480 and 481:

`case 13: // Enter`
`  if (results[currentIndex].classList.contains('selected')) {`

add this:

`if(typeof currentIndex === 'undefined')break;`

MarijnvSprundel:
### Description of the Change

Add a check if currentindex is undefined then break, this prevents a error briefly appearing in the console.

Closes #2725

### Alternate Designs

None, no designs were made.

### Possible Drawbacks

None, just a quick check.

### Verification Process

Well firstly I got tasked to fix this bug on the website, which quickly led me to the autosuggest.js. But this process was painful since the minified version was throwing the errors, so I had to dig trough a one line file.

The fix I currently have implemented inserts this line of code between the other lines using regex, but it would be better for this fix to be in the plugin itself.

### Checklist:

### Changelog Entry

- Fixed a brief type error in the console

### Credits

Props @MarijnvSprundel

### DIFF
--- a/assets/js/autosuggest.js
+++ b/assets/js/autosuggest.js
@@ -479,7 +479,9 @@ function init() {
 				break;
 			case 13: // Enter
 				if (results[currentIndex].classList.contains('selected')) {
-					if(typeof currentIndex === 'undefined')break;
+					if (typeof currentIndex === 'undefined') {
+						break;
+					}
 					// navigate to the item defined in the span's data-url attribute
 					selectItem(input, results[currentIndex].querySelector('.autosuggest-link'));
 				}
--- a/assets/js/autosuggest.js
+++ b/assets/js/autosuggest.js
@@ -479,6 +479,7 @@ function init() {
 				break;
 			case 13: // Enter
 				if (results[currentIndex].classList.contains('selected')) {
+					if(typeof currentIndex === 'undefined')break;
 					// navigate to the item defined in the span's data-url attribute
 					selectItem(input, results[currentIndex].querySelector('.autosuggest-link'));
 				}
--- a/assets/js/autosuggest.js
+++ b/assets/js/autosuggest.js
@@ -478,10 +478,10 @@ function init() {
 				}
 				break;
 			case 13: // Enter
-				if (results[currentIndex].classList.contains('selected')) {
-					if (typeof currentIndex === 'undefined') {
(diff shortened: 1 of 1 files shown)


felipeelia:
Hi @MarijnvSprundel, thanks for Pull Request! Do you mind addressing the code standard adjustment highlighted by the automatic check so we can have this one merged?

felipeelia:
Hi @MarijnvSprundel, giving the PR a try, the error is still happening even with the code in place. That makes sense, as the code is checking for `currentIndex` after it was already used in `results[currentIndex]`. We could move up that if you added, making it like:
```
if (typeof currentIndex === 'undefined') {
	break;
}
if (results[currentIndex].classList.contains('selected')) {
	// navigate to the item defined in the span's data-url attribute
	selectItem(input, results[currentIndex].querySelector('.autosuggest-link'));
}
```
but that is not necessary, as we can make use of the [optional chaining](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining) feature instead. With that, the only change we would need to fix the problem is to replace
```
if (results[currentIndex].classList.contains('selected')) {
```
with
```
if (results[currentIndex]?.classList.contains('selected')) {
```

Do you mind changing the PR to look like that? Thanks!

felipeelia:
Hi @MarijnvSprundel, reviewing this one again and saw b493f32. Any reason why you decided not to use optional chaining instead of the if?

MarijnvSprundel:
I tested it with the optional chaining but it still seemed to throw the error, but maybe I implemented it the wrong way. Feel free to change it.

felipeelia:
Closing in favor of #2754.

PR #2748, opened 2022-05-09
Syncs: status and error handling

### LINKED ISSUE #2737: BUG: Stuck WP-CLI sync generates a repeated output in the Sync Page
**Describe the bug**

Currently, if a WP-CLI sync process gets killed without clearing the index meta option, any visit to the Sync Page will output the same message over and over again, wrongly suggesting that the index is still running. That happens because the Sync page does not communicate directly to the WP-CLI PHP process but rather reads from an option stored in the database.

**Steps to Reproduce**
2. Run `php -d max_execution_time=1 "$(which wp)" elasticpress index --per-page=2` so it times out
3. Visit the Sync Page
4. Click on "Show Log"
5. See the same message being outputted
6. Run `wp elasticpress clear-index`
7. See the sync page stopping to output the error

**Expected behavior**
Ideally, the plugin:
- Would not output the same message several times. The second and following times it'd update the same message it could say "Still processing (X seconds)", for example.
- WP-CLI sync wouldn't be stuck. Initial tests with a code like the following clear the index every time the process ends (even if timing out):
```
register_shutdown_function( array( $this, 'clear_index' ) );
```
We have to improve it though, as we want to tell the admin user that the last sync process actually failed.

felipeelia:
### Description of the Change

The code in this PR handles fatal errors during sync processes, so it does not get stuck but rather gets properly cleaned and logged. It also adds a new WP-CLI command: `get-last-sync` with the information of the latest sync performed, no matter the method (dashboard or cli.)

Closes #2737

### Verification Process

In addition to the regular checks, running `php -d max_execution_time=1 "$(which wp)" elasticpress index --per-page=2` makes the sync timeout, generating the main error the PR tries to fix.

### Changelog Entry

Added: New `get-last-sync` WP-CLI command.
Changed: Log errors and remove indexing status on failed syncs.

### Credits

Props @felipeelia

### DIFF
--- a/includes/classes/Command.php
+++ b/includes/classes/Command.php
@@ -1036,11 +1036,10 @@ class Command extends WP_CLI_Command {
 	 * @param array $assoc_args Associative CLI args.
 	 */
 	public function get_last_cli_index( $args, $assoc_args ) {
-
-		$last_sync = get_site_option( 'ep_last_cli_index', array() );
+		$last_sync = Utils\get_option( 'ep_last_cli_index', array() );
 
 		if ( isset( $assoc_args['clear'] ) ) {
-			delete_site_option( 'ep_last_cli_index' );
+			Utils\delete_option( 'ep_last_cli_index' );
 		}
 
 		$this->pretty_json_encode( $last_sync, ! empty( $assoc_args['pretty'] ) );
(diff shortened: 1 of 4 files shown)


tott (approved):
This looks good to me

Issue #2681, opened 2022-03-25
BUG: Synchronization Issues


cbratschi:
**Describe the bug**

Synchronization using the new sync UI did not progress after a while without showing any errors. Using WP-CLI ended in out of memory conditions while indexing posts and failed afterwards in users too. So we split the command to use separate cycles for each indexable. However, this ended in users and terms to contain only 1000 documents.

I saw similar open issues but our issue is probably a little bit different. Therefore creating a new bug report to discuss further details.

**Steps to Reproduce**

1. Synchronize
2. Wait only progress stops
3. Displays forever the same sync message

```
Processed users 15400 - 15750 of 277623. Last Object ID: 797854
Processed users 15400 - 15750 of 277623. Last Object ID: 797854
Processed users 15400 - 15750 of 277623. Last Object ID: 797854
Processed users 15400 - 15750 of 277623. Last Object ID: 797854
```

**Expected behavior**

Should synchronize all indexables.

**Screenshots**

**Environment information**
 - Device: 
 - OS: 
 - Browser and version: 
 - WordPress version: 
 
 - ElasticPress version:  4.0.1
 - Elasticsearch version:  7.9.3
 - Where do you host your Elasticsearch server? 
 - What ElasticPress features do you have currently active? 
 - Plugins and version: 
 - Theme and version: 
 
 - <details><summary>Site Health Info:</summary>
   
   </details>

**Additional context**

cbratschi:
More details about the WP-CLI issue.

Full index runs into out of memory conditions after using more than 1.6 GB of memory.

```
~/wp-cli/wp-cli.phar elasticpress index --setup --per-page=1000 --show-bulk-errors
Processed posts 61000 - 62000 of 205929. Last Object ID: 550888
Killed
```

We switched to separate calls. Where posts are resumed:

```
~/wp-cli/wp-cli.phar option delete ep_index_meta
~/wp-cli/wp-cli.phar elasticpress index --setup --per-page=1000 --show-bulk-errors --indexables=post
Processed posts 50000 - 51000 of 205929. Last Object ID: 597796
Killed

~/wp-cli/wp-cli.phar option delete ep_index_meta
~/wp-cli/wp-cli.phar elasticpress index --per-page=1000 --show-bulk-errors --indexables=post --offset=51000
Processed posts 120000 - 121000 of 205929. Last Object ID: 316790
Killed

~/wp-cli/wp-cli.phar option delete ep_index_meta
[...]

PR #2762, opened 2022-05-13
Implement proper timers for the index/sync command

felipeelia:
### Description of the Change

As stated [in this comment](https://github.com/10up/ElasticPress/issues/2681#issuecomment-1125749013), the times displayed by the WP-CLI sync/index command are currently wrong. That is happening because the code relies on WP's timer_start() but that does not work with nested calls: as Indexable is also calling that function, the timer resets for every batch of documents.

The PR introduces a timer_* collection of functions, inspired a bit by what WP core has in its wpdb class. 

### Changelog Entry

Fixed: "Time elapsed" and "Total time elapsed" in WP-CLI `index` command.

### Credits

Props @felipeelia @archon810

### DIFF
--- a/includes/classes/Command.php
+++ b/includes/classes/Command.php
@@ -56,6 +56,15 @@ class Command extends WP_CLI_Command {
 	protected $assoc_args = [];
 
 	/**
+	 * Internal timer.
+	 *
+	 * @since 4.2.0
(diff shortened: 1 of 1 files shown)


tott (approved):
This looks good to me

PR #2771, opened 2022-05-17
Fix --network-wide flag in WP-CLI index command

### LINKED ISSUE #2766: BUG: Using --url when running index on a multisite not working. Getting full network index always.
I looked everywhere and didn't see any mention about the `index` command no longer supporting the global WP-CLI command `--url` so reaching out here to see what's up? This may not be a bug but it was unexpected for me.

I'm working on upgrading to v4 of ElasticPress on a couple of large multisite networks. Historically I could control the indexing to a single site when needed, but after upgrading to versions 4.0.0+ the CLI command wants to index the full network no matter what.

While normally I want a full network index, I have found in one particular network there is a large sum of content to index across 60+ sites. When running the full index it eventually errors out once it reaches the end of the index with `Error: Network alias creation failed for posts ...`, and I was trying to debug further by syncing a single site but found the `index` command doesn't seem to honor the `--url` parameter anymore after 4.0.0. This still works fine in 3.6.6 though?

**Steps to Reproduce**

1. In your command prompt, run the index CLI command for a single site in your subdirectory multisite install `wp elasticpress index --setup --yes --url=https://test-multisite.test/some-sub-site`
2. Observe that the indexing logs report starting from site ID 1 and continue through all the sites in the network.

**Expected behavior**
I expect the index command to ONLY focus on indexing the site provided in the `--url` parameter offered by WP-CLI which was supported in ElasticPress 3.6.6.

**Screenshots**
**NOTE: The `doc` in my command is an alias for 10updocker**
Screenshot from 2022-05-16 14-44-11

**Environment information**
 - Device: All devices tested on (locally and on hosting)
 - OS: 
 - Browser and version: 
 - WordPress version: 5.9.3
 
 - ElasticPress version: 4.0.0+
 - Elasticsearch version: Unsure on version
 - Where do you host your Elasticsearch server? elasticpress.io
 - Plugins and version: 
 - Theme and version: 
 
   </details>

**Additional context**

felipeelia:
### Description of the Change

As stated in #2766, ElasticPress 4.x introduced a change in the `index` WP-CLI command that it would always index all sites, even when the `--network-wide` flag is not passed. This PR changes that back to the 3.x behavior:

In order to index all sites, the user needs to run `wp elasticpres index ... --network-wide`, otherwise only the main site will be indexed.

In addition to that, this PR also fixes all WP-CLI commands that rely on get_index_names() method, using indices names for all sites only when it is network activated. Commands affected are:
- get-mapping
- get-indexes
- status
- stats

Closes #2766

### Changelog Entry

- Fixed: WP-CLI `index` command without the `--network-wide` only syncs the main site.
- Fixed: WP-CLI commands get-mapping, get-indexes, status, and stats only uses all sites' indices name when network activated.

### Credits

Props @felipeelia @colegeissinger

### DIFF
--- a/bin/setup-cypress-env.sh
+++ b/bin/setup-cypress-env.sh
@@ -61,7 +61,13 @@ fi
 
 npm run env run tests-cli "wp core multisite-convert"
 
-# Not sure why, wp-env makes it http://localhost:8889/:8889
+SITES_COUNT=$(npm --silent run env run tests-cli "wp site list --format=count")
+if [ $SITES_COUNT -eq 1 ]; then
+	npm run env run tests-cli "wp site create --slug=second-site --title='Second Site'"
+	npm --silent run env run tests-cli "wp search-replace localhost/ localhost:8889/ --all-tables"
+fi
+
+# Not sure why, wp-env makes it http://localhost:8889/:8889 (not related to the command above)
 npm run env run tests-cli "option set home 'http://localhost:8889'"
 npm run env run tests-cli "option set siteurl 'http://localhost:8889'"
 
--- a/includes/classes/Command.php
+++ b/includes/classes/Command.php
@@ -441,7 +441,7 @@ class Command extends WP_CLI_Command {
 	 * @return array
 	 */
 	protected function get_index_names() {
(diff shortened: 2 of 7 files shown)


tott (approved):
This looks good to me

PR #2774, opened 2022-05-18
Remove Instant Results' dependency on @wordpress/date and @wordpress/components

### LINKED ISSUE #2768: Reduce size of Instant Results dependencies
As raised on #2756 Instant Results uses a couple of `@wordpress` packages that are quite large and could be substituted by smaller alternatives. 

In particular `@wordpress/date` is used for displaying result dates, but is a large package that depends on `Moment.js`, another large package. Also, SlotFills from `@wordpress/components` is used for active Facet chips, bringing along the whole library, but could be replaced with a manual custom Portals implementation.

Replace the large package dependencies with custom or smaller alternatives.

JakePT:
### Description of the Change

Replaces the implementation of SlotFill with a basic React portal for the active facets chips. This removes the dependency on `@wordpress/components` which is very large.

Also replaces the use of `@wordpress/date`, which is also quite large, with the use of `Date.toLocaleDateString()`. This reduces the options for date formatting, but we weren't supporting that yet anyway.

Closes #2768.

### Verification Process

In Instant Results verify that the active facets can be seen and clicked to remove when filtering results. Also verify that the wp-components dependency is not loaded.

Also verify that the date on Post results is displayed correctly, and updates according to the site's language.

### Checklist:

### Changelog Entry

Changed - Removed Instant Results' dependency on @wordpress/components and @wordpress/date.

### Credits

Props @JakePT


felipeelia:
@JakePT I tried this one but I'm seeing two errors:
1. As is, I'm getting a `Cannot destructure property 'gmdateI18n' of 'wp.date' as it is undefined.` error coming from [this line](https://github.com/10up/ElasticPress/blob/develop/assets/js/instant-results/components/results/result.js#L13). Making that `import { gmdateI18n } from '@wordpress/date';` seems to fix it.
2. If I search for "post", for example, and click on the search button, it opens the modal. If I close it and click on the search button again I get an error in the console that seems to be related to [this line](https://github.com/10up/ElasticPress/blob/develop/assets/js/instant-results/index.js#L152).

JakePT:
@felipeelia Thanks. I've pushed fixes for both these issues. Let me know if it works for you. I've also pushed a change to get rid of the @wordpress/date dependency, it just means that the date format will be determined by JavaScript's `Date.toLocaleString()` method, but at least the site's language is used.

PR #2779, opened 2022-05-20
Set some default args and instance values

felipeelia:
### Description of the Change

When rendering a block, the new Facet Renderer class relies on some undefined array indexes. This PR sets some defaults so that does not happen anymore.

### Changelog Entry

Fix: Undefined index notice in Facets renderer

### Credits


tott (approved):
Looks good to me

PR #2780, opened 2022-05-20
Hide password protected posts only on searches

felipeelia:
### Description of the Change

To best mirror WordPress core behavior, we should hide password-protected posts only on searches.

### Changelog Entry

Changed: (Protected Content) Password-protected posts are only hidden on searches.
### Credits

Props @felipeelia @JakePT @burhandodhy


tott (approved):
This looks good to me

PR #2781, opened 2022-05-20
Refactor User parse_orderby and add user_registered support

felipeelia:
### Description of the Change

This PR refactors User::parse_orderby(), adds support for user_registered into that and also handles lowercase `id` as a sort clause.

### Changelog Entry

Added: Support for ordering Users by `user_registered` and lowercase `id`.

### Credits

### DIFF
--- a/includes/classes/Indexable/User/User.php
+++ b/includes/classes/Indexable/User/User.php
@@ -625,6 +625,8 @@ class User extends Indexable {
 			return $sort;
 		}
 
+		$unsupported_clauses = [ 'rand', 'include', 'login__in', 'nicename__in', 'post_count' ];
+
 		foreach ( $orderby as $key => $value ) {
 			if ( is_string( $key ) ) {
 				$orderby_clause = $key;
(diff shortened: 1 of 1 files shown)


includes/classes/Indexable/User/User.php
```diff
@@ -631,73 +634,72 @@ public function parse_orderby( $orderby, $default_order, $query_vars ) {
...
-						),
-					);
-				} elseif ( 'user_url' === $orderby_clause || 'url' === $orderby_clause ) {
-					$sort[] = array(
-						'user_url.raw' => array(
-							'order' => $order,
-						),
-					);
-				} elseif ( 'meta_value' === $orderby_clause ) {
+			if ( empty( $orderby_clause ) && 'rand' === $orderby_clause ) {
```
tott: how would that second part ever be true

tott (changes requested):
This seems wrong

includes/classes/Indexable/User/User.php
felipeelia: Totally true, @tott. I've changed that and also added a couple of other fields we don't offer support to.

tott (approved):
This looks good to me

PR #2782, opened 2022-05-20
Only try to get ES info if a server is set

felipeelia:
### Description of the Change

Currently, during the initial setup, EP tries to get the ES server info even when that is not set yet. This PR addresses that issue.

### Changelog Entry

Fixed: Prevent an unnecessary call when the ES server is not set yet.

### Credits


tott (approved):
This looks good to me

PR #2789, opened 2022-05-23
Handle errors properly when WP_DISABLE_FATAL_ERROR_HANDLER is false

felipeelia:
### Description of the Change

In #2748 we've introduced an error handling. That only works when `define( 'WP_DISABLE_FATAL_ERROR_HANDLER', true );` but that is not the default. This PR expands the error handling to also work with WP default behavior.

### Changelog Entry

Added: New `ep_sync_number_of_errors_stored` filter.

### Credits

@felipeelia @tott @JakePT


tott (approved):
This looks good to me

PR #2799, opened 2022-05-25
Update the release process

### LINKED ISSUE #2727: BUG: Composer package is no longer functional
**Describe the bug**

ElasticPress is distributed via Packagist at https://packagist.org/packages/10up/elasticpress. Since 4.0, this package is no longer installable via Composer without also installing its developer dependencies and performing a build step for the same reasons outlined in the 4.0 release notes -- the built assets are no longer present in the source repo.

The Packagist integration with GitHub does not allow a custom release asset to be used, it only uses the zip for the tag (`/refs/tags/x.y.z.zip`), so the `elasticpress-x.y.z.zip` asset is useless to Packagist.

**Potential Solution**

I had this exact problem with my Query Monitor plugin when I wanted to remove the built assets from the source repo but retain the Composer package on Packagist. I solved it by using a GitHub action to run the build script, commit the additions, and tag the release with the built files included.

The workflow is here: https://github.com/johnbillion/query-monitor/blob/develop/.github/workflows/build.yml. The subsequent release to w.org still happens with 10up/action-wordpress-plugin-deploy: https://github.com/johnbillion/query-monitor/blob/develop/.github/workflows/deploy-tag.yml. Steps to deploy can be found here: https://github.com/johnbillion/query-monitor/blob/develop/CONTRIBUTING.md#releasing-a-new-version.

The main difference between this approach and the existing https://github.com/10up/ElasticPress/blob/develop/.github/workflows/push-deploy.yml is that with QM I push to a release branch which then gets built as the release, rather than the build happening on the `release` action.

It would be great if the package on Packagist was fixed so ElasticPress could be installed via Composer again.

felipeelia:
### Description of the Change

This PR changes the release process so we can distribute the plugin via composer without needing WPackagist.

Closes #2727

### Verification Process

Tested in https://github.com/felipeelia/ElasticPress/releases/tag/4.1.0 with success.

### Changelog Entry

Changed: The plugin is now available via Composer without any additional steps required.

### Credits

Props @felipeelia @jeffpaul @johnbillion


jeffpaul:
Noting that the unit test error is likely unrelated to the changes here, but otherwise still worth resolving separately to ensure those stay working.

.gitattributes
```diff
@@ -0,0 +1,42 @@
...
+/tests export-ignore
+/vendor/ export-ignore
+/node_modules export-ignore
+
+/.babelrc export-ignore
+/.browserslistrc export-ignore
+/.editorconfig export-ignore
+/.eslintignore export-ignore
+/.eslintrc.js export-ignore
+/.distignore export-ignore
```
jeffpaul: can probably remove this line since you're getting rid of that file

.gitattributes
```diff
@@ -0,0 +1,42 @@
...
+/.gitattributes export-ignore
+/.gitignore export-ignore
+/.travis.yml export-ignore
```
jeffpaul: not sure this is needed either?

.gitattributes
```diff
@@ -0,0 +1,42 @@
...
+/hookdoc-conf.json export-ignore
+/LICENSE.md export-ignore
+/package.json export-ignore
+/package-lock.json export-ignore
+/phpcs.xml export-ignore
+/phpunit.xml.dist export-ignore
+/postcss.config.js export-ignore
+/README.md export-ignore
+/SECURITY.md export-ignore
+/run-wpacceptance.sh export-ignore
```
jeffpaul: can probably remove this as well

.github/workflows/build-and-tag.yml
```diff
@@ -0,0 +1,39 @@
...
+jobs:
+  wordpress:
+    name: Release
+    runs-on: ubuntu-latest
+    steps:
+    - name: Checkout code
+      uses: actions/checkout@v2
+
+    - name: install node v16
+      uses: actions/setup-node@v2
```
jeffpaul: this is now up to v3, any reason not to go with that version?

.github/workflows/push-deploy.yml
```diff
@@ -15,16 +15,6 @@ jobs:
     - name: Checkout code
       uses: actions/checkout@v2
```
jeffpaul: this is also at v3 now, could be worth testing if bumping there works

.github/workflows/push-deploy.yml
```diff
@@ -15,16 +15,6 @@ jobs:
...
-      with:
-        node-version: 16
-
-    - name: Build
-      run: |
-        npm ci
-        npm run build
-
     - name: WordPress Plugin Deploy
       if: "! github.event.release.prerelease"
```
jeffpaul, line 19: Since we're not including `prerelease` in line 8 above is this check necessary?

CONTRIBUTING.md
```diff
@@ -28,20 +28,20 @@ The `develop` branch is the development branch which means it contains the next
...
-1. Check WordPress.org: Ensure that the changes are live on https://wordpress.org/plugins/elasticpress/.  This may take a few minutes.
+1. [Check the _Build and Tag_ action](https://github.com/10up/ElasticPress/actions/workflows/build-and-tag.yml): a new tag named with the version number should've been created. It should contain all the built assets.
+1. Release: Create a [new release](https://github.com/10up/elasticpress/releases/new), naming the release with the new version number, and targeting the tag created in the previous step. Paste the release changelog from `CHANGELOG.md` into the body of the release and include a link to the closed issues on the [milestone](https://github.com/10up/elasticpress/milestone/#?closed=1).
+1. SVN: Wait for the [GitHub Action](https://github.com/10up/ElasticPress/actions?query=workflow%3A%22Deploy+to+WordPress.org%22) to finish deploying to the WordPress.org repository. If all goes well, users with SVN commit access for that plugin will receive an emailed diff of changes.
```
jeffpaul: Is that GHA link correct?

.github/workflows/build-and-tag.yml
```diff
@@ -0,0 +1,39 @@
...
+
+    - name: Build
+      run: |
+        npm ci
+        npm run build
+
+    - name: Setup
+      run: 'echo "VERSION=$(grep -Po ''\"version\": \"[0-9\\.]+\"'' package.json | grep -Po ''[0-9\\.]+'')" >> $GITHUB_ENV'
+
+    - name: Tag
```
jeffpaul, line 28: Are we certain we want to automate tagging on a push to `trunk`?  Will want to admit this opens us to some human error where an incomplete or inadvertent push to `trunk` will result in a tag that may not have been intended (or if the version number wasn't bumped correctly the tagging will presumably fail since the tag number already exists).

CONTRIBUTING.md
```diff
@@ -28,20 +28,20 @@ The `develop` branch is the development branch which means it contains the next
...
-1. SVN: Wait for the [GitHub Action](https://github.com/10up/ElasticPress/actions?query=workflow%3A%22Deploy+to+WordPress.org%22) to finish deploying to the WordPress.org repository.  If all goes well, users with SVN commit access for that plugin will receive an emailed diff of changes.
-1. Check WordPress.org: Ensure that the changes are live on https://wordpress.org/plugins/elasticpress/.  This may take a few minutes.
+1. [Check the _Build and Tag_ action](https://github.com/10up/ElasticPress/actions/workflows/build-and-tag.yml): a new tag named with the version number should've been created. It should contain all the built assets.
+1. Release: Create a [new release](https://github.com/10up/elasticpress/releases/new), naming the release with the new version number, and targeting the tag created in the previous step. Paste the release changelog from `CHANGELOG.md` into the body of the release and include a link to the closed issues on the [milestone](https://github.com/10up/elasticpress/milestone/#?closed=1).
```
jeffpaul, line 42: I've not tried things this way, so curious if selecting that tag will already set the branch to `trunk` or if that's still needed to be called out in this step?

jeffpaul (commented):
Ok, I've added a bunch of questions/comments, but overall seems ok to :yolo: as no real blockers here

.github/workflows/build-and-tag.yml
felipeelia: No reason at all, pushed a commit updating it everywhere.

.github/workflows/push-deploy.yml
felipeelia: No reason at all, pushed a commit updating it everywhere.

.github/workflows/push-deploy.yml
felipeelia: Actually, line 8 is for draft/published, and this line is for pre-releases (alpha, beta, etc.) or "official" releases. We probably should move this check to be appended to line 8 but that would require some additional research.

CONTRIBUTING.md
felipeelia: Which one, @jeffpaul? 41 will be once we merge and run this once. The second one (line 43) was already there (and also seems to work)

jeffpaul: Yes 43 resolves but that's the action from 3 years ago, seems like you'd want to link to the new one as https://github.com/10up/ElasticPress/actions/workflows/push-deploy.yml?query=workflow%3A%22Deploy+to+WordPress.org%22 yeah?

felipeelia: :doh: of course. Changed it now.

.github/workflows/build-and-tag.yml
felipeelia: Actually, the more I think about this, the more it makes sense to me. If someone inadvertently pushes to `trunk` and:
1. The version number is not updated: it will likely fail and do nothing
2. If the version is updated, then we've already merged the changelog, etc. If we are at that point and inadvertently pushes to trunk, we'd expect something to fail anyway.

We could mirror trunk to a `stable` branch and tag from that but I don't know if we would win much with an additional step. Happy to discuss it.

CONTRIBUTING.md
felipeelia: Yeah, that was something interesting. When you select a tag that already exists, GitHub won't ask you anything about the branch.

PR #2800, opened 2022-05-25
Transform the `id` field into `ID`

felipeelia:
This addresses a change made in WP Core on 6.0

https://github.com/WordPress/WordPress/commit/f6ef37b050db9463d7cc0f8f841b85994062c24c#diff-26773a0cc66b5ba7b3c5108652bb2367424e6a78d4a05d1e35ff6d45701e3c07R300

With the 4.2.0 release planned for tomorrow, this is the minimal change to have it working again.

### Changelog Entry

Fixed: An incompatibility with the way WP 6.0 handles WP_User_Query using fields.

### Credits

Props @felipeelia


tott (approved):
Looks weird but makes sense :)

PR #2803, opened 2022-05-26
Change GH Action used to publish to GH Pages

felipeelia:
### Description of the Change

During ElasticPress 4.2.0 release, we got an error with the GitHub Pages update. It is outlined in this [issue](https://github.com/maxheld83/ghpages/issues/35).

This PR applies the same fix used by Automattic/sensei in this PR: https://github.com/Automattic/sensei/pull/5052

As a bonus, the action we are switching to seems to have much more traction and frequent updates.

### Verification Process

We will need to merge this one to trunk to see if it works.

### Changelog Entry

Changed: GH Action used to update the documentation.

### Credits

Props @felipeelia

### DIFF
--- a/.github/workflows/build-docs.yml
+++ b/.github/workflows/build-docs.yml
@@ -64,7 +64,7 @@ jobs:
         CI: true
 
     - name: Deploy to GH Pages
-      uses: maxheld83/ghpages@v0.3.0
-      env:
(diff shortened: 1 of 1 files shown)


jeffpaul (approved):
yep, same change we've made in other places but hadn't gotten around to ALL the GH Page repos yet, nice catch!

PR #2814, opened 2022-06-01
Change sync button title attribute 

### LINKED ISSUE #2793: Change sync button title attribute
The Sync button in the ElasticPress toolbar at the top of ElasticPress admin pages has a title attribute/tooltip that reads "Start Sync", however clicking the button only takes the user to the Sync page, and no longer starts a sync immediately.

The title attribute should be updated to not suggest that a sync will start immediately. Just "Sync" or "Go to Sync page" would probably be more appropriate.

burhandodhy:
### Description of the Change
This PR changes the Sync button title attribute to `Sync Page` and Settings button title to `Settings Page`  and also add the E2E test.

Closes #2793 

### Verification Process
- Go to Settings Page.
- Hover on the Sync button. The title would be `Sync`

### Changelog Entry

Changed: Sync button Title.
### Credits

Props @burhandodhy @JakePT

### DIFF
--- a/tests/cypress/integration/admin/settings-page.spec.js
+++ b/tests/cypress/integration/admin/settings-page.spec.js
@@ -1,0 +1,7 @@
+describe('Settings Page', () => {
+	it('Can see a Sync and Settings buttons on Settings Page', () => {
+		cy.visitAdminPage('admin.php?page=elasticpress-settings');
+		cy.get('.dashicons.start-sync').should('have.attr', 'title', 'Sync Page');
+		cy.get('.dashicons.dashicons-admin-generic').should('have.attr', 'title', 'Settings Page');
+	});
+});
--- a/tests/cypress/integration/admin/settings-page.spec.js
+++ b/tests/cypress/integration/admin/settings-page.spec.js
@@ -1,7 +1,0 @@
-describe('Settings Page', () => {
-	it('Can see a Sync and Settings buttons on Settings Page', () => {
-		cy.visitAdminPage('admin.php?page=elasticpress-settings');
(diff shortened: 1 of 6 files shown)


felipeelia:
@burhandodhy instead of simply "Sync", let's say "Sync Page" (and "Settings Page".)

felipeelia:
@burhandodhy instead of simply "Sync", let's say "Sync Page" (and "Settings Page".)

tests/cypress/integration/admin/settings-page.spec.js
```diff
@@ -0,0 +1,7 @@
+describe('Settings Page', () => {
+	it('Can see a Sync and Settings buttons on Settings Page', () => {
+		cy.visitAdminPage('admin.php?page=elasticpress-settings');
+		cy.get('.dashicons.start-sync').should('have.attr', 'title', 'Sync Page');
+		cy.get('.dashicons.dashicons-admin-generic').should('have.attr', 'title', 'Settings Page');
+	});
+});
```
felipeelia: It is GREAT to see tests like these coming through, @burhandodhy, congrats, and thanks! Can we please move it from a whole new test to the already existent `general.spec.js`? The links should be on all (or almost all) pages, so although it makes sense to pick one to check, we can save a `settings-page.spec.js` to something particular to that page, like checking if tabs change, etc. What do you think?

felipeelia (changes requested):

tests/cypress/integration/admin/settings-page.spec.js
burhandodhy: @felipeelia agree with you. `settings-page.spec.js` should only have tests that are related to settings page. I move the test to general.spec.js

PR #2817, opened 2022-06-02
Fix: Admin searches in WooCommerce taxonomies return no results

### LINKED ISSUE #2791: BUG: Admin searches in WooCommerce taxonomies return no results
**Describe the bug**
When searching within WooCommerce Product Categories or product attributes in the admin with the Terms feature enabled, no results are returned.

With Debug Bar enabled two Elasticsearch queries will be shown. One of them will have no results, but the other one will have the expected results. The queries are exactly the same excerpt two differences. Firstly, the query with no results is sorted with

```
"sort": [
	{
		"meta.order.long": {
			"order": "asc"
		}
	}
],
```

And has this additional `post_filter`:

```
{
	"bool": {
		"must": [
			{
				"terms": {
					"meta.order.raw": [
						""
					]
				}
			}
		]
	}
}
```

So the cause has to be something to do with the custom sorting of terms used by WooCommerce.

**Steps to Reproduce**

1. Install WooCommerce.
2. Import WooCommerce sample data.
3. Go to _Products > Categories_ and search for "Decor". No results are returned.
4. Go to _Products > Attributes > color > Configure terms_ and search for "Gray". No results are returned.
5. Disabled the Terms feature and repeat the searches. Results should be returned as expected.

**Expected behavior**
Searching terms should return matching terms.

burhandodhy:
### Description of the Change
This PR fixes the issue where the EP returns incorrect result when the `meta_value` key is not set in `WP_Query` and `WP_Term_Query`.

 When the  `meta_value` is not set, WordPress runs the exact SQL query that it runs in case of the `meta_compare => 'Exists'` and when the `meta_value` is empty, WordPress only returns the posts where the meta value is empty. This statement is only true for the `WP Query` and not for the `Term Query`

For the `Term Query`, WordPress runs the same query either the `meta_value` is set to empty or its not set. The query in both cases is exact what we have in the case of `Exists`. To tackle this case, I added another check that check if the `meta_value` is not empty indexable Term class

Closes #2791 

### Possible Drawbacks

I didn't find any negative impact. I tried to test all the cases and its working fine. 

### Verification Process

1) Import WooCommerce Dummy data.
2) Run the below code first and then uncomment `ep_integrate` line and run it again. In both cases result should be same

```
	$args = array(
		'taxonomy'   => 'product_cat',
		// 'ep_integrate' => 1,
		'hide_empty' => false,
		'meta_key' => 'order',
		'meta_value' => 0
	);
	var_dump( get_terms( $args ) );
```
3) To run the test against `WP_Query`. Run the below code twice. First without EP and second time with EP integration. In both case the result should be same. 

```
	$args = array(
		'post_type'  => 'post',
		// 'ep_integrate' => 1,
		'posts_per_page' => -1,
		'meta_query' => array(
			array(
	 			'key' => 'invalid',
			),
		),
	);
	$query = new WP_Query( $args );

	var_dump( $query->post_count );

```
4) If you test the both snippet with develop branch, you will the result EP result will not be same what WordPress returns. 
5) To test against admin search in WooCommerce taxonomies. 
   - Go to WordPress admin dashboard -> Products -> Categories
   - Search `Clothing`
   - It should return only returns Clothing category.

### Checklist:

### Changelog Entry

Fixed - Meta queries with 'exists' as compare operator and empty meta values handling

### Credits

Props @burhandodhy

### DIFF
--- a/includes/classes/Indexable/Post/Post.php
+++ b/includes/classes/Indexable/Post/Post.php
@@ -1280,23 +1280,24 @@ class Post extends Indexable {
 		$meta_queries = [];
 
 		/**
-		 * Support meta_key
-		 *
-		 * @since  2.1
+		 * Support `meta_key`, `meta_value`, `meta_value_num`, and `meta_compare` query args
 		 */
 		if ( ! empty( $args['meta_key'] ) ) {
-			if ( ! empty( $args['meta_value'] ) ) {
-				$meta_value = $args['meta_value'];
-			} elseif ( ! empty( $args['meta_value_num'] ) ) {
-				$meta_value = $args['meta_value_num'];
+			$meta_query_array = [
+				'key' => $args['meta_key'],
+			];
+
+			if ( isset( $args['meta_value'] ) && '' !== $args['meta_value'] ) {
+				$meta_query_array['value'] = $args['meta_value'];
+			} elseif ( isset( $args['meta_value_num'] ) && '' !== $args['meta_value_num'] ) {
+				$meta_query_array['value'] = $args['meta_value_num'];
 			}
 
-			if ( ! empty( $meta_value ) ) {
-				$meta_queries[] = array(
-					'key'   => $args['meta_key'],
-					'value' => $meta_value,
-				);
+			if ( isset( $args['meta_compare'] ) ) {
+				$meta_query_array['compare'] = $args['meta_compare'];
 			}
+
+			$meta_queries[] = $meta_query_array;
 		}
 
 		/**
--- a/includes/classes/Indexable/Term/Term.php
+++ b/includes/classes/Indexable/Term/Term.php
@@ -478,7 +478,7 @@ class Term extends Indexable {
 				'key' => $query_vars['meta_key'],
 			];
 
-			if ( isset( $query_vars['meta_value'] ) ) {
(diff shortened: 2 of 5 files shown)


includes/classes/Indexable/Post/Post.php
```diff
@@ -1280,23 +1280,24 @@ public function format_args( $args, $wp_query ) {
 		$meta_queries = [];
 
 		/**
-		 * Support meta_key
-		 *
-		 * @since  2.1
+		 * Support `meta_key`, `meta_value`, and `meta_compare` query args
 		 */
 		if ( ! empty( $args['meta_key'] ) ) {
```
burhandodhy, line 1285: This doesn't add `meta_key` if `meta_value` is not set.

includes/classes/Indexable/Term/Term.php
```diff
@@ -478,7 +478,7 @@ public function format_args( $query_vars ) {
 				'key' => $query_vars['meta_key'],
 			];
 
-			if ( isset( $query_vars['meta_value'] ) ) {
+			if ( isset( $query_vars['meta_value'] ) && '' !== $query_vars['meta_value'] ) {
```
burhandodhy, line 481: We only want to pass `meta_value` to ES if its value is not empty

PR #2819, opened 2022-06-02
Support transforming legacy widgets to blocks and hide widgets when the blocks are available

### LINKED ISSUE #2792: Hide legacy widgets when not already in use
When using the block-based widgets editor both legacy widgets and blocks are available for Related Posts and Facets. This could be confusing as there are two options for the same thing.

Following the pattern of WordPress core's legacy widgets that became blocks, we should not register the legacy widgets on sites that are not already using the legacy widget somewhere _and_ are not using the [classic widgets](https://wordpress.org/plugins/classic-widgets/) plugin, [or similar](https://wordpress.org/plugins/restore-classic-widgets/). This would keep the widgets around for those who need them, while only making blocks available for all new sites.

JakePT:
### Description of the Change

This PR does two things:

1. Adds support for transforming the legacy Facet and Related Posts widgets into blocks. For these widgets when the block icon is clicked the _Transform to_ menu will include the new block versions of these widgets. The settings of the legacy widget will be carried over to the block.
2. Hides the Facet and Related Posts widgets from the Legacy Widget block. This means that when the block editor is being used there won't be two Facet or Related Posts blocks, which could cause confusion. This won't affect any existing instances of the legacy widgets, which will still be visible and editable, but it will prevent new instances being added. The legacy widgets will also still be usable if the site is using the Classic Widgets plugin, or similar, to disable the block editor for widgets.

Both changes are implemented as per the guidance in the [Block Editor Handbook](https://developer.wordpress.org/block-editor/how-to-guides/widgets/legacy-widget-block/#allowing-migration-to-a-block) and the behaviour is consistent with how most core legacy widgets are handled, such as Search.

Closes #2792

### Possible Drawbacks

While existing block instances won't be affected, the user will not be able to add new instances of the legacy widget. 

However, if the user doesn't want to switch existing instances to the block and needs to add new legacy widget instances, then there are two workarounds:

1. Use the core Widget Group block, which will wrap the block in the markup defined by the theme for the widget area.
2. Un-hide the legacy widgets by using the `widget_types_to_hide_from_legacy_widget_block` filter hook to remove their IDs from the list filtered by the hook.

### Verification Process

1. Before applying this PR, add several Facet and Related Posts legacy widget blocks with a variety of settings.
2. After applying the PR, go to the Widgets editor and the legacy widget blocks should still be visible.
3. Attempt to add a new block, when searching for "Facet" or "Related Posts" only one option from ElasticPress should appear for each and it should be the block, not the legacy widget.
4.  Install and activate [Classic Widgets](https://wordpress.org/plugins/classic-widgets/). With this activated the legacy widgets should be available in the widgets editor. 
5. Deactivate the plugin.
6. On the widgets screen, select each legacy widget and click the block's icon in the block toolbar. There should be an option to transform it into the block version of the widget. Clicking to transform the widget should convert it to the block and the settings of the widget should be reflected on the new block. If the widget had a Title this should be added as a heading block before the new block.

### Checklist:

### Changelog Entry

Added - Support for transforming instances of the legacy Facet and Related Posts widgets into blocks.
Changed - The legacy Facet and Related Posts widgets are now hidden when using the block editor.

### Credits

Props @JakePT


JakePT:
I've finally added complete tests for the Related Posts block, but the Facets block still needs new tests as the current tests are based on using the Legacy Widget as a block.

tests/cypress/integration/features/facets.spec.js
```diff
@@ -1,118 +1,235 @@
...
+	beforeEach(() => {
+		cy.emptyWidgets();
+		cy.wpCli('plugin deactivate classic-widgets');
+	});
 
-		createWidget('Facet (categories)', 'category');
+	/**
+	 * Test that the Related Posts block is functional.
+	 */
+	it('Facet block', () => {
```
felipeelia: @JakePT do you mind rephrasing this to describe what is being tested?

tests/cypress/integration/features/facets.spec.js
```diff
@@ -1,118 +1,235 @@
...
+		cy.get('@secondBlock').contains('.term', 'template').click();
+		cy.url().should('not.include', 'ep_filter_post_tag=template');
+		cy.url().should('include', 'ep_filter_category=classic');
 	});
 
+	/**
+	 * Test that the Facet widget is functional and can be transformed into the
+	 * Facet block.
+	 */
+	it('Facet widget', () => {
```
felipeelia: @JakePT same here

tests/cypress/integration/features/facets.spec.js
```diff
@@ -1,118 +1,235 @@
...
+		 */
+		cy.visit('/');
+		cy.get('.widget_ep-facet').first().as('widget');
+		cy.get('@widget').find('input').should('have.attr', 'placeholder', 'Search Tags');
+		cy.get('@widget').find('.term').should('be.elementsSortedAlphabetically');
+
+		/**
+		 * Visit the block-based widgets screen.
+		 */
+		cy.wpCli('wp plugin deactivate classic-widgets');
```
felipeelia: @JakePT we have a command to activate and deactivate plugins. I'd prefer to stick with that just to be sure any performance improvement we have is applied across the entire test suite.

tests/cypress/integration/features/facets.spec.js
```diff
@@ -1,118 +1,235 @@
...
-					cy.get('body').should('contain.text', 'Widgets saved.');
-				});
-		});
-	}
-
 	before(() => {
 		cy.maybeEnableFeature('facets');
-
 		cy.wpCli('elasticpress index --setup --yes');
+		cy.wpCli('plugin install classic-widgets');
```
felipeelia: @JakePT can we install the plugin during the env setup? It would be a matter of adding it to the [.wp-env](https://github.com/10up/ElasticPress/blob/develop/.wp-env.json) file. Hopefully that gets cached in a good way (if not now, in the future) and it doesn't slow down the test execution.

tests/cypress/support/commands.js
```diff
@@ -290,6 +271,88 @@ Cypress.Commands.add('deactivatePlugin', (slug, method = 'dashboard', mode = 'si
...
+	});
+});
+
+Cypress.Commands.add('getBlocksList', () => {
+	cy.get('.block-editor-inserter__block-list');
+});
+
+Cypress.Commands.add('insertBlock', (blockName) => {
+	cy.get('.block-editor-block-types-list__item').contains(blockName).click();
+});
```
felipeelia: I wonder if we shouldn't start a new file only with blocks. Thoughts?

felipeelia (changes requested):
Overall it looks really good @JakePT. Left some comments about specific items.

I'd like to know your thoughts about testing the "transform" part of those blocks. Isn't that worth testing in Cypress?

tests/cypress/integration/features/facets.spec.js
JakePT: @felipeelia Done.

tests/cypress/integration/features/facets.spec.js
JakePT: @felipeelia Done.

tests/cypress/integration/features/facets.spec.js
JakePT: @felipeelia Done.

tests/cypress/integration/features/facets.spec.js
JakePT: @felipeelia Done.

tests/cypress/support/commands.js
JakePT: @felipeelia Done.

felipeelia (approved):

PR #2821, opened 2022-06-02
Delete search template from ElasticPress.io when deactivating Instant Results

### LINKED ISSUE #2794: Disabling Instant Results should deactivate search API
When using ElasticPress.io, activating Instant Results saves a 'search template' to EP.io which enables the publicly accessible search API that powers Instant Results for the user's posts index. However when Instant Results is deactivated the search template is not removed, the the API remains active.

Ideally deactivating Instant Results should delete the search template, thus deactivating the API for the user's posts index.

I believe the API already supports a DELETE request to remove the search template, so this should just be a matter of firing off that request when Instant Results is deactivated.

It's conceivable that a user may want to use the API while not using Instant Results, but if we want to support this then it can probably be sufficiently achieved by just enabling Instant Results and dequeueing the script.

JakePT:
### Description of the Change

Deletes the search template from the ElasticPress.io API when Instant Results is deactivated.

Also fixes the syntax of the `@hook` inline docs for Instant Results.

Closes #2794

### Verification Process

WIP

### Checklist:

### Changelog Entry

Added - The public search API is automatically deactivated when the Instant Results feature is deactivated.

### Credits

Props @JakePT


JakePT:
Note the endpoint is on the API is a WIP. Also, this might be something we want to do when the plugin itself is uninstalled.

includes/classes/Feature/InstantResults/InstantResults.php
```diff
@@ -347,13 +369,49 @@ public function epio_save_search_template() {
...
+		 * Filters the search template API endpoint.
 		 *
 		 * @since 4.0.0
+		 * @hook ep_instant_results_template_endpoint
+		 * @param {string} $endpoint Endpoint path.
+		 * @param {string} $index Elasticsearch index.
+		 * @returns {string} Search template API endpoint.
 		 */
-		do_action( 'ep_instant_results_template_saved', $search_template, $this->index );
+		$endpoint = apply_filters( 'ep_instant_results_template_endpoint', "api/v1/search/posts/{$this->index}/template/", $this->index );
```
JakePT, line 397: This could probably be extracted into a method, rather than being defined twice. Or it might be the time to put these methods into a class for working with the API.

Issue #2796, opened 2022-05-24
BUG: Behaviour of Related Posts widget and block are not consistent 


burhandodhy:
**Describe the bug**
The Related Posts widget doesn't show anything when the page is not is_single. However, the block does show the related posts. This is happening because we have a sanity check for the widget that checks if the page is not single then returns. https://github.com/10up/ElasticPress/blob/develop/includes/classes/Feature/RelatedPosts/Widget.php#L42

**Steps to Reproduce**

1. Add Related Posts Widget and Block in footer.
2. Go to front page. 
3. Widget doesn't render anything

JakePT:
As discussed we probably don't want these to be consistent, since the block has uses outside of `is_single()`, so instead we can use a check for `in_the_loop() || is_singular()` so that at least we know that there is a relevant post to get related posts for. The legacy widget will remain unchanged because we don't want it to suddenly appear in places users did not expect it to.

@felipeelia I think that's where we ended up, right?

chaselivingston:
@JakePT let me know this isn't quite done. #2825 is a fix for a related issue but not this exact one.

JakePT:
I went to make this change but ran into an issue. `in_the_loop()` does not return `true` inside the Query Loop block, so checking `in_the_loop()` would stop the block from appearing when used inside the query loop block. 

Alternatively I though I might be able to check the block for the `postId` context that's used inside the editor, since I could use that to determine if we were inside the query loop block, but that value does not seem to be available in PHP (it's in the class, but as a protected property).

Frankly I think there's an issue with Core here that might need to be resolved before we can reliably change the behaviour here.

JakePT:
There's an open issue regarding `in_the_loop()` in the Gutenberg repo: https://github.com/WordPress/gutenberg/issues/32896 I've added a comment to hopefully revive it.

felipeelia:
Closing this one as there is nothing we can do.

PR #2825, opened 2022-06-02
Support an accurate preview of the Related Posts block inside query loop block

JakePT:
### Description of the Change

The Related Posts block currently uses the post ID returned by `select('core/editor').getCurrentPostId()` to get the list of related posts for the preview. When adding the block to a post this works as expected and it displays the related posts for the post being edited, while adding the block to a widget area, where there is no ID available, it displays a generic list of posts as a preview.

Similarly, when adding the block to a FSE template there is no ID available, so the generic list is used. However, this is also the case when using the block inside the Query Loop block, where there is an actual post that could theoretically be used. This is because `select('core/editor').getCurrentPostId()` only returns the ID for the post being edited, and there is no post when editing an FSE template. This also means that if the query loop block is used on a single post then the list of related posts for the edited post will be used, rather than those for each individual post in the loop or the generic list.

This PR updates the Related Posts block to use the `postID` [block context](https://developer.wordpress.org/block-editor/reference-guides/block-api/block-context/), which allows it to use the post ID of each post in the query loop block. The correct post ID is also returned by this context when editing single posts, or the template for a single post, so it can replace the method for getting the relevant post ID entirely.

### Possible Drawbacks

I'm not actually sure which version of WordPress core block contexts were introduced in, so this may require a higher minimum version.

### Verification Process

1. With a block theme activated edit a template that uses the Query Loop block.
2. Inside the Query Loop block, add the Related Posts block to the post template. The preview should show the correct related posts for each post in the loop.
3. Edit an individual post and add the Related Posts block. The preview should show the correct related posts for that post.
4. While editing that post, add a Query Loop block and add the Related Posts block to the post template. The preview should show the correct related posts for each post in the loop. The Related Posts block outside the loop should continue to show the related Posts for the post being edited.

### Checklist:

### Changelog Entry

Fixed - An issue where the Related Posts block would display the wrong posts in the preview when added inside a Query Loop block.

### Credits

Props @JakePT

### DIFF
--- a/assets/js/blocks/related-posts/Edit.js
+++ b/assets/js/blocks/related-posts/Edit.js
@@ -33,7 +33,7 @@ class Edit extends Component {
 		};
 
 		// Use 0 if in the Widgets Screen
-		const postId = wp.data.select('core/editor').getCurrentPostId() ?? 0;
+		const { context: { postId = 0 } = {} } = this.props;
 
 		wp.apiFetch({
 			path: addQueryArgs(`/wp/v2/posts/${postId}/related`, urlArgs),
--- a/assets/js/blocks/related-posts/block.js
+++ b/assets/js/blocks/related-posts/block.js
@@ -25,6 +25,7 @@ registerBlockType('elasticpress/related-posts', {
 			default: 5,
 		},
 	},
+	usesContext: ['postId'],
 
 	/**
 	 * Handle edit


felipeelia:
@JakePT testing with WordPress 5.8 and a regular theme, when I go to the Customizer and try to manage blocks/widgets there, I see two different sets of posts:

image
(Facets and the widget do not render anything on the homepage.) Is this difference between the previews something we can address?

JakePT:
@felipeelia I've double checked and this particular difference isn't a regression, but you're right that it's not addressed by this PR. The problem is that I don't believe there's any way for the block to 'know' the context of the widget area in the preview. 

The block currently uses `get_the_ID()` when being rendered, and that will vary based on what page is being viewed and where the widget area is on that page. I think this gets at the issue raised in #2796 and whether we should show the block at all on archive pages. But even if we did only show it on single pages I don't know if the block can know which page is being previewed. I'll reach out on Slack to see if anyone has an idea.

PR #2828, opened 2022-06-06
Restrict to convert meta value into date if year is greater than 2099

### LINKED ISSUE #2769: BUG: values that are not dates can get formatted as if they were
**Describe the bug**

Looks like https://github.com/10up/ElasticPress/blob/18a0d3b0797d258c99e519cae87a84a4fa3b6faf/includes/classes/Indexable.php#L711 is treating some values that are not dates as if they were, later causing issues with ElasticSearch. If we try to prepare the following value: 5479516sunt it gets formatted as a valid date value 107039-03-10

Screen Shot 2022-05-17 at 9 00 13

**Steps to Reproduce**

1. Have a post with post meta key test and value 5479516sunt
2. Run an indexing operation
3. See instead of default date values 1970-01-01 is 107039-03-10

**Expected behavior**

**Screenshots**

**Environment information**
 - Device: 
 - OS: 
 - Browser and version: 
 - WordPress version: 
 
 - ElasticPress version: 
 - Elasticsearch version: 
 - Where do you host your Elasticsearch server? 
 - Plugins and version: 
 - Theme and version: 
 
   </details>

**Additional context**

burhandodhy:
### Description of the Change
`prepare_date_meta_values ` converts the meta into date even they are not. This PR adds the restriction that will check that if the meta year is greater than 2099 than don't convert into `datetime` format

Closes #2769 

### Changelog Entry

Fixed: Meta values that are not date are converted into date format

### Credits

Props @burhandodhy

### DIFF
--- a/includes/classes/Indexable.php
+++ b/includes/classes/Indexable.php
@@ -723,10 +723,14 @@ abstract class Indexable {
 		if ( $new_date ) {
 			$timestamp = $new_date->getTimestamp();
 
+			// date_create converts 5479516sunt meta into 107039-03-10. We keep using default values
+			// if year in date is greater than 2099.
+			$max_year = 2099;
+
 			// PHP allows DateTime to build dates with the non-existing year 0000, and this causes
 			// issues when integrating into stricter systems. This is by design:
 			// https://bugs.php.net/bug.php?id=60288
-			if ( false !== $timestamp && '0000' !== $new_date->format( 'Y' ) ) {
+			if ( false !== $timestamp && '0000' !== $new_date->format( 'Y' ) && $new_date->format( 'Y' ) <= $max_year ) {
 				$meta_types['date']     = $new_date->format( 'Y-m-d' );
 				$meta_types['datetime'] = $new_date->format( 'Y-m-d H:i:s' );
 				$meta_types['time']     = $new_date->format( 'H:i:s' );
--- a/includes/classes/Indexable.php
+++ b/includes/classes/Indexable.php
@@ -726,12 +726,15 @@ abstract class Indexable {
 			/**
 			 * Filter the maximum year limit for date conversion.
 			 *
+			 * Use default date if year is greater than max limit. EP has limitation that doesn't allow to have year greater than 2099.
+			 *
+			 * @see https://github.com/10up/ElasticPress/issues/2769
+			 *
 			 * @hook ep_max_year_limit
 			 * @param  {int} $year Maximum year limit.
 			 * @return {int} Maximum year limit.
+			 * @since  4.2.1
 			 */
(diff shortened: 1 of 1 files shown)


includes/classes/Indexable.php
```diff
@@ -723,10 +723,14 @@ public function prepare_date_meta_values( $meta_types, $meta_value ) {
 		if ( $new_date ) {
 			$timestamp = $new_date->getTimestamp();
 
+			// date_create converts 5479516sunt meta into 107039-03-10. We keep using default values
+			// if year in date is greater than 2099.
```
felipeelia: @burhandodhy can we rewrite this a bit just to make it clear that it comes from an Elasticsearch limitation? A `@see https://github.com/10up/ElasticPress/issues/2769` can also help developers to find why we have this max.

Beyond that, I wonder if it wouldn't be good to have a filter here, so people could change it if needed. Thoughts?

felipeelia (changes requested):

includes/classes/Indexable.php
burhandodhy: @felipeelia I have added the new filter `ep_max_year_limit` and also update the comment. Let me know if further changes required.

includes/classes/Indexable.php
```diff
@@ -723,10 +723,21 @@ public function prepare_date_meta_values( $meta_types, $meta_value ) {
...
+			/**
+			 * Filter the maximum year limit for date conversion.
+			 *
+			 * @hook ep_max_year_limit
+			 * @param  {int} $year Maximum year limit.
+			 * @return {int} Maximum year limit.
+			 */
+			// Use default date if year is greater than max limit. EP has limitation that doesn't allow to have year greater than 2099.
+			// @see https://github.com/10up/ElasticPress/issues/2769
+			$max_year = apply_filters( 'ep_max_year_limit', 2099 );
```
felipeelia: This is looking better @burhandodhy, thanks. Some changes are still needed though.
1. The `Use default...` text and the `@see` tag should be part of the "docblock", i.e., instead of having `//` in front of them, they should be part of the comment with `/**`. [Here](https://github.com/10up/ElasticPress/blob/00d697614a561e337b388a472de32697b3dd3c1a/includes/classes/Feature/Search/Weighting.php#L144) is an example of a docblock with an additional comment.
2. Can you please add a `@since` tag with `4.2.1`?
Thanks!

felipeelia (changes requested):

includes/classes/Indexable.php
burhandodhy: Thanks for pointing out the mistake. Feeback is addressed.

PR #2512, opened 2021-12-09
Do not do any weighting query modifications if there is nothing in config

rebeccahum:
### Description of the Change

I'm seeing unnecessary bits of the ES query being generated on search if we are not actively using the weighting feature. For example, searching with "hello":

```
{
  "from": 0,
  "size": 10,
  "sort": [
    {
      "_score": {
        "order": "desc"
      }
    }
  ],
  "query": {
    "function_score": {
      "query": {
        "bool": {
          "should": [
            {
              "bool": {
                "must": [
                  {
                    "bool": {
                      "must": [
                        {
                          "multi_match": {
                            "query": "hello",
                            "type": "phrase",
                            "fields": [
                              "post_title^1",
                              "post_excerpt^1",
                              "post_content^1",
                              "post_author.display_name^1",
                              "terms.post_tag.name^1",
                              "terms.category.name^1",
                              "terms.ep_custom_result.name^9999"
                            ],
                            "boost": 3,
                            "operator": "AND"
                          }
                        },
                        {
                          "multi_match": {
                            "query": "hello",
                            "fields": [
                              "post_title^1",
                              "post_excerpt^1",
                              "post_content^1",
                              "post_author.display_name^1",
                              "terms.post_tag.name^1",
                              "terms.category.name^1",
                              "terms.ep_custom_result.name^9999"
                            ],
                            "type": "phrase",
                            "slop": 5
                          }
                        }
                      ]
                    }
                  }
                ],
                "filter": [
                  {
                    "match": {
                      "post_type.raw": "post"
                    }
                  }
                ]
              }
            },
            {
              "bool": {
                "must": [
                  {
                    "bool": {
                      "must": [
                        {
                          "multi_match": {
                            "query": "hello",
                            "type": "phrase",
                            "fields": [
                              "post_title^1",
                              "post_excerpt^1",
                              "post_content^1",
                              "post_author.display_name^1",
                              "terms.ep_custom_result.name^9999"
                            ],
                            "boost": 3,
                            "operator": "AND"
                          }
                        },
                        {
                          "multi_match": {
                            "query": "hello",
                            "fields": [
                              "post_title^1",
                              "post_excerpt^1",
                              "post_content^1",
                              "post_author.display_name^1",
                              "terms.ep_custom_result.name^9999"
                            ],
                            "type": "phrase",
                            "slop": 5
                          }
                        }
                      ]
                    }
                  }
                ],
                "filter": [
                  {
                    "match": {
                      "post_type.raw": "page"
                    }
[...]

PR #2836, opened 2022-06-08
Weighting: Add ability to bail early from do_weighting

rebeccahum:
### Description of the Change

This adds a filter which adds a way to _maybe_ bail early from `do_weighting()` if one wanted to keep the Search Weighting engine on. It is to solve the issue described in https://github.com/10up/ElasticPress/pull/2512, but account for any search weighting configurations.

### Alternate Designs

Instead of adding a filter, just bail under set conditions:
- `$weight_config` is empty
- there are no Custom Search Results 

### Possible Drawbacks

N/A, since this offers the most flexibility, while not affecting current EP installs

### Verification Process

1) Spin up EP with no Custom Search Results or Search Weighting set up
2) Add the below snippet:

```
add_filter( 'ep_disable_do_weighting', function ( $disable, $weight_config, $args ) {
	if ( empty( $weight_config ) ) {
		$disable = true;
	}
	return $disable;
}, 10, 3 );
```
3) Check that the query doesn't have any weighting added to it

### Checklist:

### Changelog Entry

Added - new filter ep_disable_search_weighting_config to bail early on do_weighting()

### Credits

Props @rebeccahum

### DIFF
--- a/includes/classes/Feature/Search/Weighting.php
+++ b/includes/classes/Feature/Search/Weighting.php
@@ -589,6 +589,10 @@ class Weighting {
 		 */
 		$weight_config = apply_filters( 'ep_weighting_configuration_for_search', $weight_config, $args );
 
+		if ( apply_filters( 'ep_disable_do_weighting', false, $weight_config, $args ) ) {
+			return $formatted_args;
+		}
+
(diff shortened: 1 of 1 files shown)


felipeelia:
Thanks @rebeccahum. Can you please:
1. Add a docblock to it? We can get it shipped into 4.2.1 (for the `@since` tag)
2. What do you think about also passing `$formatted_args` to it?

rebeccahum:
Done and done!

PR #2837, opened 2022-06-09
Show Facet widgets on taxonomy archives

### LINKED ISSUE #2808: BUG: Facets on taxonomy archives
**Describe the bug**

Currently, it seems the Facets feature is not working on any taxonomy archives, except WooCommerce product categories.

**Steps to Reproduce**
1. Install and enable WooCommerce
2. Activate the WooCommerce and the Facets Feature
3. Sync everything
4. Go to a product category archive
5. See the facets
6. Go to a regular taxonomy archive
7. Do not see facets

**Expected behavior**
Facets would be helpful in any taxonomy, *especially* WooCommerce attributes.

**Additional context**
This might be as simple as adding a `$query->is_tax()` to [this if](https://github.com/10up/ElasticPress/blob/4.2.0/includes/classes/Feature/Facets/Facets.php#L240).

burhandodhy:
### Description of the Change

This PR adds the ability to show the Facet widgets on the taxonomy archives. 

Closes #2808

### Verification Process

- Add the Facet Widget.
- Go to a product category archive
- See the facets

### Changelog Entry

Fixed: Show Facet widgets on taxonomy archives

### Credits

Props @burhandodhy

### DIFF
--- a/includes/classes/Feature/Facets/Facets.php
+++ b/includes/classes/Feature/Facets/Facets.php
@@ -237,11 +237,7 @@ class Facets extends Feature {
 			return false;
 		}
 
-		if ( ! ( ( function_exists( 'is_product_category' ) && is_product_category() )
-			|| $query->is_post_type_archive()
-			|| $query->is_search()
-			|| ( is_home() && empty( $query->get( 'page_id' ) ) ) )
-		) {
+		if ( ! $this->is_facetable_page( $query ) ) {
 			return false;
 		}
 
@@ -554,5 +550,16 @@ class Facets extends Feature {
 		 * @return  {array} New taxonomies
 		 */
 		return apply_filters( 'ep_facet_include_taxonomies', $taxonomies );
+	}
+
+	/**
+	 * Figure out if Facet widget can display on page.
+	 *
+	 * @param  WP_Query $query WP Query
+	 * @since  4.2.1
+	 * @return bool
+	 */
+	protected function is_facetable_page( $query ) {
+		return $query->is_home() || $query->is_search() || $query->is_tax() || $query->is_tag() || $query->is_category() || $query->is_post_type_archive();
 	}
 }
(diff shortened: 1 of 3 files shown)


felipeelia:
@burhandodhy PHP linting found a problem in the code. Do you mind addressing it and then assigning the PR back to be for review? I think we might be being too simplistic in that new list of checks but I can check that later.

burhandodhy:
@felipeelia liniting issues are fixed. For the check, I replicate the same behaviour which we have before.

felipeelia:
> For the check, I replicate the same behaviour which we have before.

Are you sure that 
```
		if ( ! ( ( function_exists( 'is_product_category' ) && is_product_category() )
			|| $query->is_post_type_archive()
			|| $query->is_search()
			|| ( is_home() && empty( $query->get( 'page_id' ) ) ) )
		) {
```
behaves the same way 
```
if ( $query->is_admin() || $query->is_singular() || $query->is_date() ) {
```
does?

Going to an author archive, just as an example, the code seems to have different behaviors.

Also, it is not needed to check if `$query->is_admin()` because we have a `if ( is_admin() ) {` a couple of lines above that already.

burhandodhy:
@felipeelia thanks for pointing out the author archive page. I totally missed out that case. I updated the check.

includes/classes/Feature/Facets/Facets.php
```diff
@@ -237,15 +237,16 @@ public function is_facetable( $query ) {
...
 		}
 
-		if ( ! ( ( function_exists( 'is_product_category' ) && is_product_category() )
-			|| $query->is_post_type_archive()
+		if ( $query->is_home()
 			|| $query->is_search()
-			|| ( is_home() && empty( $query->get( 'page_id' ) ) ) )
-		) {
-			return false;
+			|| $query->is_tax()
```
felipeelia: @burhandodhy let's try to pass `get_facetable_taxonomies()` here (and possibly also get rid of is_tag() and is_category())

burhandodhy: @felipeelia I tried, but `is_tax` doesn't work for the built-in taxonomy page. We don't have any other option other than manually check every page.

`Whether the query is for an existing custom taxonomy archive page. True for custom taxonomy archive pages, false for built-in taxonomies (category and tag archives).`  [source](https://developer.wordpress.org/reference/functions/is_tax/#return)

PR #2842, opened 2022-06-16
filter to add autosuggest styles while indexing

### LINKED ISSUE #2841: Add filter to enqueue autosuggest styles when reindexing
** Problem **
We have a project that doesn't rely on ElasticPress built-in endpoint for autosuggestions, but we do rely on the styles on the frontend. So when we're running a re-index, the styles break the appearance of the site.

** Possible Solution **
Add a filter to allow enqueuing the autosuggest styles even if the site is being reindexed

**Additional context**
This change was added here: https://github.com/10up/ElasticPress/pull/2163 because of a Javascript error, so I think enqueuing just the styles won't cause any issues

hugosolar:
### Description of the Change
The filter `ep_autosuggest_styles_if_idexing` should allow enqueueing autosuggest styles even if the website is being reindexed.
We have a project where we rely on those styles on the frontend but we're not relying on the ElasticPress built-in endpoint for autosuggest so not having those styles break the site appearance

Closes #2841 

### Alternate Designs

For now, we add a fix at the theme level where we're enqueuing the styles if are not enqueued but I think it could be helpful to force the styles if it's needed using a filter

### Possible Drawbacks

since this [change](https://github.com/10up/ElasticPress/pull/2163) was introduced as a fix to a JS error, there are no possible side-effects since are just the styles

### Verification Process
Fork the project and test my branch in my local environment. Ran a full site reindex and then test the filter is working properly

### Checklist:

### Changelog Entry

Added - `ep_autosuggest_styles_if_idexing` filter to enqueue autosuggest styles even if index is running

### Credits

Props @hugosolar

### DIFF
--- a/includes/classes/Feature/Autosuggest/Autosuggest.php
+++ b/includes/classes/Feature/Autosuggest/Autosuggest.php
@@ -385,6 +385,13 @@ class Autosuggest extends Feature {
 	 */
 	public function enqueue_scripts() {
 		if ( Utils\is_indexing() ) {
+			/**
+			 * Filter whether to enqueue autosuggest styles while a index process is running or not
+			 *
+			 * @hook ep_autosuggest_styles_if_idexing
+			 * @param {bool} $show true to enqueue styles
+			 * @return {bool} New value
+			 */
 			if ( apply_filters( 'ep_autosuggest_styles_if_idexing', false ) ) {
 				$this->add_autosuggest_styles();
 			}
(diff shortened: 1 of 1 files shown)


felipeelia:
Given that this is 3 years old already, I'm closing this one out, @hugosolar (I can't imagine you are still waiting for this one to be merged, right? -- Sorry for the super delayed answer! 😆 ).

I think, in general, autosuggest styles should not be something people should rely on, so if anyone else has this problem, I'd recommend reworking their CSS so they don't need that extra CSS file loaded when it is not needed.

PR #2843, opened 2022-06-16
update_index_settings: Account for improper index closing

rebeccahum:
### Description of the Change

We're seeing an issue where during the updating of synonyms, the index closing can time out (even though the index does end up closing) and thus, never re-open, resulting in a `index_closed_exception `. This is because in `update_index_settings()`, it returns earlier and the logic to re-open the index never gets triggered afterwards: https://github.com/10up/ElasticPress/blob/afba6b8cea8b1b7403208767257c5d2d04518988/includes/classes/Elasticsearch.php#L920

### Alternate Designs

N/A

### Possible Drawbacks

N/A

### Checklist:

### Changelog Entry

Fixed: update_index_settings() should account for the index closing action timing out and re-open index if closed

### Credits

Props @rebeccahum

### DIFF
--- a/includes/classes/Elasticsearch.php
+++ b/includes/classes/Elasticsearch.php
@@ -911,16 +911,16 @@ class Elasticsearch {
 		$closed = false;
 		if ( $close_first ) {
 			$closed = $this->close_index( $index );
+			if ( ! $closed ) {
+				// Closing was improper, let's re-open the index.
+				$opened = $this->open_index( $index );
+			}
 		}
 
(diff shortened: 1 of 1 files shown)


felipeelia:
@rebeccahum do you mind telling us in which ES version you are seeing this problem? The general approach seems fine, we are just curious to know if this is something specific to an ES version or not. Thanks!

rebeccahum:
@felipeelia The ES version we're using is 7.5.1.

PR #2851, opened 2022-06-21
Add tests for instant results feature

MARQAS:
### Description of the Change
This PR adds the cypress test for the Instant results feature.

Closes #

### Checklist:

### Changelog Entry
Added: Test for instant results feature.

### Credits

Props @MARQAS


tests/cypress/integration/features/instant-results.spec.js
```diff
@@ -0,0 +1,178 @@
...
+		cy.openWidgetsPage();
+
+		cy.get('.edit-widgets-header-toolbar__inserter-toggle').click();
+		cy.get('.block-editor-inserter__panel-content [class*="search/default"]').click({
+			force: true,
+		});
+
+		cy.get('.edit-widgets-header__actions .components-button.is-primary').click();
+		cy.get('body').should('contain.text', 'Widgets saved.');
+	}
```
felipeelia: @MARQAS why is this needed? Autosuggest tests are running without creating a new search widget.

tests/cypress/integration/features/instant-results.spec.js
```diff
@@ -0,0 +1,178 @@
...
+
+	before(() => {
+		cy.wpCli('eval "echo ElasticPress\\Utils\\get_host();"').then((epHost) => {
+			// Nothing needs to be done if EP.io.
+			if (epHost.stdout.match(/elasticpress\.io/)) {
+				return;
+			}
+			cy.activatePlugin('elasticpress-proxy', 'dashboard');
+		});
```
felipeelia: Can we abstract the check of isEp.io to a new command or simple function?

tests/cypress/integration/features/instant-results.spec.js
```diff
@@ -0,0 +1,178 @@
...
+			}
+			cy.visitAdminPage('admin.php?page=elasticpress');
+			cy.get('.ep-feature-instant-results .settings-button').click();
+			cy.get('.requirements-status-notice').should(
+				'contain.text',
+				'You are using a custom proxy. Make sure you implement all security measures needed',
+			);
+			cy.get('.ep-feature-instant-results .input-wrap').should('not.have.class', 'disabled');
+		});
+	});
```
felipeelia: So, e2e tests [don't follow the same principles](https://docs.cypress.io/guides/references/best-practices#Creating-tiny-tests-with-a-single-assertion) as unit tests and we could take advantage of that here. As the initial steps are the same as the previous one, let's merge them into a single test checking both things. Remember to change the description accordingly, okay?

felipeelia (changes requested):
@MARQAS, having in mind that we don't need small atomic tests we can merge all the following tests in just one:

Can see instant results list
Can display the number of results
Can show the modal in the same state after a reload
Can update the results after changing the search term

I am not 100% sure about the reason why it is failing on EP.io but it is probably a timeout. Can you try capturing the AJAX request and waiting for it to be completed before checking if "Blog" is visible?

assets/js/autosuggest.js
```diff
@@ -676,9 +676,10 @@ function init() {
 	 */
 	const prepareInputForAutosuggest = (input) => {
 		/**
-		 * Skip facet widget search fields.
+		 * Skip facet widget search fields and instant results.
 		 */
-		if (input.classList.contains('facet-search')) {
+		const ignoredClasses = ['facet-search', 'ep-search-input'];
```
felipeelia, line 681: We need to check if the comment search field should also be added here (props @JakePT )

PR #2861, opened 2022-06-24
Add new filter ep_skip_autosave_sync and enable autosave sync for protected_content feature

rebeccahum:
### Description of the Change

Problem: When a user has the `protected_content` feature enabled, and they work on a draft but they never hit "Save draft" and instead, rely on the WordPress autosave, it won't show up in the Draft post listing screen (/wp-admin/edit.php?post_status=draft&post_type=post):

Screen Shot 2022-06-24 at 1 43 00 PM

This is because it never gets synced, as we disable it here:

https://github.com/10up/ElasticPress/blob/aaf2a3342b61c5df958f874cfa1fb434c244201f/includes/classes/Indexable/Post/SyncManager.php#L284-L289

This PR does two things:
- Introduces the `ep_skip_autosave_sync` filter 
- Returns false for the aforementioned filter when the `protected_content` feature is enabled

### Alternate Designs

I'm not sure if it would be beneficial to pass in the `__FUNCTION__` for the second parameter during `apply_filters()` if one wanted to customize disabling the autosave sync on function-basis, since we have `action_set_object_terms`, `action_sync_on_update`, etc. to make it more granular:

```
		if ( apply_filters( 'ep_disable_autosave_sync', true, __FUNCTION__ ) ) {
			if ( defined( 'DOING_AUTOSAVE' ) && DOING_AUTOSAVE ) {
				// Bypass saving if doing autosave
				// @codeCoverageIgnoreStart
				return;
				// @codeCoverageIgnoreEnd
			}
		}
```

Curious on your thoughts @felipeelia for that.

### Possible Drawbacks

N/A.

### Verification Process

1) Enable the `protected_content` feature
2) Start drafting a new post and wait for WordPress to autosave
3) Go to the Draft post listing screen at `/wp-admin/edit.php?post_status=draft&post_type=post` and expect to see the autosaved draft

### Checklist:

### Changelog Entry

Fixed: autosaved drafts not showing up in draft post listing when using protected_content

### Credits

Props @rebeccahum


felipeelia:
Hey @rebeccahum, I'm finally getting to this one. I like the idea of adding `__FUNCTION__` as a parameter so people can opt out of some scenarios, can you please add that too?

In addition to that, do you think it'd be possible to add some tests? End-to-end tests checking if that works as expected and also does not affect published posts is what I have in mind. Thanks!

rebeccahum:
@felipeelia Done and done!

tests/cypress/support/commands.js
```diff
@@ -289,3 +289,30 @@ Cypress.Commands.add('deactivatePlugin', (slug, method = 'dashboard', mode = 'si
...
+	cy.get('h1.editor-post-title__input, #post-title-0').clearThenType(newPostData.title);
+	cy.get('.block-editor-default-block-appender__content').type(newPostData.content);
+
+	/**
+	 * Give Elasticsearch some time to process the new post.
+	 *
+	 * @todo instead of waiting for an arbitrary time, we should ensure the post is stored.
+	 */
+	// eslint-disable-next-line cypress/no-unnecessary-waiting
+	cy.wait(80000);
```
felipeelia: @rebeccahum I think this is needed to wait until the auto-draft is saved, right? Instead of adding this amount of time to the overall test time, what do you think about adding a 1-line plugin [here](https://github.com/10up/ElasticPress/tree/develop/tests/cypress/wordpress-files/test-plugins) just to set the `AUTOSAVE_INTERVAL` constant? Then we would enable it at the beginning of the command and disable it right after.

includes/classes/Feature/ProtectedContent/ProtectedContent.php
```diff
@@ -56,6 +56,7 @@ public function setup() {
 		add_filter( 'ep_post_sync_args', [ $this, 'include_post_password' ], 10, 2 );
 		add_filter( 'ep_post_sync_args', [ $this, 'remove_fields_from_password_protected' ], 11, 2 );
 		add_filter( 'ep_search_post_return_args', [ $this, 'return_post_password' ] );
+		add_filter( 'ep_skip_autosave_sync', '__return_false', 10, 2 );
```
felipeelia: @rebeccahum any reason to pass `10, 2` here?

felipeelia (changes requested):

rebeccahum:
@felipeelia Thanks for the feedback, I think I got it working now!

PR #2862, opened 2022-06-27
refactor meta key allowed check

pschoffer:
### Description of the Change

When a meta key is deleted from a post, this does NOT trigger reindex of that post due to the way of allowed metas are detected.

This change makes it so that we have the logic that evaluates if given meta_key is allowed separate from the logic that prepares metas for indexing. That way we can trigger even for meta key that is not being prepared to indexed (because it was just deleted) but is allowed to be in index (was in the indexed document before).

### Verification Process

(sorry the steps are specific to our dev-env, but I guess they still make sense?)

1) Add this filter impl:
```
add_filter( 'vip_search_post_meta_allow_list', function ( $allow, $post = null ) {
	$allow['testing'] = true;
	return $allow;
}, 10, 2 );
```
2) Add meta - `vip dev-env exec --  wp post meta set 1 testing foobar`
3) Check the document in ES - `vip dev-env exec --  wp vip-search documents get post 1 --format=json | jq '.[0].meta'`
4) Remove meta `vip dev-env exec --  wp  post meta delete 1 testing`
5) Check the document in ES - `vip dev-env exec --  wp vip-search documents get post 1 --format=json | jq '.[0].meta'`

### Checklist:

- [ ] I have read the [**CONTRIBUTING**](https://github.com/10up/.github/blob/trunk/CONTRIBUTING.md) document.

### Changelog Entry

Fixed - Post reindex on meta deletion

### Credits

@pschoffer


pschoffer:
:wave: @felipeelia I added a test, change few typos and switched the version.

Please, let me know if that is what you meant in slack.

felipeelia:
Hey @pschoffer, thanks for the changes. The PHP Lint action detected a warning with a space alignment, do you mind fixing that before we merge it?

Also, is there any reason why you decided to not call maybe_unserialize on values in filter_allowed_metas()? I imagine that was because of how things were named but wanted to confirm.

includes/classes/Indexable/Post/Post.php
```diff
@@ -871,9 +867,49 @@ public function prepare_meta( $post ) {
...
+			 * @param  {array} $prepared_meta Prepared meta
+			 * @param  {WP_Post} $post Post object
+			 * @since  3.4
+			 * @return  {array} Prepared meta
+			 */
+			return apply_filters( 'ep_prepared_post_meta', [], $post );
+		}
+
+		$filtered_metas = $this->filter_allowed_metas( $meta, $post );
+		$prepared_meta = [];
```
felipeelia:
```suggestion
		$filtered_metas = $this->filter_allowed_metas( $meta, $post );
		$prepared_meta  = [];
```

felipeelia (changes requested):

pschoffer:
Hi @felipeelia ,
sorry I missed the ping.

My thinking around `maybe_unserialize` was that I was going for separate concerns. So `filter_allowed_metas` only does filtering, but doesn't mutate the values in any way. 

`maybe_unserialize` is then still called later in `prepare_meta`, but not in `is_meta_allowed` where it is not needed.

PR #2864, opened 2022-06-27
Add E2E for Password Protected Post

### LINKED ISSUE #2614: Cypress tests

The following tests need to be written in Cypress:

### General
- [ ] Installation process: check new step 3 and changes introduced in #2734
- [ ] Check Health tools test
- [ ] Update (if the version doesn’t require a reindex)

### New Sync
- [ ] Check if post search works with EP during regular sync
- [ ] Check if post search does not work with EP during full sync

### Post Search
- [ ] Check if term edits also update related posts

### Password Protected Content
- [ ] If the user is an admin, they should be able to see the post title displayed on the front page, and on search results.
- [ ] If the user is a subscriber, they should be able to the post title displayed on the front page, and on search results.
- [ ] If the user is logged out, they should be able to the post title displayed on the front page, but not on the search result

### WooCommerce (#2923)
- [x] Check Order IDs search (admin)
- [x] Check orderby parameter (frontend)
- [x] Having orders for more than one user in the system, visit my-account/orders page and check if only user’s order are returned

### Autosuggest
- [ ] Check highligh
- [ ] Check cyrillic characters
- [ ] Check behavior while indexing

### Terms (#2931)
- [x] With Protected Content enabled, searching for a Term in the Dashboard should use ElasticPress
- [x] Ordering by any column should work and go through ElasticPress
- [x] If user deletes a term, the term should not be displayed anymore
- [x] If user deletes a parent term, all children terms should be updated
- [x] Create these three post tags (Test Post, Test Ruth, and The Most Fun Thing) and search for “the most fun thing”. Only that should be returned

### Comments (#2955)
- [x] If user activates the Comments feature, it should sync approved comments.
- [x] If user posts an anonymous comment, it should not be synced 
- [x] If user approves a pending comment, it should be synced
- [x] If user disables the options "Comment must be manually approved" and "Comment author must have a previously approved comment" on Settings > Discussion and post an anonymous comment, it should be synced.
- [x] If feature is activated, user should see “ElasticPress - Comment” widget in the dashboard.
- [x] If WooCommerce is activated, it should sync reviews

### Instant Results
- [x] If not in ElasticPress.io nor using a custom PHP proxy, user should not be able to enable the feature.
- [x] If using https://github.com/10up/elasticpress-proxy, user should see a warning about security measures
- [x] In the frontend, typing something in a search form and hitting enter should open the Instant Results popup
- [x] Changing filters and search term should update results
- [ ] Different taxonomies with the same name but applied to different post types (Categories vs. Product Categories) should have the post type label appended in the filter displayed in the front end
- [ ] Filter options should display the counter of their available results
- [x] Navigating through filters should change the URL
- [x] Reloading a filtered instant result should open the modal in the same state
- [ ] Should be able to click outside of the popup modal

### Weighting Engine
TBD

### WP-CLI (#2926)

- [x] set-algorithm-version
- [x] get-algorithm-version
- [x] get-mapping
- [x] get-cluster-indexes
- [x] get-indexes
- [ ] recreate-network-alias
- [x] epio-set-autosuggest
- [x] clear-index
- [x] stop-indexing
- [x] request
- [x] index: Check new `--static-bulk` flag in WP-CLI

burhandodhy:
### Description of the Change

This PR adds the Cypress test for the Password Protected Post

Closes #2614 

### Checklist:

- [ ] I have read the [**CONTRIBUTING**](https://github.com/10up/.github/blob/trunk/CONTRIBUTING.md) document.

### Changelog Entry

Added - E2E tests for Password Protected Post

### Credits

Props @burhandodhy


includes/classes/Indexable/Post/SyncManager.php
```diff
@@ -884,4 +888,39 @@ protected function is_tax_max_count_bigger_than_items_per_cycle( \WP_Taxonomy $t
...
+		 * Returning a non-null value will effectively short-circuit the function.
+		 *
+		 * @since 4.6.0
+		 * @hook ep_pre_kill_sync_for_password_protected
+		 * @param {null} $new_skip     Whether should skip or not before checking for a password
+		 * @param {bool} $current_skip Current value
+		 * @param {int}  $object_id    The Post ID
+		 * @return {null|bool} New value of $skip or `null` to keep default behavior.
+		 */
+		$skip_filter = apply_filters( 'ep_pre_kill_sync_for_password_protected', null, $skip, $object_id );
```
burhandodhy, line 913: Returning null doesn't look good to me because `ep_post_sync_kill` is supposed to have a bool.
 https://github.com/10up/ElasticPress/blob/burhan/add-tests-for-password-protected-posts/includes/classes/Indexable/Post/SyncManager.php#L249

felipeelia: This is a short-circuit filter, @burhandodhy, and usually, this is the way they work. It is just a simple way to cut the behavior. This is a different filter, though. Did I get your comment right?

PR #2871, opened 2022-06-29
Add E2E for Custom Results Feature

burhandodhy:
### Description of the Change
This PR adds the E2E tests for the Custom Results. 

### Credits

Props @burhandodhy


tests/cypress/integration/features/custom-results.spec.js
```diff
@@ -0,0 +1,74 @@
...
+	after(() => {
+		cy.login();
+		cy.visitAdminPage('edit.php?post_type=ep-pointer');
+		cy.get('#cb-select-all-1').click();
+		cy.get('#bulk-action-selector-top').select('trash');
+		cy.get('#doaction').click();
+
+		cy.visitAdminPage('edit.php?post_status=trash&post_type=ep-pointer');
+		cy.get('.tablenav.top #delete_all').click();
+	});
```
felipeelia: Can we remove these things before the tests rather than after them? (I've [linked to this](https://docs.cypress.io/guides/references/best-practices#Using-after-or-afterEach-hooks) already in another PR but leaving here to make it easier to find)

tests/cypress/integration/features/custom-results.spec.js
```diff
@@ -0,0 +1,74 @@
...
+			'.pointers .pointer:first-of-type .dashicons-menu',
+			'.pointers .pointer:last-of-type .dashicons-menu',
+		).then(() => {
+			cy.get('.pointers .pointer .title').each((post) => {
+				cy.wrap(post)
+					.invoke('text')
+					.then((text) => searchResult.push(text));
+			});
+			cy.get('#publish').click();
+		});
```
felipeelia: Do you mind documenting what this is doing on the screen? It is not super clear to me.

tests/cypress/integration/features/custom-results.spec.js
```diff
@@ -0,0 +1,74 @@
...
+
+		cy.visit(`?s=${searchTerm}`);
+
+		cy.get('article .entry-title').each((post, index) => {
+			cy.wrap(post).invoke('text').should('eq', searchResult[index]);
+		});
+	});
+
+	it('Can add the post in result and verify the result on search', () => {
```
felipeelia: This is also not clear to me. Isn't it adding Fantastic as a custom result but with a post that already has that in it? Would it behave differently if the custom result simply failed?

felipeelia (changes requested):

tests/cypress/integration/features/custom-results.spec.js
burhandodhy: Done

tests/cypress/integration/features/custom-results.spec.js
burhandodhy: Done

tests/cypress/integration/features/custom-results.spec.js
burhandodhy: This creates a new search and adds a custom post on search and than verify the result on frontend

tests/cypress/integration/features/custom-results.spec.js
```diff
@@ -0,0 +1,89 @@
...
+		cy.get('body').then(($body) => {
+			if ($body.find('#bulk-action-selector-top').length > 0) {
+				cy.get('#cb-select-all-1').click();
+				cy.get('#bulk-action-selector-top').select('trash');
+				cy.get('#doaction').click();
+
+				cy.visitAdminPage('edit.php?post_status=trash&post_type=ep-pointer');
+				cy.get('.tablenav.top #delete_all').click();
+			}
+		});
```
felipeelia: Thanks for moving this @burhandodhy! I should've thought about this before but what do you think about deleting these in a wp-cli call?

tests/cypress/integration/features/custom-results.spec.js
```diff
@@ -0,0 +1,89 @@
...
+		cy.visitAdminPage('post-new.php?post_type=ep-pointer');
+		cy.intercept('GET', 'wp-json/elasticpress/v1/pointer_preview*').as('ajaxRequest');
+
+		cy.get('#titlewrap input').type(searchTerm);
+		cy.wait('@ajaxRequest').its('response.statusCode').should('eq', 200);
+
+		cy.intercept('GET', 'wp-json/elasticpress/v1/pointer_search*').as('ajaxRequest');
+
+		// search for the post.
+		cy.get('.search-pointers').type('hello world');
```
felipeelia: Where is this post created @burhandodhy? If it is created by the initial import, do you mind creating it here (and deleting it before running the test again?)

felipeelia (approved):

PR #2880, opened 2022-07-01
Move search algorithms to separate classes

### LINKED ISSUE #2846: New SearchAlgorithm class

The solution outlined below is the first step toward #2086.

Right now we have at least 3 different search algorithms. In post searches ([here](https://github.com/10up/ElasticPress/blob/4.2.0/includes/classes/Indexable/Post/Post.php#L1412)) it can be selected using the `ep_search_algorithm_version` filter but for Users and other indexable ([here](https://github.com/10up/ElasticPress/blob/4.2.0/includes/classes/Indexable/User/User.php#L480)) it is not easily changeable.

Creating a class with the following attributes/methods would make it easier to add new algorithms and allow the user to select them (also developers would simply create a new class and use a filter to set EP to use it.)

- Name
- Description
- Search Algorithm (receiving a search term and an array of fields and returning the array to be used in the ES query)

In this first moment, we do not need any additional UI or changes in the current behavior. It is supposed to be just the build of an underlying layer for future iterations.

felipeelia:
### Description of the Change

This PR addresses the changes outlined in #2846.

To-do:
- [x] Add methods to make slug, name, and description mandatory
- [x] Adjust `@since` tags
- [x] Add tests
- [x] Fix `post_` retrocompatibility

Closes #2846

### Changelog Entry

Changed - Search algorithms moved to separate classes
Deprecated -
ep_formatted_args_query
ep_match_phrase_boost
ep_match_boost
ep_fuzziness_arg
ep_match_fuzziness
ep_match_cross_fields_boost

in favor of

ep_post_formatted_args_query
ep_post_match_phrase_boost
ep_post_match_boost
ep_post_fuzziness_arg
ep_post_match_fuzziness
ep_post_match_cross_fields_boost

### Credits

Props @felipeelia


oscarssanchez:
So far it seems to be working for me!

PR #2883, opened 2022-07-08
Performance: Use get_term() instead of term_exists() for get_term_tree()

rebeccahum:
### Description of the Change
This is a follow-up to the change in https://github.com/10up/ElasticPress/pull/2687. It'd be better to use `get_term()` to utilize the Object cache, since `term_exists()` does a DB call and thus, if there are several terms with the same parent, that generates a lot of queries.

### Alternate Designs

N/A.

### Possible Drawbacks

N/A.

### Verification Process

Did the same steps described in https://github.com/10up/ElasticPress/pull/2687 but ensure that object caching is utilized.

### Checklist:

### Changelog Entry

Changed: Improved performance in get_term_tree()

### Credits

Props @rebeccahum

### DIFF
--- a/includes/utils.php
+++ b/includes/utils.php
@@ -353,7 +353,9 @@ function get_term_tree( $all_terms, $orderby = 'count', $order = 'desc', $flat =
 				$terms_map[ $term->term_id ] = $term;
 			}
 
-			if ( empty( $term->parent ) || is_wp_error( get_term( $term->parent, $term->taxonomy ) ) ) {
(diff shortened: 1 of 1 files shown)


felipeelia:
@rebeccahum looking at the [get_term()](https://developer.wordpress.org/reference/functions/get_term/#return) documentation, it can also return `null` for some specific failures, so we'll probably have to test for that in addition to the `is_wp_error()` you already have there.

rebeccahum:
@felipeelia I've updated it to account for a null return.

PR #2906, opened 2022-07-26
Fix: recreate-network-alias throw error on single site

### LINKED ISSUE #2868: BUG: recreate-network-alias CLI command returns an error without explanation on single site installs
**Describe the bug**
The WP CLI command `wp elasticpress recreate-network-alias` returns an error without explanation when run on a single site install.

**Steps to Reproduce**

1. On a single site install run `wp elasticpress recreate-network-alias`.

**Expected behavior**
The command should return a message that explains that it does nothing for a single site install.

burhandodhy:
### Description of the Change

This PR adds the check that will check throw the error message if EP is not activated on the network

Closes #2868 

### How to test the Change

- Run `wp elasticpress recreate-network-alias` command on single site.
- Command should throw the error message `ElasticPress is not activated on network.`. 

### Changelog Entry

> Fixed - recreate-network-alias CLI command returns an error without explanation on single site 

### Credits

Props @burhandodhy 

### Checklist:

### DIFF
--- a/includes/classes/Command.php
+++ b/includes/classes/Command.php
@@ -571,7 +571,7 @@ class Command extends WP_CLI_Command {
 		$this->index_occurring();
 
 		if ( ! defined( 'EP_IS_NETWORK' ) || ! EP_IS_NETWORK ) {
-			WP_CLI::error( esc_html__( 'ElasticPress is not activated on network.', 'elasticpress' ) );
+			WP_CLI::error( esc_html__( 'ElasticPress is not network activated.', 'elasticpress' ) );
 		}
 
 		$indexables = Indexables::factory()->get_all( false );
--- a/includes/classes/Command.php
+++ b/includes/classes/Command.php
@@ -570,6 +570,10 @@ class Command extends WP_CLI_Command {
 		$this->connect_check();
 		$this->index_occurring();
 
+		if ( ! defined( 'EP_IS_NETWORK' ) || ! EP_IS_NETWORK ) {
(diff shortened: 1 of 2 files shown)


tests/cypress/integration/wp-cli.spec.js
```diff
@@ -144,6 +144,12 @@ describe('WP-CLI Commands', () => {
 
 	it('Can recreate the alias index which points to every index in the network if user runs wp elasticpress recreate-network-alias command', () => {});
 
+	it('wp elasticpress recreate-network-alias throws an error message if the plugin is not activated on the network', () => {
```
felipeelia: @burhandodhy we should keep the pattern here, starting it with `Can...`, so it reads `it can...`. It could be a lengthy-phrase like `it('Can throw an error while running wp elasticpress recreate-network-alias if the plugin is not network activated')`.

includes/classes/Command.php
```diff
@@ -570,6 +570,10 @@ public function recreate_network_alias( $args, $assoc_args ) {
 		$this->connect_check();
 		$this->index_occurring();
 
+		if ( ! defined( 'EP_IS_NETWORK' ) || ! EP_IS_NETWORK ) {
+			WP_CLI::error( esc_html__( 'ElasticPress is not activated on network.', 'elasticpress' ) );
```
felipeelia: Can we rephrase that to `ElasticPress is not network activated.` @burhandodhy? This will match how we are phrasing that in [our docs](http://10up.github.io/ElasticPress/tutorial-features.html).

felipeelia (changes requested):

PR #2911, opened 2022-07-27
New add_synonyms_to_settings method + default_search analyzer

### LINKED ISSUE #2877: Create `default_search` analyzer while setting up synonyms

Currently, ElasticPress has only a [`default` analyzer](https://github.com/10up/ElasticPress/blob/4.2.1/includes/mappings/post/7-0.php#L64) used on both index and search time. That is usually not a problem but when we add the Synonym's token filter [here](https://github.com/10up/ElasticPress/blob/4.2.1/includes/classes/Feature/Search/Synonyms.php#L479) we end up having  `word_delimiter_graph` and `synonym_graph` being applied at the same time.

As that is [not fully supported](https://github.com/elastic/elasticsearch/issues/32694#issuecomment-411800889), we eventually get a `[illegal_argument_exception] startOffset must be non-negative, and endOffset must be >= startOffset, and offsets must not go backwards startOffset=??,endOffset=??,lastStartOffset=?? for field '??'` error like seen in #2838.

As we don't index everything after saving synonyms, we could apply the `synonym_graph` only during search time. Having `word_delimiter_graph` applied only during index time should not be a problem as well.

To have that happen, we need to create a `default_search` analyzer during the synonyms setup instead of simply adding  `synonym_graph` to the default one.

felipeelia:
### Description of the Change

This PR adds synonyms in a new default_analyzer, so its filter doesn't conflict with word_delimiter_graph during index time.

**It still needs some tests**

Closes #2877

### Changelog Entry

> Fixed - Synonyms are now applied only during search time, fixing an incompatibility with the `word_delimiter_graph` filter.

### Credits

Props @felipeelia

### DIFF
--- a/includes/classes/Feature/Search/Synonyms.php
+++ b/includes/classes/Feature/Search/Synonyms.php
@@ -355,9 +355,8 @@ class Synonyms {
 	 * @return array
 	 */
 	public function add_search_synonyms( $mapping, $index ) {
-		$synonyms    = $this->get_synonyms();
-		$indices     = $this->get_affected_indices();
-		$filter_name = $this->get_synonym_filter_name();
+		$synonyms = $this->get_synonyms();
+		$indices  = $this->get_affected_indices();
 
 		// Ensure we should affect this mapping.
 		if ( ! in_array( $index, $indices, true ) ) {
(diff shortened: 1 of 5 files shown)


felipeelia:
Moving synonyms to be applied only during search time revealed a problem with how we use "alternatives".

As per [Elasticsearch docs](https://www.elastic.co/guide/en/elasticsearch/reference/current/analysis-synonym-tokenfilter.html):
> Explicit mappings match any token sequence on the LHS of "=>" and **replace** with all alternatives on the RHS.

So, if we have `shoes => sneakers`, all `shoes` in a post will be replaced by `sneakers`. Right now, that replacement happens on both index and search times going unnoticed most of the time. There is one scenario where that is verifiable though:

1. Empty all alternatives in the Synonyms Screen and save
2. Create a post with `This is my content with shoes` as content and save. It will be indexed as `shoes`.
3. Search for `shoes` and see the post there
4. Add `shoes => sneakers` as alternatives
5. Search for `shoes` and see the post gone
6. Save the post again. Now `shoes` will be replaced with `sneakers` in the index
7. Search for `shoes` and see the post back

To have alternatives to the way we phrase it right now, it is necessary to add the main term in the list, like `shoes => sneakers, shoes`.

The questions would be:
1. Do we want to keep the intended behavior (hiding from the user the main term in the list of "alternatives")? In this case, the user would see `shoes => sneakers` but we would send `shoes => shoes, sneakers` to Elasticsearch.
2. Do we want to rephrase that in a way the user can decide if that is a replacement or an alternative? In this case, it would be up to the user to decide between `shoes => sneakers` or `shoes => shoes, sneakers`.

felipeelia:
After some internal discussion, we will have some UX treatment on the screen to clarify to users how the feature works. Ideas so far are to rename "Primary Term" to "Search Term" and "Comma separated list of alternatives" to "Match list" or "Search rewrites".

nickolas-kola:
UX deliverable created, please refer to internal channels.

felipeelia:
Closing in favor of #3610

PR #2913, opened 2022-07-28
Change term objects in `format_hits_as_terms` to `WP_Term` instead of `stdClass`

### LINKED ISSUE #2912: BUG: the `format_hits_as_terms` returns a term as `stdClass` instead of `WP_Term`
**Describe the bug**
I've been trying to hunt down an issue where a filter on wp-admin/edit-tags.php?taxonomy=post_tag&post_type=post URL expects data to come back as an instance of `WP_Term` but it instead comes back as `stdClass`, and I think it might be related to the `format_hits_as_terms` method in [these lines](https://github.com/10up/ElasticPress/blob/develop/includes/classes/Indexable/Term/QueryIntegration.php#L206-L251). It looks to me like this method is returning `$term` as an instance of `stdClass`, never converting it to `WP_Term` even though it does use the standard formatting for a `WP_Term` object.

I first noticed this using The Events Calendar plugin because they have a call to the `post_tag_row_actions` where they force `WP_Term` on the parameter. But in any case, the [WP docs](https://developer.wordpress.org/reference/hooks/taxonomy_row_actions/) specify WP_Term as the object type, too.

**Steps to Reproduce**

1. Index post_tag, at least, with some post tags.
2. Install The Events Calendar.
3. Go to `/wp-admin/edit-tags.php?taxonomy=post_tag&post_type=post` and search for a value that will return some tags.
4. The error I get is like this:

```php
Fatal error: Uncaught Error: Tribe\Events\Taxonomy\Taxonomy_Provider::event_tag_actions(): Argument #2 ($tag) must be of type WP_Term, stdClass given, called in /wp-includes/class-wp-hook.php on line 307
in /wp-content/plugins/the-events-calendar/src/Tribe/Taxonomy/Taxonomy_Provider.php on line 53

Call stack:

Tribe\E\T\Taxonomy_Provider::event_tag_actions()
wp-includes/class-wp-hook.php:307
WP_Hook::apply_filters()
wp-includes/plugin.php:191
apply_filters()
wp-admin/includes/class-wp-terms-list-table.php:539
WP_Terms_List_Table::handle_row_actions()
wp-admin/includes/class-wp-list-table.php:1446
WP_List_Table::single_row_columns()
wp-admin/includes/class-wp-terms-list-table.php:345
WP_Terms_List_Table::single_row()
wp-admin/includes/class-wp-terms-list-table.php:253
WP_Terms_List_Table::display_rows_or_placeholder()
wp-admin/includes/class-wp-list-table.php:1289
WP_List_Table::display()
wp-admin/edit-tags.php:613
```

**Expected behavior**
I would expect to be able to search post tags, and I think in this case I would expect ElasticPress to return the term results as `WP_Term` objects.

**Environment information**
 - Device: MacBook Pro (for my local environment)
 - OS: macOS 12.4 (for my local environment)
 - WordPress version:6.0.1 (for both local and remote)
 
 - ElasticPress version: I'm running 3.6.5 that comes with WordPress VIP hosting in their mu-plugins folder, but I also see this bug on the `develop` branch.
 - Elasticsearch version: 7.17.4 in my local environment. I'm not sure what version VIP runs.
 - Where do you host your Elasticsearch server? I'm hosting mine locally, but I see the same issue on WordPress VIP.
 - What ElasticPress features do you have currently active? Post Search, Protected Content, Related Posts, Terms, Custom Search Results

**Additional context**
I also tried locally editing the filter that The Events Calendar uses. I made it like this:

```php
public function event_tag_actions( $actions, $tag ) {
    error_log( 'tag is ' . print_r( $tag, true ) );
    return $this->container->make( Event_Tag::class )->event_tag_actions( $actions, $tag );
}
```

Then when I run that log, here's what it returns:

```php
[27-Jul-2022 21:49:54 UTC] tag is stdClass Object
(
    [ID] => 8134
    [site_id] => 1
    [term_id] => 8134
    [name] => 2008 election
    [slug] => 2008-election
    [term_taxonomy_id] => 214476
    [taxonomy] => post_tag
    [description] => 
    [parent] => 0
    [count] => 1
    [meta] => Array
        (
        )

    [elasticsearch] => 1
    [filter] => display
)
```

jonathanstegall:
### Description of the Change

The description of the `format_hits_as_terms` method reads like this:

```php
/**
 * Format the ES hits/results as term objects.
 *
 * @param  array $terms The terms that should be formatted.
 * @param  array $new_terms Array of terms from cache.
 * @param  array $query_vars Query variables.
 * @since  3.1
 * @return array
*/
```

This pull request makes the term objects into `WP_Term` objects instead of `stdClass` objects. This brings it in line with how WordPress uses this result in actions like `post_tag_row_actions`, which specify that the term parameter should be a `WP_Term` object.

I'm not aware of any drawbacks, but I do realize that this method has been returning `stdClass` objects, even if incorrectly, for some time. I think likely this is causing issues unless the code using it is not using standard WordPress methods, but that is worth noting.

An alternative would be to add a filter instead, allowing developers to use `WP_Term` on their own. Ex: `$term = apply_filters( 'ep_wp_query_terms_object', $term );`. I don't think that is ideal, but if it's necessary for backward compatibility I would understand that.

Closes #2912 

### How to test the Change

The way I found this was using The Events Calendar:

1. Index `post_tag`, at least, with some post tags.
2. Install The Events Calendar.
3. Go to `/wp-admin/edit-tags.php?taxonomy=post_tag&post_type=post` and search for a value that will return some tags. Instead of giving a fatal error, it will return the same data but formatted as a `WP_Term` object. Verify by logging in [these lines](https://github.com/the-events-calendar/the-events-calendar/blob/master/src/Tribe/Taxonomy/Taxonomy_Provider.php#L53-L55) like this:

```php
public function event_tag_actions( $actions, WP_Term $tag ) {
    error_log( 'tag is ' . print_r( $tag, true ) );
    return $this->container->make( Event_Tag::class )->event_tag_actions( $actions, $tag );
}
```

### Changelog Entry

> Fixed - Change term objects in `format_hits_as_terms` to use `WP_Term` instead of `stdClass` to match WordPress expectations.

### Credits

Props @jonathanstegall

### Checklist:

I don't see any documentation or relevant tests, but am willing to update those if they do apply.

### DIFF
--- a/includes/classes/Indexable/Term/QueryIntegration.php
+++ b/includes/classes/Indexable/Term/QueryIntegration.php
@@ -241,6 +241,7 @@ class QueryIntegration {
 			}
 
 			$term->elasticsearch = true; // Super useful for debugging.
+			$term = new \WP_Term( $term ); // Necessary for WordPress actions that expect WP_Term as the object type.
 
 			if ( $term ) {
 				$new_terms[] = $term;


jonathanstegall:
I'm noticing that one of the tests failed but it's not clear to me that it's related to the code. Is there anything I should do with this?

PR #2914, opened 2022-08-01
Limit WooCommerce integration with ability to allow it

### LINKED ISSUE #2809: Limit WooCommerce integration

Currently, the [translate_args](https://github.com/10up/ElasticPress/blob/4.2.0/includes/classes/Feature/WooCommerce/WooCommerce.php#L238) and the [should_integrate_with_query](https://github.com/10up/ElasticPress/blob/4.2.0/includes/classes/Feature/WooCommerce/WooCommerce.php#L1090) methods are too broad. For example, running:

```
		$posts = get_posts(
			[
				'post_type' => 'product',
				'post__in'  => [ 1, 2, 3 ],
			]
		);
```
Will add `ep_integrate` to the query. If the theme has 3 or 4 calls like that, it would result in 3 or 4 external requests, most likely making performance worst rather than helping it.

We need to limit that integration to requests that really make sense, obviously allowing users to overwrite that if needed.

MARQAS:
### Description of the Change
Filter to allow WooCommerce. It will limit the usage of WooCommerce integration to the places where it is actually needed. 

Closes #2809 

### Changelog Entry

> Added - Filter to allow/limit WooCommerce integration
> Changed - `should_integrate_with_query` function

### Credits

Props @MARQAS 

### Checklist:


includes/classes/Feature/WooCommerce/WooCommerce.php
```diff
@@ -1122,36 +1126,20 @@ protected function should_integrate_with_query( $query ) {
...
+		if ( defined( 'WC_API_REQUEST' ) && WC_API_REQUEST ) {
 			return false;
 		}
 
-		/**
-		 * Cant hook into WC API yet
-		 */
-		if ( defined( 'WC_API_REQUEST' ) && WC_API_REQUEST ) {
-			return false;
+		if ( method_exists( $query, 'is_search' ) && ! $query->is_search() && empty( $query->query_vars['s'] ) ) {
```
felipeelia: @MARQAS I think we can tweak this a little bit. What do you think about checking for `$query->query_vars['ep_integrate']` right before this if? I think we will want to keep running through ES queries that have ep_integrate set. That change will also reflect on some of the changes you made for tests.

Do you think you could write in which scenarios we would expect the query to run through ES or not, and then write unit tests to test that?

felipeelia (changes requested):

includes/classes/Feature/WooCommerce/WooCommerce.php
```diff
@@ -1113,45 +1114,36 @@ protected function should_integrate_with_query( $query ) {
...
-
-		$post_parent = $query->get( 'post_parent', false );
-
-		/**
-		 * Do nothing for single product queries
-		 */
-		if ( ! empty( $product_name ) || $query->is_single() ) {
+		if ( ! Utils\is_integrated_request( $this->slug ) ) {
 			return false;
 		}
```
felipeelia: We will need to reverse the order of these two ifs to reflect how we have it in the other places: first, we check if `is_integrated_request`, and after we check for `ep_integrate`. In the current codebase, an AJAX request needs both `ep_integrate` and its own filter `ep_integrate...` to run through ElasticPress.

includes/classes/Feature/WooCommerce/WooCommerce.php
```diff
@@ -1113,45 +1114,36 @@ protected function should_integrate_with_query( $query ) {
...
 
-		/**
-		 * ElasticPress does not yet support post_parent queries
-		 */
-		if ( ! empty( $post_parent ) ) {
+		if ( defined( 'WC_API_REQUEST' ) && WC_API_REQUEST ) {
 			return false;
 		}
```
felipeelia: This needs to be moved to before any `return true;` we may have, as things won't work properly in their API request.

includes/classes/Feature/WooCommerce/WooCommerce.php
```diff
@@ -1113,45 +1114,36 @@ protected function should_integrate_with_query( $query ) {
...
-			return false;
+		if ( method_exists( $query, 'is_search' ) && ! $query->is_search() && empty( $query->query_vars['s'] ) ) {
+			/**
+			 * Filter to allow WooCommerce integration
+			 *
+			 * @hook ep_woocommerce_integration
+			 * @param  {bool} $allow True to allow ep integration
+			 * @param  {WP_Query} $query WP Query to evaluate
+			 * @return  {bool} New allow value
+			 */
```
felipeelia: Let's add a `@since` tag here :)

tests/php/features/TestWooCommerce.php
```diff
@@ -76,7 +76,8 @@ public function testProductsPostTypeQueryOn() {
 		add_action( 'ep_wp_query_search', array( $this, 'action_wp_query_search' ), 10, 0 );
 
 		$args = array(
-			'post_type' => 'product',
+			'post_type'    => 'product',
+			'ep_integrate' => true,
 		);
 
 		$query = new \WP_Query( $args );
```
felipeelia: It seems we should change this test entirely: we actually should make sure it will NOT be integrated with EP if it doesn't need it, right?

tests/php/features/TestWooCommerce.php
```diff
@@ -106,7 +107,8 @@ public function testProductsPostTypeQueryProductCatTax() {
 		add_action( 'ep_wp_query_search', array( $this, 'action_wp_query_search' ), 10, 0 );
 
 		$args = array(
-			'tax_query' => array(
+			'ep_integrate' => true,
+			'tax_query'    => array(
 				array(
 					'taxonomy' => 'product_cat',
 					'terms'    => array( 'cat' ),
```
felipeelia: Same here.

felipeelia (changes requested):

felipeelia:
@MARQAS before we merge this one we will need a list of real scenarios where we would expect the feature to integrate or not with Elasticsearch. Then that will need to become proper tests. Can you please work on that? Thanks.

oscarssanchez:
https://github.com/10up/ElasticPress/pull/2914/commits/f66b88011ac471fd8fa013025024fc641626b443 looks good but am not sure if the two instances of ep_integrate removed from tests is all we need. Did we create the list of scenarios @MARQAS ?

felipeelia:
@MARQAS I've created a new sheet in our doc with a format I think will make things easier for us to understand. Can you please let me know when you have that completely filled? Feel free to add any new lines/scenarios you think I missed.
After we have that complete, we can plan to write tests for each scenario. Thanks!

includes/classes/Feature/WooCommerce/WooCommerce.php
```diff
@@ -1198,45 +1208,33 @@ protected function should_integrate_with_query( $query ) {
...
+		if ( true === $allow ) {
+			return true;
 		}
-
-		/**
-		 * Cant hook into WC API yet
-		 */
-		if ( defined( 'WC_API_REQUEST' ) && WC_API_REQUEST ) {
-			return false;
+		if ( method_exists( $query, 'is_search' ) && ! $query->is_search() && empty( $query->query_vars['s'] ) ) {
```
felipeelia: [question] The `is_search()` method is available since WP 3.1.0. Is there any reason why we are checking if it exists?

includes/classes/Feature/WooCommerce/WooCommerce.php
```diff
@@ -1198,45 +1208,33 @@ protected function should_integrate_with_query( $query ) {
...
+			/**
+			 * Filter to allow WooCommerce integration
+			 *
+			 * @hook ep_woocommerce_integration
+			 * @param  {bool} $allow True to allow ep integration
+			 * @param  {WP_Query} $query WP Query to evaluate
+			 * @since  4.4.0
```
felipeelia: We will need to update this to 4.5.0

felipeelia (changes requested):

felipeelia:
Closing this in favor of (the already merged) #3259

Issue #2887, opened 2022-07-12
Some feedback and ideas around the dummy request in Autosuggest.php


kovshenin:
Hey folks! I was reviewing the code in Autosuggest.php recently and a few things stood out to me, which I'd like to share with you below. Please note that I do not have deep knowledge of the ElasticPress codebase so some of the points may be quite off, but either way I think a few of them will be useful.

1. The primary purpose of the `generate_search_query()` method seems to be able to obtain a JSON payload with a ploceholder, which can later be modified and sent to an ES endpoint via JavaScript. I understand the need for a dummy `new WP_Query` to be able to generate that JSON payload, but what I don't understand is the fact that it's being followed-through in `intercept_search_request` with an actual `wp_remote_request()`. We already have the `$this->autosuggest_query` set, so why are we doing a remote request, then caching that request (with the stripped response) and never using it anywhere?
2. A call to `wp_remote_request()` may result in a `WP_Error` object, and the caching techniques in `intercept_search_request()` may result in that error object being stored in Redis/Memcached indefinitely, because a `WP_Error` is not equal to `false`. Luckily though, it doesn't seem to be used anyway, as mentioned earlier.
3. In that same `intercept_search_request()` method, the cache key for transients is different from a cache key to be used with a persistent cache backend. Not sure why that's the case, nor why we have a dedicated `ep_autosuggest` group to potentially hold a single item for a fake dummy request that's not even needed.
4. There's some logic inside WordPress transients that already defaults to using object cache functions instead of the DB if a persistent backend is available, so the whole split on `wp_using_ext_object_cache()` is really unnecessary, but we do have to settle on a cache key.
5. The way persistent object cache is handled in `delete_cached_query()` is a bit questionable. Most persistent backends, even ones extending `WP_Object_Cache` will not do anything if their local `->$cache` is touched directly. There is no uniform way to delete an entire group unfortunately, as many backends do not even have a concept of a "group", however, if we use a single cache key, just like with transients, we will not have to delete any groups.
6. The `remove_filter( 'ep_intercept_remote_request', '__return_true' );` in `generate_search_query()` may remove an explicit `__return_true` set by something else (at the default priority) earlier. If we want to add a filter and remove it soon after, it's best to avoid such global callbacks. Perhaps an anonymous `$return_true` function, or a unique method, similar to `return_empty_posts` would work better.

I'll be happy to elaborate on any of these problems, so let me know if you have any questions or ideas.

felipeelia:
[...]

PR #2916, opened 2022-08-01
Use `wp_cache_flush_group()` when available

tillkruss:
### Description of the Change

Unsetting the group key in `$cache` is a shot in the dark, let's use WordPress 6.1 new `wp_cache_flush_group()` when available

Follow-up to #2915.

### Credits
Props @tillkruss

### Checklist:

### DIFF
--- a/includes/classes/Feature/Autosuggest/Autosuggest.php
+++ b/includes/classes/Feature/Autosuggest/Autosuggest.php
@@ -652,8 +652,13 @@ class Autosuggest extends Feature {
 	public function delete_cached_query() {
 		global $wp_object_cache;
 		if ( wp_using_ext_object_cache() ) {
-			// Delete the entire group.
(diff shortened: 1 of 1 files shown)


cmcandrew:
If this could be treated with some urgency, we have been unable to upgrade beyond ElasticPress 3.6.6

Since 4.x with Object Cache Pro plugin exhibits run-away Memory Usage whilst indexing

felipeelia:
Thanks for the PR @tillkruss! I'll merge it now.

For future reference, I'll leave here links to:
- [the ticket where the function was introduced](https://core.trac.wordpress.org/ticket/4476) (to be available in WordPress 6.1)
- [a comment with an explanation about the related code](https://github.com/10up/ElasticPress/issues/2887#issuecomment-1185651771)

tillkruss:
@felipeelia Sure thing, will do! 👍

PR #2917, opened 2022-08-02
Display last sync info in site health screen

### LINKED ISSUE #2856: Output info about the last sync for Dashboard users

In ElasticPress 4.2.0, we've added a new `get-last-sync` WP-CLI command (see #2748.) Dashboard users could also have access to that.

It can either be a new element in the Sync Page or a new entry in Health Tools' Info Screen

oscarssanchez:
### Description of the Change

This PR adds the sync info in the site health screen

Closes #2856

Closes #

### Changelog Entry

> Added - Display last sync info in site health screen

### Credits

Props @oscarssanchez , @felipeelia 

### Checklist:


includes/classes/SyncManager.php
```diff
@@ -267,6 +269,31 @@ public function action_delete_blog_from_index( $blog_id ) {
...
+	 * @param array $debug_info The debug info for site health screen.
+	 * @return array The debug info for site health screen.
+	 */
+	public function last_sync_health_info( $debug_info ) {
+		$debug_info['ep_last_sync'] = array(
+			'label'  => esc_html__( 'ElasticPress - Last Sync', 'elasticpress' ),
+			'fields' => [],
+		);
+
+		$sync_info = \ElasticPress\IndexHelper::factory()->get_last_index();
```
felipeelia: As we already are in the ElasticPress namespace, do we really need the `\ElasticPress\` part here, @oscarssanchez ?

felipeelia:
@oscarssanchez did you test this if the last sync threw some errors? The following snippet will likely throw errors related to the field limits, if you need:
```
add_filter(
	'ep_prepare_meta_data',
		}
		return $post_meta;
	},
	10,
	2
);
```

oscarssanchez:
> @oscarssanchez did you test this if the last sync threw some errors? The following snippet will likely throw errors related to the field limits, if you need:
> 
> ```
> add_filter(
> 	'ep_prepare_meta_data',
> 	function( $post_meta, $post ) {
> 		for ( $i = 0; $i < 100; $i++ ) {
> 			$post_meta[ "test_meta_{$i}_title_{$post->ID}" ] = 'Lorem';
> 			$post_meta[ "test_meta_{$i}_body_{$post->ID}" ]  = 'Ipsum';
> 		}
> 		return $post_meta;
> 	},
> 	10,
> 	2
> );
> ```

Yes errors seem to be logged correctly as well.

includes/classes/SyncManager.php
```diff
@@ -267,6 +269,31 @@ public function action_delete_blog_from_index( $blog_id ) {
...
+	public function last_sync_health_info( $debug_info ) {
+		$debug_info['ep_last_sync'] = array(
+			'label'  => esc_html__( 'ElasticPress - Last Sync', 'elasticpress' ),
+			'fields' => [],
+		);
+
+		$sync_info = IndexHelper::factory()->get_last_index();
+
+		foreach ( $sync_info as $label => $value ) {
+			$debug_info['ep_last_sync']['fields'][ sanitize_title( $label ) ] = [
```
felipeelia: @oscarssanchez is there any reason why we are sanitizing it here and not in the line below?

includes/classes/SyncManager.php
```diff
@@ -62,6 +62,8 @@ public function __construct( $indexable_slug ) {
 		add_action( 'archive_blog', array( $this, 'action_delete_blog_from_index' ) );
 		add_action( 'deactivate_blog', array( $this, 'action_delete_blog_from_index' ) );
 
+		add_action( 'debug_information', array( $this, 'last_sync_health_info' ) );
+
```
felipeelia: @oscarssanchez I am not sure if SyncManager.php would be the best file to place this code. Did you consider creating a new class under the includes/classes/Screen folder and calling it in Screen.php (like we have for Sync)? We could create a separate HealthInfo as well.

tests/cypress/integration/general.spec.js
```diff
@@ -105,4 +105,40 @@ describe('WordPress can perform standard ElasticPress actions', () => {
...
+			'#health-check-accordion-block-ep_last_sync table.widefat.striped.health-check-table tbody tr:nth-child(6) td',
+		).should('contain.text', 'errors');
+
+		cy.get(
+			'#health-check-accordion-block-ep_last_sync table.widefat.striped.health-check-table tbody tr:nth-child(7) td',
+		).should('contain.text', 'end_date_time');
+
+		cy.get(
+			'#health-check-accordion-block-ep_last_sync table.widefat.striped.health-check-table tbody tr:nth-child(8) td',
+		).should('contain.text', 'end_time_gmt');
```
felipeelia: As this is the same info we have in end_date_time, I think we can remove this one. One thing I'd like to see added to the `ep_last_index` is the method used to sync (wp-cli or dashboard, for example). Likely a change that needs to happen in IndexHelper::update_last_index.

felipeelia (changes requested):
Overall it looks good @oscarssanchez. Although it is very developer-ish information, I still think we could use some better labels there. `Total time` instead of `total_time`, for example.

Can we have a new array mapping those labels to actual terms? It would obviously default to the label itself when no real term is found.

PR #2919, opened 2022-08-02
Move taxonomy facets to a separate class

felipeelia:
### Description of the Change

This PR is a revamp of #2683, organizing things a bit more before making any change.

Related to https://github.com/10up/ElasticPress/issues/1440 (it does not close the issue because the range implementation should be done in another PR)

### Changelog Entry

Changed: Facets are now divided by types and received their own class.

### Credits

Props @felipeelia 

### Checklist:


tott (dismissed):
This looks good to me

PR #2923, opened 2022-08-04
Add E2E tests for WooCommerce

burhandodhy:
### Description of the Change
This PR adds new test cases for the WooCommerce feature.

### Credits

Props @burhandodhy 

### Checklist:


tests/cypress/integration/features/woocommerce.spec.js
```diff
@@ -85,4 +88,40 @@ describe('WooCommerce Feature', () => {
...
+	it('Can search orders from ElasticPress in WP Dashboard', () => {
+		cy.login();
+
+		cy.maybeEnableFeature('protected_content');
+		cy.maybeEnableFeature('woocommerce');
+
+		cy.visitAdminPage('edit.php?post_type=shop_order');
+
+		cy.get('#post-search-input').type('1988');
+		cy.get('#search-submit').click();
```
felipeelia: Could we create an order here? That would also help to check if newly created orders are properly indexed.

tests/cypress/integration/features/woocommerce.spec.js
```diff
@@ -85,4 +88,40 @@ describe('WooCommerce Feature', () => {
...
+
+	it('Can a user only see his orders on the My Account Order page', () => {
+		cy.login();
+
+		cy.maybeEnableFeature('protected_content');
+		cy.maybeEnableFeature('woocommerce');
+
+		cy.visit('my-account/orders');
+
+		cy.get('.woocommerce-orders-table tbody tr').should('have.length', 3);
```
felipeelia: I see this works but shouldn't we create orders with other users too? Not sure how we have that in the initial import.

tests/cypress/integration/features/woocommerce.spec.js
```diff
@@ -85,4 +88,40 @@ describe('WooCommerce Feature', () => {
...
+
+		cy.get('#debug-menu-target-EP_Debug_Bar_ElasticPress .ep-query-debug').should(
+			'contain.text',
+			'Query Response Code: HTTP 200',
+		);
+
+		cy.get('.order_number .order-view').should('contain.text', '#1988 Mozelle Davis');
+	});
+
+	it('Can a user only see his orders on the My Account Order page', () => {
```
felipeelia: can we have it as `can not display other users' orders on the My Account Order page`?

felipeelia (changes requested):

tests/cypress/integration/features/woocommerce.spec.js
```diff
@@ -88,4 +103,80 @@ describe('WooCommerce Feature', () => {
...
+
+		// ensure order is visible to user.
+		cy.visit('my-account/orders');
+		cy.get('.woocommerce-orders-table tbody tr').should('have.length', 1);
+
+		// Test orderby parameter set to `date` in query.
+		cy.get('#debug-menu-target-EP_Debug_Bar_ElasticPress .ep-query-debug').should(
+			'contain.text',
+			"'orderby' => 'date'",
+		);
```
felipeelia: As the name of the test is `Can not display other users orders on the My Account Order page` shouldn't we log in using a different user and check (1) that EP is used and (2) the order is not displayed?

tests/cypress/integration/features/woocommerce.spec.js
```diff
@@ -88,4 +103,80 @@ describe('WooCommerce Feature', () => {
...
+
+	it('Can search orders from ElasticPress in WP Dashboard', () => {
+		cy.login();
+
+		cy.maybeEnableFeature('protected_content');
+		cy.maybeEnableFeature('woocommerce');
+
+		cy.visitAdminPage('edit.php?post_type=shop_order');
+
+		cy.get('#post-search-input').type(`${userData.firstName} ${userData.lastName}{enter}`);
```
felipeelia: We could expand this test a bit. Instead of simply searching for the user name, let's also search for order IDs, billing and shipping info, and items: https://github.com/10up/ElasticPress/blob/develop/includes/classes/Feature/WooCommerce/WooCommerce.php#L411

felipeelia (changes requested):

tests/cypress/integration/features/woocommerce.spec.js
burhandodhy: Done. The reason it was not done before was that we had some dummy orders that were linked with the admin user.

tests/cypress/integration/features/woocommerce.spec.js
burhandodhy: Done

tests/cypress/integration/features/woocommerce.spec.js
```diff
@@ -88,4 +102,121 @@ describe('WooCommerce Feature', () => {
...
+
+		// ensure order is visible to user.
+		cy.visit('my-account/orders');
+		cy.get('.woocommerce-orders-table tbody tr').should('have.length', 1);
+
+		// Test orderby parameter set to `date` in query.
+		cy.get('#debug-menu-target-EP_Debug_Bar_ElasticPress .ep-query-debug').should(
+			'contain.text',
+			"'orderby' => 'date'",
+		);
```
felipeelia: Just to be sure we are not capturing an autosuggest template query or similar, can we check for `shop_order` in this query?

tests/cypress/integration/features/woocommerce.spec.js
```diff
@@ -88,4 +102,121 @@ describe('WooCommerce Feature', () => {
...
+		});
+
+		// ensure no order is show.
+		cy.visit('my-account/orders');
+		cy.get('.woocommerce-orders-table tbody tr').should('have.length', 0);
+
+		cy.get('#debug-menu-target-EP_Debug_Bar_ElasticPress .ep-query-debug').should(
+			'contain.text',
+			'Query Response Code: HTTP 200',
+		);
```
felipeelia: same, can we check for `shop_order` on this one as well?

tests/cypress/integration/features/woocommerce.spec.js
```diff
@@ -88,4 +102,121 @@ describe('WooCommerce Feature', () => {
...
+
+		cy.get('.order_number .order-view').should(
+			'contain.text',
+			`${userData.firstName} ${userData.lastName}`,
+		);
+	});
 });
```
felipeelia, line 249: @burhandodhy can we bring to this file the test we have for Product variations skus?

felipeelia (changes requested):

PR #2926, opened 2022-08-04
Improve wpcli commands test coverage

burhandodhy:
### Description of the Change

This PR adds more test for the wpcli commands.

### Credits

Props @burhandodhy 

### Checklist:


tests/cypress/integration/wp-cli.spec.js
```diff
@@ -114,6 +114,15 @@ describe('WP-CLI Commands', () => {
...
 			});
+
+			it('Can index without using dynamic bulk requests if user specifies --static-bulk parameter', () => {
+				cy.wpCli('wp elasticpress index --static-bulk')
+					.its('stdout')
+					.should('contain', 'Indexing posts')
+					.should('contain', 'Number of posts indexed')
+					.should('contain', 'Sync complete')
+					.should('contain', 'Done');
+			});
```
felipeelia: @burhandodhy I may be wrong but I don't think this is testing anything different than the regular index. Ultimately `--static-bulk` will prevent EP to try to sync sending several requests per batch of ES documents, i.e., if per-page=500:
- Without `--static-bulk` EP will fetch 500 posts from the database, split that number of posts into small groups and send different requests
- With it, EP will fetch 500 posts from the DB and send ALL of them at once.

One way to test would be checking if the `ep_after_bulk_index` action OR the `ep_after_bulk_index_dynamically` actions were fired. Another one could be checking the number of requests fired.

tests/cypress/integration/wp-cli.spec.js
```diff
@@ -291,4 +300,72 @@ describe('WP-CLI Commands', () => {
...
+	});
+
+	it('Can get the cluster indexes information', () => {
+		cy.wpCli('wp elasticpress get-cluster-indexes').its('stdout').should('contain', 'health');
+	});
+
+	it('Can get the indexes names', () => {
+		cy.wpCli('wp elasticpress get-indexes').its('code').should('equal', 0);
+
+		cy.wpCli('wp elasticpress get-indexes --pretty').its('code').should('equal', 0);
```
felipeelia: Any chance we check for new lines and tabs here? It would be the main difference generated by the `--pretty` flag.

tests/cypress/integration/wp-cli.spec.js
```diff
@@ -291,4 +300,72 @@ describe('WP-CLI Commands', () => {
...
+	});
+
+	it('can clear a sync/index process', () => {
+		cy.wpCli('wp elasticpress clear-index').its('stdout').should('contain', 'Index cleared');
+	});
+
+	it('Can stop the indexing operation', () => {
```
felipeelia: @burhandodhy we probably could mix these two tests in just one, let's say `it('Can stop the sync operation and clear it')`, and then move the clear-index call to the bottom of the second test, replacing the `delete_option( 'ep_index_meta' ); delete_transient( 'ep_sync_interrupted');` call.

felipeelia (changes requested):

tests/cypress/integration/wp-cli.spec.js
burhandodhy: Done

tests/cypress/integration/wp-cli.spec.js
burhandodhy: Done

tests/cypress/integration/wp-cli.spec.js
burhandodhy: Done. Created a test plugin and used `ep_after_bulk_index` action to log the message

tests/cypress/integration/wp-cli.spec.js
```diff
@@ -115,6 +115,15 @@ describe('WP-CLI Commands', () => {
 				expect(indexPerPostType).to.not.equal(indexTotal);
 			});
 		});
+
+		it('Can index without using dynamic bulk requests if user specifies --static-bulk parameter', () => {
+			cy.activatePlugin('fake-log-messages');
```
felipeelia, line 120: do you mind deactivating the plugin after the test @burhandodhy ?

felipeelia (approved):

PR #2927, opened 2022-08-04
Log error messages in the dashboard

### LINKED ISSUE #2922: BUG: Wrong error message
**Describe the bug**
I have many posts to sync to ElasticSearch (500,000). I tried to sync using the plugin admin page.

After reaching 99% progress, I received an error:

`Sync Failed.  The response is not a valid JSON response.`

I tried again using the CLI, and it looked as if it was working fine, but when I checked the plugin admin page, I noticed it was showing sync failed even when the CLI process was working fine.

Then I decided to check the database for `ep_last_index` and `ep_last_cli_index`, and here I noticed this error:

```
Maximum execution time of 300 seconds exceeded
```

Which I believe is related to my PHP execution limit.

It would have been great if this error was showing on the plugin admin interface, as it would save several hours of debugging.

**Steps to Reproduce**
Have a large number of posts where syncing will take more than the PHP execution limit.
1. Go to 'ElasticPress -> Sync'
2. Click on 'Sync Now'
3. Wait for `x` seconds (based on your PHP Execution limit variable.
4. See sync failed without showing the error.

**Expected behavior**
Show the error listed in `ep_last_index` and `ep_last_cli_index` options.

felipeelia:
### Description of the Change

After #2748, fatal error messages are correctly displayed in WP-CLI. This PR makes an adjustment so it is also displayed in the Sync Dashboard.

Closes #2922

### How to test the Change

In an installation with a big number of posts, set `max_execution_time = 1` in an ini file and run the sync via the dashboard.

### Credits

Props @felipeelia @orasik

### DIFF
--- a/assets/js/sync/index.js
+++ b/assets/js/sync/index.js
@@ -124,14 +124,14 @@ const App = () => {
 			}
 
 			/**
-			 * Log any messages.
+			 * Log any error messages created by the browser.
 			 */
 			if (error.message) {
 				logMessage(error.message, 'error');
 			}
 
 			/**
-			 * Log any messages.
+			 * Log any error messages created by the back-end.
 			 */
 			if (error.data?.message) {
 				logMessage(error.data.message, 'error');
--- a/assets/js/sync/index.js
+++ b/assets/js/sync/index.js
@@ -130,6 +130,13 @@ const App = () => {
 				logMessage(error.message, 'error');
 			}
 
+			/**
+			 * Log any messages.
+			 */
+			if (error.data?.message) {
+				logMessage(error.data.message, 'error');
+			}
+
 			logMessage(__('Sync failed', 'elasticpress'), 'error');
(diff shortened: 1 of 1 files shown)


assets/js/sync/index.js
```diff
@@ -130,6 +130,13 @@ const App = () => {
 				logMessage(error.message, 'error');
 			}
 
+			/**
+			 * Log any messages.
```
JakePT: The comment could probably explain that this message will exist if the failed request includes a response with error data, as the other stuff in the function is handling errors created by the browser.

JakePT (changes requested):

assets/js/sync/index.js
felipeelia: Good catch @JakePT, thanks. I've made a change in both comments to explain the origins of each type of error. Can you please give it another look?

JakePT (approved):

PR #2931, opened 2022-08-08
Add E2E tests for terms feature

burhandodhy:
### Description of the Change

This PR adds the E2E tests for the terms feature.

### Credits

Props @burhandodhy 

### Checklist:


tests/cypress/integration/features/terms.spec.js
```diff
@@ -0,0 +1,127 @@
...
+		});
+
+		cy.get('.ep-sync-progress strong', {
+			timeout: Cypress.config('elasticPressIndexTimeout'),
+		}).should('contain.text', 'Sync complete');
+
+		cy.wpCli('wp elasticpress list-features').its('stdout').should('contain', 'terms');
+	});
+
+	it('Can searching a term in the admin dashboard use Elasticsearch', () => {
```
felipeelia: @burhandodhy can we rephrase it to `Can search a term in the admin dashboard using Elasticsearch`?

tests/cypress/integration/features/terms.spec.js
```diff
@@ -0,0 +1,127 @@
...
+
+		const term = 'amazing';
+
+		// Create a new term
+		cy.createTaxonomy({ name: term });
+
+		// Search for the term
+		cy.get('#tag-search-input').type(term);
+		cy.get('#search-submit').click();
+		cy.get('.wp-list-table tbody tr').should('have.length', 1).should('contain.text', term);
```
felipeelia, line 76: @burhandodhy can we check if the results came from Elasticsearch?

tests/cypress/integration/features/terms.spec.js
```diff
@@ -0,0 +1,127 @@
...
+
+		// Delete the term
+		cy.get('.wp-list-table tbody tr')
+			.first()
+			.find('.row-actions .delete a')
+			.click({ force: true });
+
+		// Re-search for the term and make sure it's not there.
+		cy.get('#search-submit').click();
+		cy.get('.wp-list-table tbody').should('contain.text', 'No categories found');
```
felipeelia, line 91: @burhandodhy can we check if the results came from Elasticsearch?

tests/cypress/integration/features/terms.spec.js
```diff
@@ -0,0 +1,127 @@
...
+		cy.get('#search-submit').click();
+
+		cy.get('.wp-list-table tbody tr')
+			.should('have.length', 1)
+			.should('contain.text', searchTerm);
+
+		cy.get('#debug-menu-target-EP_Debug_Bar_ElasticPress .ep-query-debug').should(
+			'contain.text',
+			'Query Response Code: HTTP 200',
+		);
```
felipeelia: @burhandodhy How could we make sure this query is really the one we are looking for?

tests/cypress/integration/features/terms.spec.js
```diff
@@ -0,0 +1,127 @@
...
+		cy.wpCli('wp elasticpress list-features').its('stdout').should('contain', 'terms');
+	});
+
+	it('Can searching a term in the admin dashboard use Elasticsearch', () => {
+		cy.login();
+		cy.maybeEnableFeature('terms');
+
+		cy.visitAdminPage('edit-tags.php?taxonomy=category');
+
+		const searchTerm = 'classic';
```
felipeelia: @burhandodhy I think this comes from the default content being imported at the beginning of the tests, is that right? If so, can we create a term during the test? The initial import slows things down quite a bit and I'm planning to remove that eventually.

tests/cypress/integration/features/terms.spec.js
```diff
@@ -0,0 +1,127 @@
...
+			cy.get('#tag-search-input').clear().type(tag);
+			cy.get('#search-submit')
+				.click()
+				.then(() => {
+					cy.get('.wp-list-table tbody tr')
+						.first()
+						.find('.row-actions .delete a')
+						.click({ force: true });
+				});
+		});
```
felipeelia: @burhandodhy although it sometimes can be counterintuitive, we should actually clean things created by tests *before* the tests: https://docs.cypress.io/guides/references/best-practices#Using-after-or-afterEach-hooks (the docs mention the hooks, but the idea remains the same.)

We could use something like we are using for users here: https://github.com/10up/ElasticPress/blob/develop/tests/cypress/integration/indexables/user.spec.js#L9

Thoughts?

tests/cypress/integration/features/terms.spec.js
```diff
@@ -0,0 +1,127 @@
...
+		// create tags.
+		tags.forEach((tag) => {
+			cy.createTaxonomy({ name: tag, taxonomy: 'post_tag' });
+		});
+
+		// search for the tag.
+		cy.get('#tag-search-input').type('the most fun thing');
+		cy.get('#search-submit').click();
+
+		cy.get('.wp-list-table tbody tr .row-title').should('contain.text', 'The Most Fun Thing');
```
felipeelia, line 113: @burhandodhy can we check if the results came from Elasticsearch?

tests/cypress/integration/features/terms.spec.js
```diff
@@ -0,0 +1,127 @@
...
+				.then(() => {
+					cy.get('.wp-list-table tbody tr')
+						.first()
+						.find('.row-actions .delete a')
+						.click({ force: true });
+				});
+		});
+	});
+
+	it('Can a child term be updated when a parent term is deleted', () => {
```
felipeelia: @burhandodhy can we rephrase this to something like `can update a child term when a parent term is deleted`? The intent is to come up with a phrase where `it` is the app doing something.

tests/cypress/support/commands.js
```diff
@@ -57,8 +57,20 @@ Cypress.Commands.add('openWidgetsPage', () => {
 	});
 });
 
-Cypress.Commands.add('createTaxonomy', (name = 'Test taxonomy', taxonomy = 'category') => {
+Cypress.Commands.add('createTaxonomy', (data) => {
```
felipeelia: @burhandodhy great improvement here, thanks!

felipeelia (changes requested):
Overall this looks great @burhandodhy! I left some comments and requested some changes but please let me know if you have any concerns or a different approach in mind.

tests/cypress/integration/features/terms.spec.js
burhandodhy: @felipeelia as we are planning to get the rid of all the dummy data. Is it okay to delete all the terms by using the following command `wp term list post_tag --field=term_id | xargs wp term delete post_tag` in the before hook?

tests/cypress/integration/features/terms.spec.js
burhandodhy: Done

tests/cypress/integration/features/terms.spec.js
burhandodhy: Done

tests/cypress/integration/features/terms.spec.js
burhandodhy: Done

tests/cypress/integration/features/terms.spec.js
burhandodhy: Valid point. I changed the test and now we are looking for the search term in the Elasticsearch result.

tests/cypress/integration/features/terms.spec.js
burhandodhy: Done. Created new term.

tests/cypress/integration/features/terms.spec.js
burhandodhy: Done

tests/cypress/integration/features/terms.spec.js
burhandodhy: Done

tests/cypress/integration/features/terms.spec.js
burhandodhy: Thanks for suggestion. I moved it to after hook

tests/cypress/integration/features/terms.spec.js
```diff
@@ -0,0 +1,152 @@
...
+		cy.get('.ep-feature-terms .settings-button').click();
+		cy.get('.ep-feature-terms [name="settings[active]"][value="1"]').click();
+		cy.get('.ep-feature-terms .button-primary').click();
+		cy.on('window:confirm', () => {
+			return true;
+		});
+
+		cy.get('.ep-sync-progress strong', {
+			timeout: Cypress.config('elasticPressIndexTimeout'),
+		}).should('contain.text', 'Sync complete');
```
felipeelia: @burhandodhy let's also check if the mapping was sent here, please. You can see how we are doing it for the WooCommerce feature [here](https://github.com/10up/ElasticPress/blob/develop/tests/cypress/integration/features/woocommerce.spec.js#L32)

tests/cypress/integration/features/terms.spec.js
```diff
@@ -0,0 +1,152 @@
...
+			cy.get('#search-submit')
+				.click()
+				.then(() => {
+					cy.get('.wp-list-table tbody tr')
+						.first()
+						.find('.row-actions .delete a')
+						.click({ force: true });
+				});
+		});
+	});
```
felipeelia: @burhandodhy Can we move that to the before hook and use wp-cli to clean it up? With the change made on #2944 we should be able to call nested wp-cli commands now.

tests/cypress/support/commands.js
felipeelia: @burhandodhy actually this was wrongly named before. Do you mind updating it to `createTerm`?

felipeelia (changes requested):

tests/cypress/integration/features/terms.spec.js
burhandodhy: Done.

tests/cypress/integration/features/terms.spec.js
burhandodhy: Done

felipeelia (approved):

PR #2942, opened 2022-08-12
Swap quotation marks to be compatible with Kibana

nickchomey:
### Description of the Change

When the highlight tags get, presumably, converted into JSON, it results in a double set of double quotation marks, such as this
`{"pre_tags":["<mark class="ep-highlight">"],`

This PR changes it so that it generates single quotes wrapped in double quotes
`{"pre_tags":["<mark class='ep-highlight'>"],`

image

image

Closes #

### How to test the Change

Generate a query that uses highlighting. Copy it into Kibana and try to run it.

### Changelog Entry

> Fixed - Bug fix - Quotation mark JSON formatting fix

### Credits

Props @nickchomey

### Checklist:

- [ x] I agree to follow this project's [**Code of Conduct**](https://github.com/10up/.github/blob/trunk/CODE_OF_CONDUCT.md).
- [ x] I have updated the documentation accordingly.
- [ x] I have added tests to cover my change.
- [ x] All new and existing tests pass.


burhandodhy (dismissed):
LGTM

felipeelia:
Hey @nickchomey, can you please address the problems found by the linter? Also, where are you copying that query from? Thanks!

nickchomey:
@felipeelia I believe I've addressed the issue - it was silly of me to change more quotation marks than was necessary. However, one of the tests seems to have failed on many different things - all of which seem completely irrelevant to this PR... Is there perhaps something broken with the test mechanism?

And I believe I was copying the query from either the Debug Bar tool or directly from the code via Xdebug step debugging - they should be the same thing. The query wasn't working so I wanted to test and debug it from Kibana, which then provided more details about the issue of the quotation marks (though I'd imagine that the response shown by the Debug Bar tool contains the same info).

PR #2954, opened 2022-08-17
Facet by Meta

felipeelia:
### Description of the Change
This PR is the following step to #2919, implementing a new type for meta fields.

It still needs tests.

### Changelog Entry

> Added - New block to facet by meta fields

### Credits

Props @felipeelia

### DIFF
--- a/includes/classes/Feature/Facets/Types/Meta/Block.php
+++ b/includes/classes/Feature/Facets/Types/Meta/Block.php
@@ -37,19 +37,31 @@ class Block {
 	public function setup_endpoints() {
 		register_rest_route(
 			'elasticpress/v1',
-			'facets/block-meta-preview',
+			'facets/meta/keys',
+			[
+				'methods'             => 'GET',
+				'permission_callback' => [ $this, 'check_facets_meta_rest_permission' ],
+				'callback'            => [ $this, 'get_rest_registered_metakeys' ],
+			]
+		);
+		register_rest_route(
+			'elasticpress/v1',
+			'facets/meta/block-preview',
 			[
 				'methods'             => 'GET',
 				'permission_callback' => [ $this, 'check_facets_meta_rest_permission' ],
 				'callback'            => [ $this, 'render_block_preview' ],
 				'args'                => [
-					'facet'   => [
+					'searchPlaceholder' => [
 						'sanitize_callback' => 'sanitize_text_field',
 					],
-					'orderby' => [
+					'facet'             => [
 						'sanitize_callback' => 'sanitize_text_field',
 					],
-					'order'   => [
+					'orderby'           => [
+						'sanitize_callback' => 'sanitize_text_field',
+					],
+					'order'             => [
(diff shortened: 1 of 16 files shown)


includes/classes/Feature/Facets/Types/Meta/Block.php
```diff
@@ -0,0 +1,213 @@
...
+	/**
+	 * Outputs the block preview
+	 *
+	 * @param \WP_REST_Request $request REST request
+	 * @return string
+	 */
+	public function render_block_preview( $request ) {
+		global $wp_query;
+
+		add_filter( 'ep_is_facetable', '__return_true' );
```
tott, line 140: would it make sense to provide more context in the filter to be able to make a decision?

includes/classes/Feature/Facets/Types/Meta/FacetType.php
```diff
@@ -0,0 +1,322 @@
...
+		 * @return {null|array} Distinct meta values array or `null` to keep default behavior.
+		 */
+		$custom_meta_values = apply_filters( 'ep_facet_meta_custom_meta_values', null, $meta_key );
+		if ( null !== $custom_meta_values ) {
+			return $custom_meta_values;
+		}
+
+		$meta_values = get_transient( self::TRANSIENT_PREFIX . $meta_key );
+		if ( ! $meta_values ) {
+			$meta_values = \ElasticPress\Indexables::factory()->get( 'post' )->get_all_distinct_values( "meta.{$meta_key}.raw" );
```
tott: wondering if we need to truncate those values. there could be huge data in meta fields eventually. maybe truncate out only first 100 chars or so?

tott (commented):
Just some minor comments. One thing i'm wondering though - How will things work when people have multiple facet blocks of the same type. I did not see any variation in filters based on the block id or the like. Have you tested this?

includes/classes/Feature/Facets/Types/Meta/FacetType.php
felipeelia: As I can see some usefulness in bringing more data in other scenarios, in 2626fb4 I've added the number as a parameter.

includes/classes/Feature/Facets/Types/Meta/Block.php
felipeelia: Addressed in f9ae942. I will apply a similar change to #2919 as well, especially for the renderer class part.

felipeelia:
With the change made in 7c1b813, this is the code needed to make Facet by Meta blocks available:

```
add_filter(
	'ep_facet_types',
	function ( $facet_types ) {
		if ( ! isset( $facet_types['meta'] ) && class_exists( '\ElasticPress\Feature\Facets\Types\Meta\FacetType' ) ) {
			$facet_types['meta'] = '\ElasticPress\Feature\Facets\Types\Meta\FacetType';
		}
		return $facet_types;
	}
);
```

PR #2955, opened 2022-08-19
E2E tests for Comments Feature

burhandodhy:
### Description of the Change

Add E2E tests for Comments Feature

### Credits

Props @burhandodhy 

### Checklist:


tests/cypress/integration/features/comments.spec.js
```diff
@@ -0,0 +1,212 @@
...
+		cy.visitAdminPage('admin.php?page=elasticpress-sync');
+
+		// start sync and test results.
+		cy.get('.ep-sync-button--sync').click();
+		cy.get('.ep-sync-panel').first().as('syncPanel');
+		cy.get('@syncPanel').find('.components-form-toggle').click();
+		cy.get('@syncPanel')
+			.find('.ep-sync-messages', { timeout: Cypress.config('elasticPressIndexTimeout') })
+			.should('contain.text', `Number of comments indexed: ${defaultApprovedComments}`);
+	});
```
felipeelia: Any reason why we are not testing that in the previous test?

tests/cypress/integration/features/comments.spec.js
```diff
@@ -0,0 +1,212 @@
...
+				timeout: Cypress.config('elasticPressIndexTimeout'),
+			})
+			.should('contain.text', `Number of comments indexed: ${defaultApprovedComments}`);
+
+		// trash the comment
+		cy.visitAdminPage('edit-comments.php?comment_status=moderated');
+		cy.get('.column-comment .trash a').first().click({ force: true });
+	});
+
+	it('Can sync comments if approved manually', () => {
```
felipeelia: Is there any reason why we don't simply approve the comment created in the previous test?

felipeelia (changes requested):
@burhandodhy I left some comments, feel free to reach out via slack if I missed something obvious here. We will need to test this feature with Protected Content enabled as well (those tests will likely live in the `protected-content.spec.js` file. Are you planning to tackle that in a separate PR?

tests/cypress/integration/features/comments.spec.js
burhandodhy: The main idea is to keep each test independent.

tests/cypress/integration/features/comments.spec.js
burhandodhy: We can do it, but I think keeping the separate test makes more sense

felipeelia: Actually that does make sense for unit tests, for e2e tests we need to think about performance as well. Cypress's Best Practices guide has a really good text about that: https://docs.cypress.io/guides/references/best-practices#Creating-tiny-tests-with-a-single-assertion
Basically, it would indeed make sense to keep separate tests if you were testing something after deleting the comment. If you are deleting that to create a new one with the same info, we could reuse that, right?

tests/cypress/integration/features/comments.spec.js
felipeelia: same as above

tests/cypress/integration/features/comments.spec.js
burhandodhy: Thank you for pointing me in the right direction. I updated the tests.

PR #2958, opened 2022-08-23
Hide taxonomies from facet block whose show_ui is set to false.

### LINKED ISSUE #2957: Only show those taxonomies in Facet blocks for which show_ui & public parameter are set to true.

Most of the time, developers create a taxonomy that is not available in the admin dashboard, for example the `ep_custom_result ` [taxonomy](https://github.com/10up/ElasticPress/blob/develop/includes/classes/Feature/SearchOrdering/SearchOrdering.php#L338) in ElasticPress. Facet Blocks list the name all taxonomies in the select option. In my opinion, we should only list those taxonomies whose public and show_ui parameters are set to true.
image

Only show those taxonomies for which `show_ui` & `public` both are set to true.

burhandodhy:
### Description of the Change

This PR hides the taxonomies from facet block whose `show_ui` is set to false.

Closes #2957

### How to test the Change

- Go to Widgets screens
- Add Facet Block 
- Make sure Taxonomy field does not have `Custom Result` option

### Changelog Entry

> Changed - Hide taxonomies from facet block whose show_ui is set to false.

### Credits

Props @burhandodhy 

### Checklist:


felipeelia:
@burhandodhy do you mind fixing the conflict on this one so we can merge it? Thanks!

burhandodhy:
> @burhandodhy do you mind fixing the conflict on this one so we can merge it? Thanks!

Done

PR #2959, opened 2022-08-23
Support customizing the template for search results in Instant Results

### LINKED ISSUE #2527: Support customizing the template for search results in Instant Results
Currently the output of Instant Results search results is not customizable beyond CSS. Ideally it would be possible for a developer to customize or replace the template to arrange things as desired, or add additional data.

Some way to replace the template. Possibilities include applying a [JavaScript filter](https://developer.wordpress.org/block-editor/reference-guides/packages/packages-hooks/) to the search results component, or some way to replace the component with a file in the theme. I have not tested how well either would work with React components.

JakePT:
### Description of the Change
This implements a JavaScript hook using `@wordpress/hooks` to support developers replacing the component used to render search results in Instant Results.

In #2527 I described an approach where the filter wraps the return value to filter the rendered element upon render, but in this PR I ended up taking a different approach where I am filtering the component itself. I did this for a few reasons:

1. The filter only runs once, when the component is defined.
2. Conditionally replacing the component per-result is still possible via a wrapper component approach, which I've documented.
3. I think it's easier to understand what needs to be returned from any filter callbacks, as a React component is a well known concept, whereas filtering just the rendered result could cause confusion and give the incorrect impression that raw HTML could be returned to the filter or that other rendering methods would be supported.

Something that was also considered was the naming of JavaScript hooks. @felipeelia and I explored a few other plugins and Gutenberg itself and didn't find much consistency. What this PR proposes is `plugin.Feature.valueBeingFiltered` In this case that's `elasticpress.InstantResults.Result`.

The `Result` value is capitalized to reflect the fact that a React component is being filtered and those are conventionally named, and represented in JSX, capitalized. If the value of a regular function such as `getPostTypesFromForm()` were to be filtered that would use `elasticpress.InstantResults.getPostTypesFromForm`.

Some other things to think about are whether the namespace should be `elasticpress` or `ep`, to be somewhat consistent with the PHP hooks, and how to capitalize the feature names since they don't represent an actual object in JavaScript. I went with `InstantResults` over `instantResults` to match the PHP class and for aesthetic reasons.

Closes #2527

### How to test the Change
The included documentation update includes some examples of how the filter can be used. Implementing these would confirm if things are functioning as expected.

### Credits
Props @JakePT 

### Checklist:


assets/js/instant-results/components/common/result.js
```diff
@@ -0,0 +1,88 @@
...
+		</article>
+	);
+};
+
+/* eslint-disable jsdoc/check-tag-names */
+
+/**
+ * Filter the Result component.
+ *
+ * @hook elasticpress.InstantResults.Result
```
felipeelia: @JakePT Instead of disabling the rule, I'd prefer setting it as WooCommerce is doing [here](https://github.com/woocommerce/woocommerce/blob/8952dac7c02a113e80b97d5e57d0eb184796f22a/packages/js/eslint-plugin/configs/recommended.js#L32). One interesting thing on that repo is how they are using `@filter` instead of `@hook` [here](https://github.com/woocommerce/woocommerce/blob/feba6a8dcd55d4f5c7edc05478369c76df082293/plugins/woocommerce-admin/client/analytics/report/revenue/config.js#L20), for example.

I know the decision on this will probably be influenced by the new doc generator we are going to use but I'm curious to hear your thoughts on which would be the best one.

FWIW, this is how I had our `.eslintrs.js` so I didn't need the `eslint-disable jsdoc/check-tag-names` line:
```
const defaultEslintrc = require('10up-toolkit/config/.eslintrc');

module.exports = {
	...defaultEslintrc,
	rules: {
		'jsdoc/check-tag-names': [
			'error',
			{
				definedTags: ['jest-environment', 'filter', 'action', 'slotFill', 'scope'],
			},
		],
	},
};
```

docs/theme-integration.md
```diff
@@ -148,3 +148,70 @@ window.epAutosuggestQueryFilter = (query, searchText, input) => {
...
+const CustomResult = ({ date, title, url }) => {
+	return (
+		<div>
+			<strong><a href={url}>{title}</a></strong> {date}
+		</div>
+	)
+};
+
+wp.hooks.addFilter('elasticpress.InstantResults.Result', 'customResult', () => CustomResult);
+```
```
felipeelia: @JakePT could we have at least this simpler example in "plain" JS as well? WordPress core documentation makes it available on both ways in some places ([here](https://developer.wordpress.org/block-editor/reference-guides/filters/block-filters/#editor-blockedit), for example) and that might lower the barrier for new developers.

Also, can we make that first line ` ```js ` so GitHub understands its JavaScript? (I imagined GH would detect that automatically but it doesn't seem to be the case.)

felipeelia (changes requested):
@JakePT in addition to my comments, we also need to fix a conflict in theme-integration.md. Do you mind handling that? Thanks.

ps.: It would be _great_ if we could have e2e tests checking the new filter as well. Would you be able to add it?

assets/js/instant-results/components/common/result.js
JakePT: @felipeelia I've:

- Updated `@hook` to `@filter`.
- Updated .eslintrc to accept `@filter`.
- Updated `@since` version.
- Resolved merge conflict in `theme-integration.md`.

Still to do:

- Add E2E tests (it might be good to get #2851 done first).

docs/theme-integration.md
JakePT: @felipeelia I have added `js` to the code blocks but still need to add non-JSX examples.

JakePT: @felipeelia I've added non-JSX documentation.

assets/js/instant-results/components/common/result.js
JakePT: @JakePT I have added an E2E test for the filter.

PR #2983, opened 2022-09-05
Migrate Autosuggest filters to proper WordPress hooks

### LINKED ISSUE #2961: Migrate Autosuggest filters to proper WordPress hooks
ElasticPress Autosuggest supports modifying its functionality by defining JavaScript functions on the window with a particular name. This started with the `epDataFilter` function that allows modifying the data returned from Elasticserach for use by Autosuggest and functions as a rudimentary hook.

```
if (typeof window.epDataFilter !== 'undefined') {
	return window.epDataFilter(data, searchTerm);
}
```

Additional 'filters' have been added since, and they use the same approach for consistency. The problems are that this is now inconsistent with how modern WordPress supports filtering JavaScript and it only supports a single callback (at least without workarounds like copying existing callbacks and calling them within a new callback).

Since we are now implementing [@wordpress/hooks](https://developer.wordpress.org/plugins/hooks/) to support filtering Instant Results functionality (#2589) this issue is to propose that we migrate Autosuggest filters to use `wp.hooks.applyFilters()` and `wp.hooks.doAction()` instead of these functions on the window object. This will allow multiple filters to be used, won't require that callbacks populate the global namespace, and the development experience will be consistent with JavaScript hooks provided by WordPress itself and other plugins and themes.

In the case of `epDataFilter` the above code would become something like:

```
return wp.hooks.applyFilters('elasticpress.Autosuggest.esSearchData', data, searchTerm);
```

For backwards compatibility we could separately add any existing window functions as callbacks the new hooks:

```
if (typeof window.epDataFilter !== 'undefined') {
	wp.hooks.addFilter('elasticpress.Autosuggest.esSearchData', window.epDataFilter);
}

### PR DESCRIPTION
**Pull request by JakePT**
### Description of the Change
Migrates the custom 'filter' functions in Autosuggest to use proper WordPress hooks. Existing uses of filter functions are applied as callbacks to the new hooks from a back-compat file.

Closes #2961

### How to test the Change
- Add the example code from the _existing_ documentation and confirm they still behave as expected.
- Replace the code with the example code from the updated documentation and confirm the result still matches what is described.

### Changelog Entry
- Added `ep.Autosuggest.itemHTML`, `ep.Autosuggest.listHTML`, `ep.Autosuggest.query`, and `ep.Autosuggest.element` JavaScript hooks to Autosuggest and migrated filter functions to hook callbacks for backwards compatibility.

### Credits

Props @JakePT

### Checklist:

### CODE REVIEW

File: assets/js/autosuggest/back-compat.js [hunk 1]
```diff
@@ -0,0 +1,40 @@
+/**
+ * WordPress dependencies.
+ */
+import { addFilter } from '@wordpress/hooks';
+
+if (typeof window.epDataFilter !== 'undefined') {
+	addFilter('ep.Autosuggest.data', 'ep/epDatafilter', window.epDatafilter);
```
JakePT, line 7: These callback names should probably reflect that they are for backwards compatibility.

assets/js/autosuggest/index.js
```diff
@@ -552,26 +580,31 @@ function init() {
 
 			let query = buildSearchQuery(searchText, placeholder, queryJSON);
```
JakePT, line 582: `buildSearchQuery` should probably just return a parsed value.

JakePT: buildSearchQuery is publicly exposed so changing the return value would break compatibility.

assets/js/autosuggest/index.js
```diff
@@ -195,13 +201,17 @@ async function esSearch(query, searchTerm) {
...
-		if (typeof window.epDataFilter !== 'undefined') {
-			return window.epDataFilter(data, searchTerm);
-		}
-
-		return data;
+		/**
+		 * Filter the Elasticsearch response data used for Autosuggest.
+		 *
+		 * @hook ep.Autosuggest.data
+		 * @since 4.4.0
```
felipeelia: @JakePT we can ship this one in 4.3.1, do you mind changing the `@since` tag? Also, with the comment I made in the Instant Result PR, we will probably have to tweak the `@hook` here as well (or at least remove the `jsdoc/check-tag-names` disable.)

felipeelia (changes requested):

assets/js/autosuggest/index.js
JakePT: @felipeelia Done.

Issue #1415, opened 2019-07-24
When using protected content the admin may show outdated information under some scenarios


dustinrue:
**Describe the bug**
With protected content enabled interactions in the admin can result in inconsistent information.

**Steps to Reproduce**

1. On a small site with minimal plugins
2. Create a post and save it
3. In the post listing, click on trash to put the item in the trash
4. Note that when the page refreshes the item is still shown in the listing
5. Refresh the listing, note that the item has now no longer in the list

**Expected behavior**
Clicking trash would result in the post no longer being listed

**Environment information**

 - Plugins and version: I tested using 2.8 through 3.1
 - Theme and version: default WP theme on 5.2.2
 - Other installed plugin(s) and version(s): no other plugins

**Cause and potential solution**
This issue occurs because the request to update the post in Elasticsearch doesn't wait for Elasticsearch to commit the change, that is Elasticsearch returns immediately. The change isn't reflected in search results until the next refresh interval has fired which defaults to 1 second. When the WordPress admin loads again it does a search against an old copy of the index and gets outdated results back.

The potential solution is to add `?refresh=wait_for` to the index request for the affected document(s). For performance I wouldn't recommend issuing a refresh command manually after indexing the document, only when making the index request. Manually issuing a refresh will just result in a doubled up refresh (because ES is already doing scheduled refreshes anyway) potentially causing more delay than necessary.

I also wouldn't use `wait_for` unless the command came from the admin page while users are making changes they would expect to be fresh on the next page load. 

It maybe useful to allow for disable waiting in the event ES is configured with a longer than usual refresh interval.

felipeelia:
Just tested this and `$path = apply_filters( 'ep_bulk_index_request_path', $index . '/_bulk?refresh=wait_for', $body, $type );` really fixes the problem.

felipeelia:
Let's test with more posts at once.

PR #2984, opened 2022-09-05
Fix: Intermittent tests failures

### LINKED ISSUE #2971: BUG: Intermittent tests failures
**Describe the bug**

It seems we have unit and e2e tests that are intermittently failing. Examples are:
- Unit Tests: related to pointers in custom search results
- E2e: Searching for product variation sku
- E2e: Terms (Can a term be removed from the admin dashboard after deleting it)

burhandodhy:
### Description of the Change

This PR fixes the following tests.

1) `testGetPointerData` => The WP-Query in `get_pointer_data_for_localize` doesn't preserver the posts id. Passing the `post__in` to orderby fix the issue. [Related Issue](https://wordpress.stackexchange.com/questions/108585/wp-query-orderby-post-in-remains-ineffective-in-the-loop)

2) The [wait](https://github.com/10up/ElasticPress/pull/2984/files#diff-9b32b6c1f3b12ef31dff7f49b2b490b0c2ede0424dbe2b303f8b2c6f6587af7aR90) is necessary. The problem is that test runs too quickly, and occasionally the category is not removed from the ES and because of this ES returns the deleted category. Related Issue https://github.com/10up/ElasticPress/issues/1415

Closes #2971

### Changelog Entry

> Fixed -  Intermittent tests failures

### Credits

Props @burhandodhy 

### Checklist:

### DIFF
--- a/tests/cypress/integration/features/terms.spec.js
+++ b/tests/cypress/integration/features/terms.spec.js
@@ -86,6 +86,9 @@ describe('Terms Feature', () => {
 			.find('.row-actions .delete a')
 			.click({ force: true });
 
+		// eslint-disable-next-line cypress/no-unnecessary-waiting
+		cy.wait(2000);
+
 		// Re-search for the term and make sure it's not there.
 		cy.get('#search-submit').click();
 		cy.get('.wp-list-table tbody').should('contain.text', 'No categories found');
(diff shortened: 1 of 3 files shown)


tests/cypress/integration/features/terms.spec.js
```diff
@@ -86,6 +86,9 @@ describe('Terms Feature', () => {
 			.find('.row-actions .delete a')
 			.click({ force: true });
 
+		// eslint-disable-next-line cypress/no-unnecessary-waiting
+		cy.wait(2000);
```
burhandodhy, line 90: This wait is necessary. The problem is that test runs too quickly, and occasionally the category is not removed from the ES and because of this ES returns the deleted category

PR #2985, opened 2022-09-05
fix: change labels to prevent confusion

alaa-alshamy:
### Description of the Change

As those stats information coming from Elasticsearch `_stats` endpoint and under `_all.total`, the labels should be Total not Avg to prevent misunderstanding like what happened with me. see the mentioned code bellow:

https://github.com/10up/ElasticPress/blob/develop/includes/classes/Stats.php#L215

Closes #

### How to test the Change

Just open the `Index Health` page

### Changelog Entry

> Fixed - Bug fix

### Credits

### Checklist:

### DIFF
--- a/assets/js/stats.js
+++ b/assets/js/stats.js
@@ -106,7 +106,7 @@ if (queriesTimeChart) {
 	new Chart(queriesTimeChart, {
 		type: 'pie',
 		data: {
-			labels: ['Avg indexing time in ms', 'Avg query time in ms'],
+			labels: ['Total indexing time in ms', 'Total query time in ms'],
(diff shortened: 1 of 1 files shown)


felipeelia:
Thanks for opening the PR @alaa-alshamy. Talking internally we've decided to remove the entire box but before we do that, do you have any valid use case where that time was valuable/useful for you? Thanks!

alaa-alshamy:
Thanks for your reply @felipeelia.
No I don't have but if there's a way to got the real avg time of the queries it will be awesome.

felipeelia:
Closing in favor of #3115.

Issue #2978, opened 2022-09-02
Support more search form fields in Instant Results


JakePT:
**Is your enhancement related to a problem? Please describe.**
Currently the Instant Results modal is opened by submitting a form containing a search input. This opens the modal and sets the search term in Instant Results. Additionally, if any fields with the name `post_type` are found, this will be used to apply a post type filter in Instant Results. 

The problem is that these are the only two fields supported, so if a form has a category dropdown or sorting option, any selections in those fields would not be reflected in Instant Results.

**Describe the solution you'd like**
At the moment we manually check for those two fields and set them as search arguments when the modal is opened. We could support any field that uses the name of our parameters by converting the form data to URL parameters and passing those through the existing `getArgsFromUrlParams()` function that's used for history navigation.

This would enable support for setting more search parameters from a search form, but Instant Results doesn't always use the same parameter names that pre-existing search forms might. For example, a default `wp_dropdown_categories()` will use the name `cat`, but Instant Results expects the parameter `tax-category` for filtering by category. 

So if we want to support Instant Results for existing search forms not built with Instant Results in mind, or even future changes to the search block as outlined on https://github.com/WordPress/gutenberg/pull/40585, we will need to map WordPress query vars to our parameter names.

**Additional context**
Related to #2507.

PR #2987, opened 2022-09-06
Publicly expose a method for opening Instant Results.

### LINKED ISSUE #2507: Allow more control over which forms trigger Instant Results
Currently, Instant Results is attached to any `input[name="s"]` elements on the page, and triggered whenever the form that contains this input is submitted. This targets the Search widget, block, the output of `get_search_form()` in most themes, and any form that submits a query that includes the standard WordPress search parameter (`s`).

Therefore it's not currently possible to _remove_ Instant Results from specific search form instances, or to bind it to other forms or even buttons.

A potential use case for this is that because Instant Results only has a single configuration for facets, the user may want to only enable it for Products search.

There are several options for addressing this:

- Implement a filter to allow developers control over the selector used to target forms and inputs.
- Implement a setting to allow any user to change the selector used, a la Autosuggest.
- Expose an API to allow developers to bind Instant Results to any element or event they wish.

Implementation of any solution is _slightly_ more complicated compare to Autosuggest, as the current behaviour requires a form _and_ an input, as Instant Results is displayed based on the submit event on the form, but the search term is pulled from the instance of the input in that form. 

If an API were exposed it would need the capability to not only open the Instant Results modal, but also to potentially populate the search term and other parameters.

JakePT:
### Description of the Change
Publicly expose a method for opening Instant Results, allowing third parties to open Instant Results manually.

I originally implemented this as a custom action hook, but I anticipate some refactoring for #2978 and didn't want to commit to an API just yet. For now this just makes `openModal()` publicly available.

Closes #2507

### How to test the Change
Calling the following code to a theme or plugin should cause Instant Results to open when the button is clicked:

```html
<button id="search-blocks" type="button">Search "Blocks"</button>

<script>
window.addEventListener('DOMContentLoaded', () => {
	document.getElementById('search-blocks').addEventListener('click', () => {
		window.epInstantResults.openModal({ search: "blocks" });
	});
});
</script>
```

### Changelog Entry
Added a method for developers to manually open Instant Results, `window.epInstantResults.openModal()`.

### Credits
Props @JakePT 

### Checklist:


felipeelia:
@JakePT It seems the e2e test is not passing (I've triggered it several times now.) Do you mind fixing it before we merge this one? Thanks!

PR #2989, opened 2022-09-06
Add filter to remove unavailable terms from facet

### LINKED ISSUE #2760: Exclude facet options if there they are unavailable

When you enter a Woocommerce category or use the search and the taxonomy is not available to facet on all the options are greyed out.

I would like the option to hide them completely.

This becomes a problem when you have a "brand" attribute with 100+ brands. I rather only show the relevant brands instead of all brands in the store.

Same goes for product categories. The list can become really big with greyed out categories instead of just showing the relevant ones.

Hide all greyed out facets completely from the facet.

burhandodhy:
### Description of the Change

This PR introduces the new filter `ep_facet_taxonomy_terms` that allows filtering the list of terms in Fact. By using this filter, the user can remove the unavailable terms from the facet. Previously suggested [solution](https://github.com/10up/ElasticPress/issues/2760#issuecomment-1144889783) has an issue that shows the search box even the available terms are lower than the threshold. 

Example: The below code will remove the unavailable terms from facet 

```
add_filter( 'ep_facet_taxonomy_terms', function ( $terms, $taxonomy ) {

	$filtered_terms = array_filter( $terms, function( $term ) {
		return $term->count > 0 ;
	} );

	return $filtered_terms;
}, 10, 2 );
```

Closes #2760 

### How to test the Change

- Add the above code in theme functions.php
- Add the Facet block in widgets
- Search for any term and it will only show those terms which are available.

### Changelog Entry

> Added - `ep_facet_taxonomy_terms` filter to filter the Facet terms

### Credits

Props @burhandodhy 

### Checklist:

### DIFF
--- a/includes/classes/Feature/Facets/Types/Taxonomy/Renderer.php
+++ b/includes/classes/Feature/Facets/Types/Taxonomy/Renderer.php
@@ -122,14 +122,15 @@ class Renderer {
 		}
 
 		/**
-		 * Filter the facet terms.
+		 * Filter the taxonomy facet terms.
 		 *
-		 * @hook ep_facet_search_terms
+		 * @since  4.4.0
+		 * @hook ep_facet_taxonomy_terms
 		 * @param  {array} $terms Terms
 		 * @param  {string} $taxonomy Taxonomy name
 		 * @return  {array} New terms
 		 */
-		$terms_by_slug = apply_filters( 'ep_facet_search_terms', $terms_by_slug, $taxonomy );
+		$terms_by_slug = apply_filters( 'ep_facet_taxonomy_terms', $terms_by_slug, $taxonomy );
 
 		/**
 		 * Check to make sure all terms exist before proceeding
@@ -148,7 +149,7 @@ class Renderer {
 		$orderby = isset( $instance['orderby'] ) ? $instance['orderby'] : 'count';
 		$order   = isset( $instance['order'] ) ? $instance['order'] : 'count';
 
-		$terms = Utils\get_term_tree( $terms, $orderby, $order, true );
+		$terms = Utils\get_term_tree( $terms_by_slug, $orderby, $order, true );
 
 		$outputted_terms = array();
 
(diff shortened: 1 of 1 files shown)


includes/classes/Feature/Facets/Types/Taxonomy/Renderer.php
```diff
@@ -121,6 +121,16 @@ public function render( $args, $instance ) {
...
 
+		/**
+		 * Filter the facet terms.
+		 *
+		 * @hook ep_facet_search_terms
+		 * @param  {array} $terms Terms
+		 * @param  {string} $taxonomy Taxonomy name
+		 * @return  {array} New terms
+		 */
+		$terms_by_slug = apply_filters( 'ep_facet_search_terms', $terms_by_slug, $taxonomy );
```
felipeelia: @burhandodhy can we please add `@since 4.4.0` on this one? Also, as we now have a meta facet as well, a probably better name would be `ep_facet_taxonomy_terms` (not sure if we want to include "search" on this one, what do you think?)

includes/classes/Feature/Facets/Types/Taxonomy/Renderer.php
```diff
@@ -266,7 +276,7 @@ public function render( $args, $instance ) {
 				<?php endif; ?>
 
 				<?php
-				foreach ( $terms as $term ) :
+				foreach ( $terms_by_slug as $term ) :
```
felipeelia: Are you sure about this change? It seems we would lose the tree-organization made on parent/child categories (also, `$terms` doesn't seem to be used anywhere else.)

felipeelia (changes requested):

includes/classes/Feature/Facets/Types/Taxonomy/Renderer.php
burhandodhy: @since added. The other filters in same file do have the "search" but I think we should go without it.

includes/classes/Feature/Facets/Types/Taxonomy/Renderer.php
burhandodhy: Thanks. I totally missed that. Code updated.

felipeelia (approved):

PR #2991, opened 2022-09-08
Add Support for site__in and site__not_in parameters

### LINKED ISSUE #1781: Does the ‘sites’ parameter allow negative int for exclusion?
Can I do
```
new WP_Query( array(
    's'     => 'search phrase',
    'sites' => array( -2, -3 ),
) );
```
to exclude site 2 and 3?

burhandodhy:
### Description of the Change

This PR adds the support for the `site__in` and `site__not_in` for the  Post, Term and Comment Query. The old `sites` parameter will throw the deprecated warnings but keep working as it used to. 

Closes #1781

### Changelog Entry

> Added - `site__in` and `site__not_in` parameter support for Post, Term and Comment Query.
> Deprecated - `sites` parameter is set to depreciated for. Post, Term and Comment Query.

### Credits

Props @burhandodhy 

### Checklist:

### DIFF
--- a/includes/classes/Indexable/Comment/QueryIntegration.php
+++ b/includes/classes/Indexable/Comment/QueryIntegration.php
@@ -133,7 +133,7 @@ class QueryIntegration {
 		}
 
 		if ( ! empty( $query->query_vars['site__not_in'] ) ) {
-			$site__not_in = ! is_array( $query->query_vars['site__not_in'] ) ? array( $query->query_vars['site__not_in'] ) : $query->query_vars['site__not_in'];
+			$site__not_in = (array) $query->query_vars['site__not_in'];
 		}
 
 		/**
--- a/includes/classes/Indexable/Comment/QueryIntegration.php
+++ b/includes/classes/Indexable/Comment/QueryIntegration.php
@@ -123,12 +123,12 @@ class QueryIntegration {
 		if ( ! empty( $query->query_vars['sites'] ) ) {
 
 			_deprecated_argument( __FUNCTION__, '4.4.0', esc_html__( 'sites is deprecated. Use site__in instead.', 'elasticpress' ) );
-			$site__in = ! is_array( $query->query_vars['sites'] ) ? array( $query->query_vars['sites'] ) : $query->query_vars['sites'];
+			$site__in = (array) $query->query_vars['sites'];
 			$scope    = 'all' === $query->query_vars['sites'] ? 'all' : $site__in;
 		}
 
 		if ( ! empty( $query->query_vars['site__in'] ) ) {
-			$site__in = ! is_array( $query->query_vars['site__in'] ) ? array( $query->query_vars['site__in'] ) : $query->query_vars['site__in'];
+			$site__in = (array) $query->query_vars['site__in'];
 			$scope    = 'all' === $query->query_vars['site__in'] ? 'all' : $site__in;
 		}
 
--- a/includes/classes/Indexable/Comment/QueryIntegration.php
+++ b/includes/classes/Indexable/Comment/QueryIntegration.php
@@ -174,7 +174,7 @@ class QueryIntegration {
 				if ( ! Utils\is_site_indexable( $site_id ) ) {
 					continue;
 				}
-				$index[] = Indexables::factory()->get( 'post' )->get_index_name( $site_id );
+				$index[] = Indexables::factory()->get( 'comment' )->get_index_name( $site_id );
 			}
 
 			$this->index = implode( ',', $index );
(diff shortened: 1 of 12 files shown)


includes/classes/Indexable/Comment/QueryIntegration.php
```diff
@@ -116,8 +116,24 @@ public function maybe_filter_query( $results, WP_Comment_Query $query ) {
...
+			$scope    = 'all' === $query->query_vars['sites'] ? 'all' : $site__in;
+		}
+
+		if ( ! empty( $query->query_vars['site__in'] ) ) {
+			$site__in = (array) $query->query_vars['site__in'];
+			$scope    = 'all' === $query->query_vars['site__in'] ? 'all' : $site__in;
+		}
+
+		if ( ! empty( $query->query_vars['site__not_in'] ) ) {
+			$site__not_in = ! is_array( $query->query_vars['site__not_in'] ) ? array( $query->query_vars['site__not_in'] ) : $query->query_vars['site__not_in'];
```
felipeelia: can't we just cast it right away @burhandodhy ? Like 
```
$site__not_in = (array) $query->query_vars['site__not_in'];
```

tests/php/indexables/TestPostMultisite.php
```diff
@@ -1770,4 +1770,242 @@ public function testArchivedSite() {
...
+
+		$this->assertTrue( $query->elasticsearch_success );
+		$this->assertEquals( $query->post_count, 2 );
+		$this->assertEquals( $query->found_posts, 2 );
+
+		$this->cleanUpSites( $sites );
+	}
+
+	/**
+	 * Test a simple post content search on a subset of network sites with depreciated `sites` parameter
```
felipeelia: There is a typo here @burhandodhy: `depreciated` should be `deprecated`

tests/php/indexables/TestPostMultisite.php
```diff
@@ -1770,4 +1770,242 @@ public function testArchivedSite() {
...
+
+		$this->assertTrue( $query->elasticsearch_success );
+		$this->assertEquals( $query->post_count, 4 );
+		$this->assertEquals( $query->found_posts, 4 );
+
+		$this->cleanUpSites( $sites );
+	}
+
+	/**
+	 * Test a simple post content search with depreciated `sites` parameter
```
felipeelia: Same typo here

tests/php/indexables/TestPostMultisite.php
```diff
@@ -1770,4 +1770,242 @@ public function testArchivedSite() {
...
+
+			restore_current_blog();
+		}
+
+		$args = array(
+			's'     => 'findme',
+			'sites' => 'all',
+		);
+
+		add_action( 'ep_wp_query_search', array( $this, 'action_wp_query_search' ), 10, 0 );
```
felipeelia: Why is this needed?

felipeelia (changes requested):
EXCELLENT work on the unit tests, @burhandodhy, thanks. I've left some comments here and there with some minor changes.

One thing we will need to address (and this became obvious with your new code) is that single-site and multisite tests are always run. It would be great if we could split those up and really only run the multisite tests in multisite.

In order to parallelize these tests executions, we could create a new workflow (simply duplicating [this XML](https://github.com/10up/ElasticPress/blob/develop/.github/workflows/test.yml)) and have single tests running in one and multisite tests running in the other. What do you think? Is that something you want to tackle in this PR?

In addition to that change, we will also want to exclude "Multisite.php" classes in single-site.xml.dist

burhandodhy:
> One thing we will need to address (and this became obvious with your new code) is that single-site and multisite tests are always run. It would be great if we could split those up and really only run the multisite tests in multisite.
> 
> In order to parallelize these tests executions, we could create a new workflow (simply duplicating [this XML](https://github.com/10up/ElasticPress/blob/develop/.github/workflows/test.yml)) and have single tests running in one and multisite tests running in the other. What do you think? Is that something you want to tackle in this PR?
> 
> In addition to that change, we will also want to exclude "Multisite.php" classes in single-site.xml.dist

I removed the multisite tests from the `single-site.xml.dist`. We already have another file `phpunit.xml.dist` which runs the all the tests including the multisite tests. Do we still need a separate file for only multisite tests?

tests/php/indexables/TestPostMultisite.php
burhandodhy: Following the pattern of previous tests. We don't need this any more. Will create a new tickets to remove this line from the old tests

tests/php/indexables/TestPostMultisite.php
burhandodhy: Done.

includes/classes/Indexable/Comment/QueryIntegration.php
burhandodhy: Done

tests/php/indexables/TestPostMultisite.php
burhandodhy: Fixed

felipeelia (approved):

PR #2997, opened 2022-09-13
Use must instead of should for 'or' relations

### LINKED ISSUE #2932: BUG: Meta query using relation=or while also sorting by meta key creates disjunction with sorting meta key exists
**Describe the bug**
Creating a meta query for multiple fields with relation OR, while also sorting based on a meta field, creates a query that checks for the disjunction of the original meta query clauses together with `sorting key exists`.

**Steps to Reproduce**
As an example consider the resulting elastic query of following wp query:
```
[
  'ep_integrate' => true,
  'post_type' => 'event',
  'meta_key' => 'event_start',
  'meta_query' => [
    'relation' =>'OR',
    [
      'key' => 'event_start',
      'value' => date('Ymd'),
      'compare' => '>=',
      'type' => 'NUMERIC',
    ],
    [
      'key' => 'event_end',
      'value' => date('Ymd'),
      'compare' => '>=',
      'type' => 'NUMERIC',
    ],
  ],
  'orderby' => 'meta_value_num',
  'order' => 'ASC',
]
```
creates the elastic query: 
```
{
   ...
  "sort": [
    {
      "meta.event_start.long": {
        "order": "asc"
      }
    }
  ],
  "post_filter": {
    "bool": {
      "must": [
        {
          "bool": {
            "should": [
              {
                "exists": {
                  "field": "meta.event_start"
                }
              },
              {
                "bool": {
                  "must": [
                    {
                      "range": {
                        "meta.event_start.long": {
                          "gte": "20220808"
                        }
                      }
                    }
                  ]
                }
              },
              {
                "bool": {
                  "must": [
                    {
                      "range": {
                        "meta.event_end.long": {
                          "gte": "20220808"
                        }
                      }
                    }
                  ]
                }
              }
            ]
          }
        },
        ...
}
```

**Expected behavior**
The necessary sorting key exists check in the resulting elastic query should be located in the first `must` clauses array instead of the `should` clauses array.
E.g.:
```
{
  "post_filter": {
    "bool": {
      "must": [
        {
          "exists": {
            "field": "meta.event_start"
          }
        },
        {
          "bool": {
            "should": [
              {
                "bool": {
                  "must": [
                    {
                      "range": {
                        "meta.event_start.long": {
                          "gte": "20220808"
                        }
                      }
                    }
                  ]
                }
              },
              {
                "bool": {
                  "must": [
                    {
                      "range": {
                        "meta.event_end.long": {
                          "gte": "20220808"
                        }
                      }
                    }
                  ]
                }
              }
            ]
          }
        }
      ]
    }
  }
}
```

**Workaround**
For anyone having the same issue, i was able to create a workaround by wrapping the meta query in another array, thereby forcing elasticpress to create another query nesting layer.
```
[
  'ep_integrate' => true,
  'post_type' => 'event',
  'meta_key' => 'event_start',
  'orderby' => 'meta_value_num',
  'order' => 'ASC',
  'meta_query' => [
    [
      'relation' =>'OR',
      [
        'key' => 'event_start',
        'value' => date('Ymd'),
        'compare' => '>=',
        'type' => 'NUMERIC',
      ],
      [
        'key' => 'event_end',
        'value' => date('Ymd'),
        'compare' => '>=',
        'type' => 'NUMERIC',
      ],
    ]
  ],
]
```

**Environment information**
 - Elasticpress 4.2.2 (Bug was not present in version 4.2.0)
 - WP 6.0.1
 - Elastic Query Inspection via Debug Bar ElasticPress

MARQAS:
### Description of the Change
This PR will change the condition from using should to must for 'or' relations in meta queries. 

Closes #2932 

### How to test the Change

- While adding the query to sort the posts using key, it was taking the key inside the should clause, but now it should be under the must clause
- example: 
``` [
  'ep_integrate' => true,
  'post_type' => 'event',
  'meta_key' => 'event_start',
  'meta_query' => [
    'relation' =>'OR',
    [
      'key' => 'event_start',
      'value' => date('Ymd'),
      'compare' => '>=',
      'type' => 'NUMERIC',
    ],
    [
      'key' => 'event_end',
      'value' => date('Ymd'),
      'compare' => '>=',
      'type' => 'NUMERIC',
    ],
  ],
  'orderby' => 'meta_value_num',
  'order' => 'ASC',
] 
```
- Observe that the key is inside the must clause instead of should clause

### Changelog Entry

> Changed - The clause for meta_query when sorting by key and using 'or' relation

### Credits

Props @MARQAS 

### Checklist:


felipeelia (changes requested):
@MARQAS although the code you have here might fix the outlying issue, I'm not sure this will actually fix the problem outlined in #2932 without introducing a side-effect.

The problem seems to be coming from [this block of code](https://github.com/10up/ElasticPress/blob/4.3.0/includes/classes/Indexable/Post/Post.php#L1332-L1351), where we add a new meta query "block" if the `'meta_key'` parameter is set in WP_Query.

The problem outlined in #2932 will only happen when the WP_Query object has a meta_key set AND a meta_query using that same meta field inside an OR condition.

We will also need a new PHP Unit test for this one, so it might be helpful to start with the test and adjust the code so it passes.

felipeelia:
@MARQAS the code you had before was giving you a false positive: if you switch the first clause to `<=`, it should return all 3 posts but it returned just 2. In general, every time we have an `or` in WP it will become a `should` in Elasticsearch, and that was not happening.
I've pushed a commit changing things a bit and I think we are all set with this one. Do you mind giving it a review and letting me know if you have any questions? Thanks!

MARQAS:
@felipeelia Yes, It is working as expected. Thank you!

PR #2998, opened 2022-09-13
Prevent clicking Related Post links in the editor.

### LINKED ISSUE #2967: BUG: Related Posts links are clickable in the editor
**Describe the bug**
When using the Related Posts block it's possible to click on the links and be directed away from the editor.

**Steps to Reproduce**

1. Add a "Related Posts" block.
2. While in the editor click on one of the links.

**Expected behavior**
The link should not be followed.

**Additional context**
The Latest Posts block deals with this using a custom click handler: https://github.com/WordPress/gutenberg/blob/trunk/packages/block-library/src/latest-posts/edit.js#L157-L166

JakePT:
### Description of the Change
Prevents clicking on links in the Related Posts block in the editor. 

Originally I had planned on mimicking the behaviour of the core Latest Posts block, and the code is based on [that implementation](https://github.com/WordPress/gutenberg/blob/trunk/packages/block-library/src/latest-posts/edit.js#L157-L166), which displays a notice when attempting to click a link. However, on investigating more core blocks this behaviour is unique to the Latest Posts block and all other blocks, such as the various query loop blocks like the linked post title or date, simply silently preventDefault on any links. To avoid the overhead of needing to manage a notice I've gone with the more common convention.

This PR also converts the Related Posts block's edit component to a functional component.

Closes #2967

### How to test the Change
1. Add a Related Posts block to the editor.
2. Attempt to follow the link for one of the results.
3. The link should not be followed.

### Changelog Entry
Fixed - Clicking a Related Posts link while in the editor will no longer follow the link.

### Credits

Props @JakePT 

### Checklist:

### DIFF
--- a/assets/js/blocks/related-posts/Edit.js
+++ b/assets/js/blocks/related-posts/Edit.js
@@ -4,11 +4,8 @@
 import apiFetch from '@wordpress/api-fetch';
 import { AlignmentToolbar, BlockControls, InspectorControls } from '@wordpress/block-editor';
 import { PanelBody, Placeholder, Spinner, QueryControls } from '@wordpress/components';
-import { useInstanceId } from '@wordpress/compose';
-import { useDispatch } from '@wordpress/data';
 import { Fragment, RawHTML, useEffect, useState } from '@wordpress/element';
 import { __ } from '@wordpress/i18n';
-import { store as noticeStore } from '@wordpress/notices';
 import { addQueryArgs } from '@wordpress/url';
 
 /**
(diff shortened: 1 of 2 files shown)


assets/js/blocks/related-posts/Edit.js
```diff
@@ -1,109 +1,125 @@
...
+		event.preventDefault();
 
-	render() {
-		const {
-			attributes: { alignment, number },
-			setAttributes,
-			className,
-		} = this.props;
-		const { posts } = this.state;
+		createWarningNotice(__('Links are disabled in the editor.'), {
```
JakePT: @felipeelia I'm not an i18n expert, but can I omit the text domain if I want this message to use the core localisation for the same message in a core block? Or would I be better off just using the ElasticPress text domain and letting it be localised separately?

PR #3002, opened 2022-09-15
Fixed features are still active when disabling them in settings.

sun:
### Description of the Change

Problem
- When disabling a feature in the plugin settings, then the feature is still active on the site.

Goal
- We want to disable the Facets feature, because it manipulates taxonomy queries, causing them to yield wrong results (outside of scope of this issue).

Cause
- Disabling a feature sets the 'force_inactive' property on it, but the property is never checked when the settings are read.

Notes
- The 'active' property name is a bit misleading, it only indicates that all requirements of a feature are being met.

### How to test the Change

1. Disable the Facets feature.
2. Go to the Widgets settings page; you should not see the Facets widgets as available widgets anymore.

### Changelog Entry

> Fixed features are still active when disabling them in settings.

### Credits

Props @sun, @bogdanarizancu

w.org tha_sun, bogdanarizancu

### Checklist:

### DIFF
--- a/includes/classes/Feature.php
+++ b/includes/classes/Feature.php
@@ -186,7 +186,7 @@ abstract class Feature {
 
 		$active = false;
 
-		if ( ! empty( $feature_settings[ $this->slug ] ) && $feature_settings[ $this->slug ]['active'] && empty( $feature_settings[ $this->slug ][ 'force_inactive' ] ) ) {
+		if ( ! empty( $feature_settings[ $this->slug ] ) && $feature_settings[ $this->slug ]['active'] && empty( $feature_settings[ $this->slug ]['force_inactive'] ) ) {
 			$active = true;
 		}
 
--- a/includes/classes/Feature.php
+++ b/includes/classes/Feature.php
@@ -186,7 +186,7 @@ abstract class Feature {
 
 		$active = false;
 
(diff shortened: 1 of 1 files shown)


felipeelia:
Hi @sun,

Do you mind expanding your use case a bit? Although I see your point, I was not able to reproduce the problem, i.e., when I deactivate the Facets feature in the ElasticPress dashboard it does stay deactivated. Are you trying to do that programmatically? Thanks!

sun:
No, we're not changing the status programmatically.

Did you only test the admin UI or did you actually confirm that the Facets feature is no longer loaded and taxonomy queries (e.g. term listing page) are no longer modified?

We were seeing modifications to the queries in our debugging and discovered that the corresponding code lives in the Facets feature. After disabling the feature, we were still seeing the modifications. Looking into this we discovered that the feature status is not actually taken into account.

felipeelia:
> Did you only test the admin UI or did you actually confirm that the Facets feature is no longer loaded and taxonomy queries (e.g. term listing page) are no longer modified?

I actually confirmed that the feature is not loaded at all in a term listing page if I have Facets disabled (In fact, no ES query is fired in that scenario.)

Do you use object cache by any chance? Could this be a result of the option that stores the feature status not being properly updated in a cache layer somewhere?

felipeelia:
As we didn’t have any movement on this PR, we are now closing it. Thanks

PR #3007, opened 2022-09-19
Add option to toggle term counter

### LINKED ISSUE #2610: Instant Results: Add option to toggle term count

#2563 introduced the term count in Instant Results widgets, always displayed by default. Although it can be hidden with one CSS line, we should add an option to show/hide that.

In case it is needed, this is the CSS line to be added (to the theme or WordPress's Customizer Additional CSS):
```
.ep-search-checkbox__count { display: none; }
```

MARQAS:
### Description of the Change
This PR adds an option to enable/disable the term count option in the Instant results widget. 

Closes #2610

### How to test the Change
- By default the counter is enabled
- Go to ElasticPress features -> Instant results -> Disable the term count option
- Notice that the number of posts is disabled

### Changelog Entry

> Added - An option to toggle the term count in Instant results

### Credits

Props @MARQAS 

### Checklist:

### DIFF
--- a/includes/classes/Feature/InstantResults/InstantResults.php
+++ b/includes/classes/Feature/InstantResults/InstantResults.php
@@ -83,6 +83,7 @@ class InstantResults extends Feature {
 			'highlight_tag' => 'mark',
 			'facets'        => 'post_type,category,post_tag',
 			'match_type'    => 'all',
+			'term_count'    => '1',
 		];
 
 		$settings = $this->get_settings() ? $this->get_settings() : array();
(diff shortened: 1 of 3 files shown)


includes/classes/Feature/InstantResults/InstantResults.php
```diff
@@ -923,4 +938,18 @@ public function get_args_schema() {
...
+	/**
+	 * Sanitizes our term count settings.
+	 *
+	 * @param array $settings Array of current settings
+	 * @return mixed
+	 */
+	public function sanitize_count_settings( $settings ) {
+		if ( ! empty( $settings['instant-results']['term_count'] ) ) {
+			$settings['instant-results']['term_count'] = (bool) $settings['instant-results']['term_count'];
+		}
```
felipeelia: @MARQAS couldn't this be simplified just calling 
```
$settings['instant-results']['term_count'] = ! empty( $settings['instant-results']['term_count'] );
```

felipeelia (changes requested):
Just a minor change @MARQAS, let me know what you think about it. @JakePT, can I get your review here as well?

includes/classes/Feature/InstantResults/InstantResults.php
JakePT: @MARQAS @felipeelia Is this function required at all? None of the other Enabled/Disabled settings on the other features have anything like this. Also, sanitising as boolean isn't really worth it as the values are saved as '0' or '1' strings anyway, and the default value is `'1'` too.

PR #3011, opened 2022-09-20
Single sync option if index is deleted

### LINKED ISSUE #2962: BUG: Sync page shows a successful previous sync even if the index has been deleted
**Describe the bug**
The status of the ElasticPress Sync page is currently based on the results of the last sync operation. If a sync has not been performed yet the sync page will only offer a single option for syncing (#2858) and a "You just need to sync your content" admin notice will appear. If a sync has been performed the sync page will show the status of the last sync and display an additional option to delete and sync.

The problem is that if an index is deleted, such as by running the `wp elasticpress delete-index` WP CLI command, the status of the sync page is still based on the last sync operation. This could result in a situation where an index no longer exists but the Sync page appears to show that things have been synced successfully and the admin notice instructing the user to sync their content will not appear.

**Steps to Reproduce**

1. Install and set up ElasticPress then index your content
2. In WP-CLI run `wp elasticpress delete-index`
3. Visit the Sync page.

**Expected behavior**
Ideally the sync page and admin notice should reflect whether or not an index exists, and if one does not exist they should behave as if a previous sync has not been performed. 

I'm not sure whether the correct resolution would be to delete data about the last sync when the index is deleted, or to expand the logic of the sync page and admin notice to also check for the existence of the index. The former would be simpler but there may be reasons to keep the last sync data after deleting the index.

MARQAS:
### Description of the Change
This PR will address the issue where the Sync screen was showing 2 sync options along with the Last sync info. This will only show the Sync option which is visible during the initial syn.  

Closes #2962 

### How to test the Change

- In WP-CLI run wp elasticpress delete-index --index-name=$index_name --yes
- Visit the Sync page
- Observe that it shows single sync option without last sync info

### Changelog Entry

> Fixed - Sync page shows a successful previous sync even if the index has been deleted

### Credits

Props @MARQAS 

### Checklist:


includes/classes/Screen/Sync.php
```diff
@@ -172,6 +173,8 @@ public function admin_enqueue_scripts() {
 		}
 
 		$ep_last_index = IndexHelper::factory()->get_last_index();
+		// Get the stats of the current index
+		$index_stats = Stats::factory()->get_totals();
```
felipeelia: `get_totals()` are not the best way to check what we need @MARQAS. What we will need to do is to compare the results of `Command::get_cluster_indexes()` and `Command::get_index_names()`. As we won't be calling methods from the Command class outside the WP_CLI context, we can create similar methods in the Elasticsearch class and start calling those from the Command class and here, in the Sync class.

felipeelia (changes requested):
The approach you have here almost works @MARQAS but if you delete only one index, you get a false positive.
Steps to reproduce:
1. Enable "Post Search" and "Users" features
2. Run a sync and see 2 indices in the Health screen
3. Delete only one index (`wp elasticpress delete-index --index-name=elasticpresstest-post-1 --yes`)
4. See 2 boxes in the Sync page (it should have only the Full Sync one)

includes/classes/Elasticsearch.php
```diff
@@ -1615,4 +1616,48 @@ protected function add_query_log( $query ) {
...
+	/**
+	 * Return all indices from the cluster.
+	 *
+	 * @since 4.4.0
+	 * @return array
+	 */
+	public function get_cluster_indices() {
+		$path = '_cat/indices?format=json';
+
+		$response = self::factory()->remote_request( $path );
```
felipeelia: @MARQAS can we make it `$response = $this->remote_request( $path );` instead?

includes/classes/Screen/Sync.php
```diff
@@ -11,6 +11,8 @@
 use ElasticPress\IndexHelper;
 use ElasticPress\Screen;
 use ElasticPress\Utils;
+use ElasticPress\Stats as Stats;
+use ElasticPress\ElasticSearch as ElasticSearch;
```
felipeelia: Two things here: if we are calling the class by the same name, we don't need the `as ...` part. Also, the correct name of the last class is `Elasticsearch` in lowercase `s`.

tests/cypress/integration/dashboard-sync.cy.js
```diff
@@ -221,4 +221,45 @@ describe('Dashboard Sync', () => {
 
 		setPerIndexCycle();
 	});
+
+	it('Should only display a single sync option if index is deleted', () => {
+		/**
+		 * Reset settings.
+		 */
+		cy.wpCli('wp elasticpress delete-index --yes');
```
felipeelia: Instead of deleting *all* indices, can we test it if we delete only one (while keeping another)?

assets/js/sync/components/sync-page.js
```diff
@@ -29,11 +29,22 @@ import SyncPanel from './sync/panel';
...
+	isComplete,
+	isDeleting,
+	isEpio,
+	isSyncing,
+	log,
+	indexStats,
+	...props
+}) => {
+	const isInitialSync =
+		props.lastSyncDateTime === null || indexStats.docs === 0 || indexStats.length === 0;
```
felipeelia: I think this will give you a false positive. In fact, if you have Comments and Posts enabled and delete only one of those indices, you will have stats but it still should display the full sync box.

felipeelia (changes requested):
@MARQAS, In addition to my comments. as you've moved the get_index_names and get_cluster_indices methods to the Elasticsearch class, shouldn't that generate a change in Command.php? Thanks!

PR #3028, opened 2022-09-27
Rename wp-cli commands

### LINKED ISSUE #2981: Rename WP-CLI commands

Some of the WP-CLI command names are misleading: `clear-index`, for example, seems to imply that an Elasticsearch index would be cleared, when it actually clears a sync process. In some other cases, we use `indexes`, when the official document prefers `indices`.

We can rename some commands to make things easier to remember. As we do not want to break backward compatibility, some additional steps might be needed:

1. In most cases the PHP method is also wrong. We need to rename the current method and add a version of it with a `_deprecated_function` call. As we will end up with a bunch of those methods, we could create a new class just for that (let's say DeprecatedCommand) and put everything there.
2. WP-CLI accepts `@alias`, so we could fix the `@subcommand` lines and keep the old versions working with that. `get_last_sync` is an example.

Here is the list of public methods that would need to be renamed:

- get_cluster_indexes
- get_indexes
- index (this should be called "sync" to align with what we are using in the dashboard)
- clear_index
- get_indexing_status (get_ongoing_sync_status or something similar)
- get_last_cli_index
- stop_indexing

MARQAS:
### Description of the Change
This PR is used to rename the WP-CLI commands as requested in #2981 

Closes #2981 

### How to test the Change
- Run the old WP-CLI commands, they will give a deprecation warning along with the expected output
- Run the renamed commands, they will give the expected output

### Changelog Entry

> Added - `get-cluster-indices`, `get-indices`, `sync`, `clear-sync`, `get-ongoing-sync-status`, `get-last-cli-sync`, `stop-sync` WP-CLI commands
> Deprecated - `get-cluster-indexes`, `get-indexes`, `index`, `clear-index`, `get-indexing-status`, `get-last-cli-index`, `stop-indexing` WP-CLI commands

### Credits

Props @MARQAS 

### Checklist:


felipeelia (approved):

felipeelia (changes requested):
@MARQAS I've been thinking about this one and before we merge it there is a couple of changes we will need to make here:
1. To avoid breaking someone else's code that can be extending our Command class, instead of a whole new class, let's make DeprecatedCommand a trait and include it in the regular Command class. This way we keep methods in a separate file but they will still be available to others.
2. The `_deprecated_function` function does not output a warning to the user. Let's keep it there but also make a call to `WP_CLI::warning` and add a notice to the command doc. Examples:
https://github.com/wp-cli/language-command/blob/8e25528afc90b709aaebba429752a86d12c500d1/src/Core_Language_Command.php#L334
https://github.com/wp-cli/scaffold-command/blob/4ae0066a64db0b93b4f923554cb8ab2b921e16c9/src/Scaffold_Command.php#L207

felipeelia:
After 93e47a3, this still needs to be done (all in DeprecatedCommand.php):

1. Remove all `@alias` from the commands
2. Add `WP_CLI::warning()` calls to each command
3. Update the documentation of each command adding preceding the main desc. with `DEPRECATED` and adding the alternative to the long description
4. Test it (manual testing should be enough for this, so we don't bloat our test suites with deprecated stuff)
5. Fix the documentation of `clear_index` to include `$args` and `$assoc_args`
6. Fix git conflicts

PR #3030, opened 2022-09-28
Fix: EP returns nothing when an invalid value is passed to the tax_query field parameter

### LINKED ISSUE #3025: BUG: EP returns nothing when an invalid value is passed to the tax_query field parameter.
**Describe the bug**

By default, the tax_query `field` parameter only accepts `term_id`, `name`, `slug` or `term_taxonomy_id` If any other value passes to `field` parameter, WordPress falls back to default value that is `term_id`. This fallback case is not handled by the EP and, due to that, the EP returns no post for the query. 

**Steps to Reproduce**

1. Run the below snippet  and no post will be return
```
add_action( 'init',function() : void {


	$args = array(
		'ep_integrate' => true,
		'posts_per_page'  => 10,
		'tax_query' => array(
			array(
				'taxonomy' => 'category',
				'field'    => 'id', // invalid field name
				'terms'    => termID, // update the value
			),
		),
	);
	$query = new \WP_Query( $args );
	if ( $query->have_posts() ) : ?>
			<?php while ( $query->have_posts() ) : $query->the_post(); ?>
				<?php the_title(); ?>
			<?php endwhile; ?>
	<?php endif; ?>
	<?php wp_reset_postdata();
}  );

```
2. Set `ep_integrate` to `false` and see WordPress will return the posts

burhandodhy:
### Description of the Change

This PR fixes the issue where the EP returns no post when the invalid value passes to `field` parameter to tax_query. 

Closes #3025 

### How to test the Change

Use the below snippet. It should return the posts

```
add_action( 'init',function() : void {


	$args = array(
		'ep_integrate' => true,
		'posts_per_page'  => 10,
		'tax_query' => array(
			array(
				'taxonomy' => 'category',
				'field'    => 'id', // invalid field name
				'terms'    => termID, // update the value
			),
		),
	);
	$query = new \WP_Query( $args );
	if ( $query->have_posts() ) : ?>
			<?php while ( $query->have_posts() ) : $query->the_post(); ?>
				<?php the_title(); ?>
			<?php endwhile; ?>
	<?php endif; ?>
	<?php wp_reset_postdata();
}  );

```

### Changelog Entry

> Fixed - Returns no posts when an invalid value is passed to the tax_query field parameter

### Credits

Props @burhandodhy 

### Checklist:


includes/classes/Indexable/Post/Post.php
```diff
@@ -2153,4 +2149,31 @@ public function get_search_algorithm( string $search_text, array $search_fields,
...
+			case 'slug':
+				$field = 'slug';
+				break;
+			case 'term_taxonomy_id':
+				$field = 'term_taxonomy_id';
+				break;
+			default:
+				$field = 'term_id';
+				break;
+		}
```
felipeelia: @burhandodhy instead of a switch, can we go with a "from-to" array, like we now have [here](https://github.com/10up/ElasticPress/blob/993cba781aef2cca5a888e796684121d400a02a7/includes/classes/Indexable/Post/Post.php#L1210)? It seems to be a more readable approach. We would have something like:
```php
	protected function parse_tax_query_field( $field ) : string {
		$from_to = [
			'name'             => 'name.raw',
			'slug'             => 'slug',
			'term_taxonomy_id' => 'term_taxonomy_id',
		];

		return $from_to[ $field ] ?? 'term_id';
	}
```

felipeelia (changes requested):
Just a minor change @burhandodhy (and a git conflict that needs to be fixed.) Thanks!

includes/classes/Indexable/Post/Post.php
burhandodhy: Done

PR #3031, opened 2022-09-28
Fix: Facet returns no result for a term having accent characters

### LINKED ISSUE #3023: BUG: Facet returns no result for a term having accent characters
**Describe the bug**

Facet returns no result when the term slug has a accent character 

**Steps to Reproduce**

1. Create a new category `غير-مصنف`
2. Assign some posts to new category
3. Go to frontend and select `غير-مصنف` from taxonomy widget

**Expected behavior**

It should return the posts from `غير-مصنف` category

**Screenshots**

burhandodhy:
### Description of the Change

This PR fixes the issue where the facet returns no result for a term having accent characters

Closes #3023 

### How to test the Change

- Create a new category `غير-مصنف`
- Assign some posts to new category
- Go to frontend and select `غير-مصنف` from taxonomy widget
- EP should returns the from `غير-مصنف` category

### Changelog Entry

> Fixed - Facet returns no result for a term having accent characters

### Credits

Props @burhandodhy 

### Checklist:


felipeelia:
Facets e2e tests are consistently failing in this one @burhandodhy. Do you mind giving a look before we merge it? Thanks!

burhandodhy:
> Facets e2e tests are consistently failing in this one @burhandodhy. Do you mind giving a look before we merge it? Thanks!

@felipeelia Tests are fixed. The problem was that the `sanitize_title` function was not a good choice for meta fields and `sanitize_text_field` for taxonomies. So to fix the problem, I introduced the new method `sanitize_callback` in classes. Now every class will have the ability to change the satanize function without effecting the other classes.

felipeelia:
@burhandodhy after giving this one a deeper look, it seems we might be going in the wrong direction. Instead of fixing the sanitization of the slug, your new code actually showed me we have a different problem:

If we create a post with a meta field with that same value, the meta facet will generate a URL with a different value than the taxonomy:

https://ep-es7.test/?ep_filter_category=%25d8%25ba%25d9%258a%25d8%25b1-%25d9%2585%25d8%25b5%25d9%2586%25d9%2581&ep_meta_filter_my_first_field=%D8%BA%D9%8A%D8%B1-%D9%85%D8%B5%D9%86%D9%81 (see how `ep_filter_category` is different from `ep_meta_filter_my_first_field` but they should be the same.)

Investigating it a little bit, it seems things work well if we use `$term->name` instead of `$term->slug`. Do you mind giving it another look? Feel free to ping me on slack to discuss the best way to move forward.

includes/classes/Feature/Facets/FacetType.php
```diff
@@ -34,4 +34,12 @@ abstract public function get_filter_name() : string;
 	 * @return string The filter name.
 	 */
 	abstract public function get_filter_type() : string;
+
+	/**
+	 * Get the facet sanitize function.
+	 *
+	 * @return string The function name.
+	 */
+	abstract public function get_sanitize_callback() : string;
```
felipeelia: @burhandodhy so, given the fact that probably `sanitize_text_field` would be the "default" method, what do you think about making it a concrete method that defaults to sanitize_text_field (with a filter for it as well, like `ep_facet_default_sanitize_callback`?

includes/classes/Feature/Facets/Facets.php
```diff
@@ -360,29 +360,27 @@ public function get_aggs( $response, $query, $query_args, $query_object ) {
...
 
 			foreach ( $filter_names as $filter_type => $filter_name ) {
 				if ( 0 === strpos( $key, $filter_name ) ) {
-					$facet = str_replace( $filter_name, '', $key );
+					$facet             = str_replace( $filter_name, '', $key );
+					$sanitize_callback = $sanitize_callbacks[ $filter_type ];
 
 					$filters[ $filter_type ][ $facet ] = array(
-						'terms' => array_fill_keys( array_map( 'trim', explode( ',', trim( $value, ',' ) ) ), true ),
+						'terms' => array_fill_keys( array_map( $sanitize_callback, explode( ',', trim( $value, ',' ) ) ), true ),
```
felipeelia: I guess we could use the opportunity to refactor this line a bit. Several methods are being called in the same line :)

felipeelia (changes requested):
@burhandodhy great progress on this one, thanks. I've left yet another comments/suggestions/changes here, let me know your thoughts about it.
One thing that I'm trying to do is to add unit tests for all filters/actions being called. Do you think it would be possible to add unit tests for these new filters as well? Thanks!

ps.: there is also a conflict with the facets e2e tests, if you don't mind fixing it.

burhandodhy:
@felipeelia I made the changes that you requested. I also added the tests for the filters.

tests/php/features/TestFacetTypeMeta.php
```diff
@@ -262,4 +262,38 @@ public function testInvalidateMetaValuesCache() {
...
+		parse_str( "ep_meta_filter_new_meta_key_1={$test_meta}", $_GET );
+		$selected = $facet_feature->get_selected();
+
+		// test sanitize_text_field runs by default on taxonomy facets
+		$expected_result = sanitize_text_field( $test_meta );
+		$this->assertArrayHasKey( $expected_result, $selected['meta']['new_meta_key_1']['terms'] );
+
+		$sanitize_function = function() {
+			return 'sanitize_title';
+		};
```
felipeelia: Can we just make sure we are testing something here? I was thinking of something like:

```suggestion
		$sanitize_function = function ( $function ) {
			$this->assertSame( 'sanitize_text_field', $function );
			return 'sanitize_title';
		};
```

We probably should also test if the filter is being applied in the first place (that `assertSame` wouldn't be called if the filter is not called.) Perhaps a global variable changed inside the filter and tested after the `->get_selected();` call or something similar.

tests/php/features/TestFacetTypeMeta.php
```diff
@@ -262,4 +262,38 @@ public function testInvalidateMetaValuesCache() {
...
+	/**
+	 * Test get_sanitize_callback method.
+	 *
+	 * @since 4.4.0
+	 * @group facets
+	 */
+	public function testGetSanitizeCallback() {
+
+		$facet_feature = Features::factory()->get_registered_feature( 'facets' );
+		$test_meta     = 'This is s test meta';
```
felipeelia: There is a typo here @burhandodhy but I think we could one step further and actually test a string that needs to be sanitized. What do you think?

tests/php/features/TestFacetTypeTaxonomy.php
```diff
@@ -175,4 +175,38 @@ public function testAddQueryFilters() {
...
+		parse_str( "ep_filter_taxonomy={$test_taxonomy}", $_GET );
+		$selected = $facet_feature->get_selected();
+
+		// test sanitize_title runs by default on taxonomy facets
+		$expected_result = sanitize_title( $test_taxonomy );
+		$this->assertArrayHasKey( $expected_result, $selected['taxonomies']['taxonomy']['terms'] );
+
+		$sanitize_function = function() {
+			return 'sanitize_text_field';
+		};
```
felipeelia: Same as above

felipeelia (changes requested):
We are almost there, @burhandodhy!

tests/php/features/TestFacetTypeMeta.php
burhandodhy: Done

tests/php/features/TestFacetTypeMeta.php
burhandodhy: Thank you for pointing it out. @felipeelia we are already testing the sanitized string https://github.com/10up/ElasticPress/pull/3031/files#diff-d36fb3af9a833294f34cf95964cbdc054fb4d401805400564aa71bc8ce01db90R196

felipeelia (approved):
Thanks for all the hard work on this one @burhandodhy!

PR #3037, opened 2022-09-30
Apply deprecated filters without overwritting values

### LINKED ISSUE #3033: BUG: ep_post_match_fuzziness value overridden by default value of deprecated ep_match_fuzziness value
**Describe the bug**

The deprecated filter `ep_match_fuzziness` is run after its replacement filter `ep_post_match_fuzziness` despite the first filter being defined with `add_filter` and returning a value.

As such, fuzziness is always set to "auto", the default value for the deprecated filter.

**Steps to Reproduce**

1. Add the filter

```php
add_filter('ep_post_match_fuzziness', function ($fuzziness): int {
  return 1;
});
```

When the search is performed this is called (Version_400.php line 87):

```php
apply_filters( "ep_{$indexable_slug}_match_fuzziness", 'auto', $search_fields, $query_vars ),
```

Then this is called (Version_400.php line 162):

```php
$query['bool']['should'][1]['multi_match']['fuzziness'] = apply_filters_deprecated(
  'ep_match_fuzziness',
  [ 'auto', $search_fields, $query_vars ],
  '4.3.0',
  'ep_post_match_fuzziness'
);
```

...which overrides the previous value to "auto".
 
**Expected behavior**

If the replacement filter is defined, then the deprecated filter should not be called.

**Environment information**
 - WordPress version: 6.0.2
 - ElasticPress version: 4.3.1

felipeelia:
### Description of the Change

Closes #3033

### Changelog Entry

> Fixed - Deprecated filters for search algorithms do not overwrite values set with the newer filters

### Credits

Props @felipeelia @marc-tt 

### Checklist:

### DIFF
--- a/includes/classes/SearchAlgorithm/DefaultAlgorithm.php
+++ b/includes/classes/SearchAlgorithm/DefaultAlgorithm.php
@@ -148,7 +148,7 @@ class DefaultAlgorithm extends \ElasticPress\SearchAlgorithm {
 		 */
 		$query['bool']['should'][0]['multi_match']['boost'] = apply_filters_deprecated(
 			'ep_match_phrase_boost',
-			[ 4, $search_fields, $query_vars ],
+			[ $query['bool']['should'][0]['multi_match']['boost'], $search_fields, $query_vars ],
 			'4.3.0',
 			'ep_post_match_phrase_boost'
 		);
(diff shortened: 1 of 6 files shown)


burhandodhy (approved):
LGTM

PR #2588, opened 2022-02-08
WIP: Feature/mapping weighting dashboard

mehidi258:
### Description of the Change

Closes #1690 

### Checklist:

- [x] I have read the [**CONTRIBUTING**](/CONTRIBUTING.md) document.
- [x] My code follows the code style of this project.
- [x] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [ ] I have added tests to cover my change.
- [ ] All new and existing tests passed.

### Credits

Props @

PR #3068, opened 2022-10-19
New weighting screen with meta support

### LINKED ISSUE #1690: Weighting Engine: Post meta support
Registering meta keys requires additional effort. We can manage searchable meta fields if we provide a UI on the weighting engine.

Adding post meta capabilities to the weighting engine would be a good solution.

**Designs**
Screen Shot 2020-03-06 at 14 47 06

* Showing most-used meta keys with a group query and showing them for each post_type. (It's against simplicity and we might get performance issues on large-scale websites.)
* Registering post meta section for each post_type. (It didn't seem the best solution for UI. It's not easy to manage if the same key will be used across multiple post_types)

**Additional context**

We might consider showing (read-only) meta fields that registered via `ep_search_fields`.

JakePT:
### Description of the Change
This PR implements a new `@wordpress/element`-based weighting dashboard with support for making meta fields searchable.

elasticpress test_wp-admin_admin php_page=elasticpress-weighting (1)

Originally based on #2588 but with significant rewrites that were required to work with real data and adapt to changes decided on in subsequent UX discussions.

This PR changes the way meta is indexed by ElasticPress. Currently, by default, ElasticPress only indexes:

- Public meta (keys not prefixed with a `_`)
- Keys added using the `ep_prepare_meta_allowed_protected_keys` filter. 
 
With this change applied:

1. ElasticPress will only index keys added using `ep_prepare_meta_allowed_protected_keys`. This includes keys added by the WooCommerce feature, to ensure that the integration works properly. No other keys will be indexed by deafult, public or private.
2. These keys will be listed on the Weighting screen automatically, in the _Metadata_ section, and the user can choose to make them searchable, and weigh them.
3. Fields and groups added using the `ep_weighting_fields_for_post_type` filter, as per [this blog post](https://www.elasticpress.io/blog/2020/06/custom-fields-and-weighted-search-with-elasticpress-part-2/), are supported by the new screen. The new default _Metadata_ section uses the key `ep_metadata` to avoid conflicts with custom groups, such as the one added in that post's example.
4. The user can manually add additional keys to the index from the weighting screen using the _Add field_ form in the _Metadata_ section of each post type.
5. Behind the scenes ElasticPress knows to index manually-added keys by finding them in the weighting settings, so a key can be set to be indexed but not searchable by adding the key from the weighting dashboard without checking _Searchable_.
6. The previous behaviour can be restored using the `ep_meta_mode` filter. When `'auto'` is returned by the filter ElasticPress will again index all public meta keys, but users will not be able to manually add keys using the weighting dashboard. When ElasticPress is network activated, this will continue to be the default behaviour.

This has the potential to be a breaking change, which is why it was moved to 5.0. Issues could arise if a site is using queries that are integrated with ElasticPress that assume all public meta keys will be indexed. To resolve this issue the user can add the required keys using the new functionality of the weighting screen, or they could restore the previous behaviour using the `ep_meta_mode` filter.

Closes #1690 

### How to test the Change
WIP

### Changelog Entry
Added - New weighting dashboard with support for making meta fields searchable.

### Credits

Props @JakePT, @mehidi258

### Checklist:


JakePT:
This is now in a state where the front end is fully functional (although a confirmation notice on save wouldn't go amiss), so I think it's time to hand it over for the back-end portion to be implemented. There are a number of things that need to be done on the back end.

The main thing is to respect the new `ep_meta_mode` option. When set to `auto` (the default) this is the default behaviour of ElasticPress prior to this change. It should continue to index public meta keys and any keys required by any active Features. However when this is set to `manual` only fields required by features or manually added to the weighting configuration should be indexed. In the UI the list of fields required by features is determined by the `ep_prepare_meta_allowed_protected_keys` filter, which is how the WooCommerce feature indexes [its metadata](https://github.com/10up/ElasticPress/blob/4.3.1/includes/classes/Feature/WooCommerce/WooCommerce.php#L57-L131), but there may be a better way to handle this.

At the current time any fields added to the weighting configuration will remain even if manual management of metadata is switched back off. This has the advantage of keeping any weighting changes the user made even if they temporarily disable manual metadata management, but could cause issues if weighting is applied for fields that are no longer indexed, or if unexpected fields are searchable. The two possible solutions for this are:

1. If `ep_meta_mod` is `auto` when applying weighting, any fields like `meta.${key}.value` from the weighting configuration should be ignored. The complication is that the SKU and Variation SKU fields are metadata.
2. When saving the weighting configuration, if `ep_meta_mod` is `auto`, sanitise the saved configuration to remove weighting for any meta fields. Again, the SKU and Variation SKU fields will need special handling.

In either case, handling the SKU and Variation SKU fields (and potentially fields added by users or their theme/plugins) could be achieved by cross referencing the weighting configuration with `get_weightable_fields_for_post_type()` and checking which group the field belongs to.

So I believe the remaining work required, most of it back end, is:

1. Only index selected fields and fields required by features when manual metadata management is enabled.
2. Properly handle saved weighting configuration for meta fields that are no longer indexed or should no longer be searchable once manual management is disabled.
3. Add some sanitisation/validation to the weighting configuration.
4. Add a save confirmation message and error handling to the weighting screen (Front-end).

Additionally it may be worth adding information to the response when saving the weighting configuration to indicate whether a reindex is required so that the front end can offer to sync.

JakePT:
@felipeelia Following our discussion today I have now:

1. Given the Metadata field group a unique ID to avoid conflicts with https://www.elasticpress.io/blog/2020/06/custom-fields-and-weighted-search-with-elasticpress-part-2/. If a user has used code based on the example at that link to add fields, those fields will be displayed in a separate group to the new Metadata group. This means they will appear regardless of whether manual metadata management has been enabled. Such fields would also continue to be indexed regardless of the manual metadata setting because they will have a weighting configuration.
2. Sanitised the weighting configuration on save so that if manual metadata management is disabled any weighting configuration for fields in the metadata group is removed. This ensures that the weighting configuration is valid for weighting regardless of whether metadata is being managed manually.

With those changes in place, all that is required on the back-end now is changes to indexing so that:

 - If `ep_meta_mode` (`$weighting->get_ep_meta_mode()`) is `manual` then the only meta keys that are indexed are:
   1. Fields that are required by features (currently determined by `ep_prepare_meta_allowed_protected_keys`).
   2. Fields with a weighting configuration (`$weighting->get_weighting_configuration()` are indexed.
 - `ep_meta_mode` (`$weighting->get_ep_meta_mode()`) is `auto` (the default), then the current behaviour is followed and the indexed meta keys are:
   1. Fields that are required by features (currently determined by `ep_prepare_meta_allowed_protected_keys`).
   2. Public meta keys (i.e. not prefixed with `_`).

With the sanitising in place no changes to weighting should be required.

felipeelia:
@JakePT I've pushed some commits addressing the backend part that was needed. I've also made some changes that I'd like you to review:

- Brought back the old methods and marked them as deprecated (who knows who depends on those out there...)
- Make sure we always use 'auto' in multisite, as we don't have the weighting dashboard on those
- Do not enqueue weighting files on install (e2e tests were failing due to a React error here -- the element was not found on the page)
- Pull the meta fields from the database instead of Elasticsearch

Things that need to be done:

1. Fix weighting.cy.js e2e tests
2. Write new tests for this new functionality (I wrote some unit tests, there is room for more. We will also want e2e tests)
3. Think about caching those meta fields coming from the database (I can bring that to our sync tomorrow)

Are numbers 1 and 2 things you think you could help with?

Thanks!

felipeelia:
@JakePT an update on the above:

1. Done
2. Still needs to be done
3. We've talked and there is no need to cache it. I've added some filters in case someone doesn't want to run the query or wants to tweak its results.

JakePT:
@felipeelia I have pushed updates to the tests, including a test for adding and weighting a meta field. I also added a settings saved confirmation notice.

Regarding the meta fields dropdown, I've noticed that now that we're using the `SELECT DISTINCT meta_key` query a lot of metadata from post types that aren't searchable is included. In particular I noticed a lot of fields for menu items. Perhaps we'll need to query these only for searchable post types? And if we're doing that we might as well only get fields relevant for each post type. What do you think?

Also, I gotta confess I'm not a huge fan of the new copy for the manual metadata checkbox, for a few reasons:

1. I think just using "meta" instead of "metadata" or "custom fields" is a tiny bit more confusing for people who aren't WordPress devs. I prefer "metadata" over "custom fields" because not all metadata necessarily has an associated UI field.
2. The previous copy actually covered more information since it was explicit that metadata was not searchable without it.
3. I don't like that it says "full" control, since that's not entirely true due to the fields required by features.
4. I don't think it's clear to regular users what is "public" about custom fields without a `_`, because in many cases that data might not necessarily be truly "public" in the non-technical sense.
5. I think having the "(may require re-sync)" in parentheses in the checkbox label looks a bit messy.
 
Here's another pass at the original text that might work:

>[ ] Enable searchable metadata
>By default ElasticPress will automatically sync all metadata required for ElasticPress to function, as well as all fields without a `_` prefix, but that metadata is not searchable. Check this box to take control over which metadata fields are synced and to allow metadata fields to be made searchable.
> **Note:** Enabling this option, or any changes to which metadata fields are synced, will require re-syncing your content.

I'm now working on automatically syncing as per your last Slack message.

JakePT:
Revisiting the status of this. I have brought the branch up to date with develop. As I understand it these are some of the outstanding issues that need to be resolved before we can proceed with this:

- [x] **Revisit metadata.** Currently this is an option but there was talk of removing the option and making manual management the default behaviour as a breaking change for 5.0.
   - [ ] If this is the default behaviour do we need to automatically index all the WooCommerce fields?
- [x] **Revise copy.** Regardless of the decision above there are some strings that refer to 'indexing', instead of 'syncing'.
- [x] **Revisit syncing.** When fields are added a sync will need to be performed. I don't believe this is currently in place.
- [ ] **Revisit unit tests.** Since the last update to the feature branch some unit tests have been introduced that will fail for the new screen. Unit tests will need to be updated.
- [ ] **Revisit E2E tests.** There are some E2E tests but we should check that coverage is sufficient.
- [ ] **Facetable fields.** To properly support block themes and Instant Results blocks we will need a way to mark fields for use in aggregations. A checkbox on this page would be an ideal place to do so.

JakePT:
I have pushed a few changes to:

1. Remove the checkbox to enable manual management of metadata.
2. Changed Weighting::get_meta_mode() to return the value of a filter, rather than an option. I changed the default value to `manual`, so the filter can be used to restore the 4.x default behaviour by setting it to `auto`.
3. Tweaked some copy to consistently refer to "indexing" as "syncing".
4. Added functionality to automatically sync if new meta fields are added. This works by checking if any new meta fields have been added to the weighting configuration _and_ if that field is not already synced automatically. Syncing will not happen automatically if a field is removed. If a sync is required the button will update to "Save Changes and Sync".

JakePT:
- [x] Use `auto` mode for metadata when in Network Mode.
- [x] Only show custom fields for relevant post types.
- [x] Tidy up Add custom field section.
- [x] Hide Taxonomies for Pages.

PR #3100, opened 2022-11-01
Add Option to exclude post from Search

### LINKED ISSUE #253: Ignore certain posts when indexing
Just starting a discussion here. 

Our site has a lot of private pages that we don't want to show up in the SERP, but we don't set these to private status in WP because we often direct non-users to them. I'm thinking a per-post meta-box that includes an option to exclude the post from the ES index.

Is this a potential for core EP? Or would it be better suited for a plugin. Anyone have additional thoughts?

burhandodhy:
### Description of the Change

This PR adds the new option to exclude the certain post from Search.

Closes #253

### Changelog Entry

> Added - Option to exclude certain posts from Search

### Credits

Props @burhandodhy @felipeelia @JakePT 

### Checklist:


JakePT:
@felipeelia @burhandodhy I've done a review of the front end in progress and pushed some tweaks:

1. I updated the `auth_callback` for the meta field to explicitly check permissions for the specific field and the specific post being edited.
2. I organised the JS into  a file/folder for the feature, for consistency with most of the other scripts, and I updated the hook that it's enqueued with so that checking the current page isn't necessary.
3. I simplified the code for the checkbox plugin. The useState and useEffect calls weren't really necessary, and I adopted useSelect and useDispatch for handling the data. With previous implementation the "Update" button wasn't becoming enabled when the checkbox was changed for some reason, but this seems to have fixed that issue.
4. I updated the "Exclude from search" label to be sentence case, for consistency with the other checkboxes, such as "Stick to the top of the blog".

Let me known if you have any questions about the changes.

felipeelia:
Before the release: let's create an article in the Help Center explaining how this works.

burhandodhy (commented):
LGTM

PR #3111, opened 2022-11-04
Fix: Instant Result label issue

### LINKED ISSUE #3097: BUG: In Instant Result, term labels are encoded if it contains special HTML character
**Describe the bug**

When a term contains special HTML characters, the Instant Result shows those characters as encoded HTML.

**Steps to Reproduce**

1. Create a new category for example `Macaroni & Cheese`
2. Assign post to that category
3. Enable the Instant Result feature
4. Search for a post on the frontend
5. `&` character is encoded to `&amp;`

**Expected behavior**

The Label shouldn't be encoded.

**Screenshots**

image

burhandodhy:
### Description of the Change

This PR fixes the issue where the HTML is not decoded in the term label

Closes #3097

### Changelog Entry

> Fixed - Encoded Category Label shows in Instant Result

### Credits

Props @burhandodhy 

### Checklist:

### DIFF
--- a/assets/js/instant-results/components/common/checkbox.js
+++ b/assets/js/instant-results/components/common/checkbox.js
@@ -14,12 +14,20 @@ import { WPElement } from '@wordpress/element';
  * @returns {WPElement} Component element.
  */
 export default ({ count, id, label, ...props }) => {
+	let labelHTML = label;
+	if (count) {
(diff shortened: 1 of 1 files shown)


JakePT:
I'm closing this PR in favor of #3113 which uses a package to decode HTML entities only, while still allowing React to escape the result, and applies the decoding at a stage where the active filter chip will also display properly.

PR #3130, opened 2022-11-11
Status Report Page

### LINKED ISSUE #3077: Better Debugging and Status Report
- [ ] Change the existing debug info in Health Tools to `'private' => false,` so they can be retrieved while copying/pasting it
- [ ] Add a notice in the Sync Page if we detect the site has more public meta than Elasticsearch would technically support
- [ ] Add some information about the number of meta fields (total and per post-type) and the number of fields already stored in Elasticsearch

felipeelia:
### Description of the Change
This PR improves debugging and also adds a new "Status Report" page in the dashboard

Closes #3077

### To-Do

- [x] Unit Tests
  - [x] process_too_many_fields_notice method
  - [x] ep_update_index_settings action
  - [x] get_index_total_fields_limit method
  - [x] get_distinct_meta_field_keys_db method
  - [x] get_distinct_meta_field_keys_db_per_post_type method
  - [x] get_indexable_meta_keys_per_post_type method
- [x] Copy Review
  - [x] process_too_many_fields_notice (warning, "You may have...")
  - [x] process_too_many_fields_notice (error, "You definitely have...")
  - [x] Meta Keys (Limited): "Due to the number of posts on the site, the result set per post type is limited."
- [x] Visual
  - [x] Status Report Page (general alignment, etc.)
- [x] Fix
  - [x] Last Sync info (totals vs. synced with more than one indexable)
- [x] List post content related reports (post count and meta) per site in multisite installations

### Changelog Entry

> Added - New Status Report page.

### Credits

Props @felipeelia, @brandwaffle 

### Checklist:


includes/classes/AdminNotices.php
```diff
@@ -743,6 +744,66 @@ protected function process_yellow_health_notice() {
...
+		$es_field_limit = Elasticsearch::factory()->get_index_total_fields_limit( $index_name );
+		$es_field_limit = $index_settings ?? apply_filters( 'ep_total_field_limit', 5000 );
+
+		$predicted_es_field_count = $count_fields_db * 8;
+
+		if ( $predicted_es_field_count > $es_field_limit ) {
+			return [
+				'type'    => 'error',
+				'dismiss' => true,
+				'html'    => __( 'You definitely have more fields than ES will support.', 'elasticpress' ),
```
tott: let's remove the "definitely" :)

tott: is there a way we can provide more instructions on what to do next. ie link to support article

includes/classes/Elasticsearch.php
```diff
@@ -1672,15 +1682,50 @@ public function get_index_names() {
...
-	 * @return WP_Error|array WP_Error on failure or The response
+	 * @return array Array of indices in Elasticsearch
 	 */
 	public function get_cluster_indices() {
 		$path = '_cat/indices?format=json';
 
 		$response = $this->remote_request( $path );
 
-		return $response;
+		return json_decode( wp_remote_retrieve_body( $response ), true );
```
tott: I think this might need some more error handling and casting to ensure it is not causing any errors in case of unexpected results.

includes/classes/Indexable/Post/Post.php
```diff
@@ -2385,4 +2385,234 @@ protected function parse_tax_query_field( string $field ) : string {
...
+
+		$meta_keys        = [];
+		$post_ids_batches = $this->get_lazy_post_type_ids( $post_type );
+		foreach ( $post_ids_batches as $post_ids ) {
+			$new_meta_keys = $this->get_meta_keys_from_post_ids( $post_ids );
+
+			$meta_keys = array_unique( array_merge( $meta_keys, $new_meta_keys ) );
+		}
+
+		set_transient( $cache_key, $meta_keys, DAY_IN_SECONDS );
```
tott: any chance this hits the 1MB limit? do we need to truncate meta_keys?

includes/classes/Elasticsearch.php
felipeelia: Casting added in 4258eda. Regarding error handling, do you have any specific scenario in mind @tott? So far, I'm relying on the underlying functions:
- If `$response` is an error, wp_remote_retrieve_body will return an empty string
- If `json_decode` fails, it will return null
Thanks!

includes/classes/Indexable/Post/Post.php
felipeelia: Good point, thanks @tott. I've addressed that in e746a52

felipeelia:
This link contains a preview of the page: https://imgur.com/4F1UQMZ

includes/classes/StatusReport/ElasticPressIo.php
```diff
@@ -0,0 +1,167 @@
...
+	 * @return array|null
+	 */
+	protected function get_instant_results_field() : array {
+		$index = Indexables::factory()->get( 'post' )->get_index_name();
+
+		if ( ! $index ) {
+			return [];
+		}
+
+		$feature  = new InstantResults\InstantResults();
```
JakePT, line 148: @felipeelia I had to use `new` here instead of using the Features factory otherwise this would only get the search template for the main site's index. Let me know if this is likely to be a problem or if you have a better approach. Also, I'm wondering if we need to do something similar for the Autosuggest parameters on multisite?

This has also revealed that the search template for secondary sites on a multisite network will not get saved without WP CLI because there are no weighting settings to save per-site and it's not possible to sync per-site through the dashboard. I'll create a separate issue for that.

PR #3162, opened 2022-11-28
Create a new ElasticPress.io page with client messages

felipeelia:
### Description of the Change

This PR introduces a message system between the EP.io infrastructure and EP.io customers, making it possible to warn users about maintenance windows, for example.

### Changelog Entry

> Added - ElasticPress.io messages system.

### Credits

Props @felipeelia @JakePT 

### Checklist:


JakePT:
@felipeelia I have pushed a change to add a badge counter to the ElasticPress admin menu item, and added a transient to support this without having to query for messages on every page load. Still to be discussed is whether these need to be displayed on their own admin screen or if they can go in Index Health or similar.

includes/classes/ElasticPressIo.php
```diff
@@ -0,0 +1,56 @@
...
+		$messages  = get_transient( $transient );
+
+		if ( false !== $messages ) {
+			return $messages;
+		}
+
+		$response = \ElasticPress\Elasticsearch::factory()->remote_request( 'endpoint-messages' );
+		$messages = (array) json_decode( wp_remote_retrieve_body( $response ), true );
+
+		set_transient( $transient, $messages, DAY_IN_SECONDS );
```
JakePT: Is this too long? Not long enough?

JakePT:
I’ve updated the PR. I have:
 - Added an `ElasticPressIo` class for interacting with ElasticPress.io (a todo may be to move some of the Autosuggest and Instant Results template stuff into here).
 - Move the method for retrieving endpoint messages to that class. I added code to cache the results in a transient with a 24 hour expiry. This was necessary for the count badge.
 - Added a counter to the ElasticPress admin menu item, and Status Report submenu item, indicating the number of messages found with the above method.
 - Added support for 'messages' to Status Report reports. The ElasticPress.io report gets messages using the above method.
 - Updated the Status Report UI to print messages as notices under each report heading.
 - Moved the ElasticPress.io report, and the Failed Queries report, to the top of the Status Report.

Screenshot 2022-12-21 at 9 46 27 pm

anjulahettige:
Ping ElasticPress every 1 hour.

PR #3175, opened 2022-12-02
Orders autosuggest foundations

JakePT:
### Description of the Change
The PR implements the foundations for the orders autosuggest for WooCommerce feature. It includes:

- A class for the Orders sub-feature, this incudes:
  - Creating and saving the search template for the order search API on ElasticPress.io.
  - Retrieving and setting a temporary token for authenticated front-end requests to ElasticPress.io.
  - Loading front-end assets.
- The UI.

The class is not instantiated in the WooCommerce feature, so it can not be currently activated.

Screenshot 2023-02-01 at 12 39 22 am

Remaining to be implemented, as per #3277, are:

- A new field and ngram that includes the order number, first name, last name, and email address to support partial word matches and highlighting.
- A UI toggle for enabling Orders Autosuggest.
- Logic to index Orders when feature is active, even if Protected Content is disabled.
- Logic to require index when the feature is activated.
- Tests.

### How to test the Change
1. Connect to an EP.io environment that supports the order search API.
2. Activate the WooCommerce and Protected Content features.
3. Add `( new Orders() )->setup();` to the `setup()` method of the WooCommerce feature class.
4. Perform an index.
5. Visit _WooCommerce > Orders_ and start searching for orders. Suggestions should be offered in a listbox below the search form.
 
### Changelog Entry
- Added: Autosuggest for WooCommerce Orders.

### Credits
@JakePT 

### Checklist:


includes/classes/Feature/WooCommerce/Orders.php
```diff
@@ -0,0 +1,388 @@
...
+
+if ( ! defined( 'ABSPATH' ) ) {
+	exit; // Exit if accessed directly.
+}
+
+/**
+ * WooCommerce Orders Feature
+ *
+ * @since 4.5.0
+ * @package ElasticPress\Feature\WooCommerce
```
felipeelia: @JakePT do you mind moving the `@since` tag to the file docblock? Also, I guess we can avoid `@packages` at a class level and keep only the file level one. We are currently using different solutions in different places, so let's keep it as simple as possible and change it back if we decide differently later.

includes/classes/Feature/WooCommerce/Orders.php
```diff
@@ -0,0 +1,388 @@
...
+	 * Enqueue admin assets.
+	 *
+	 * @param string $hook_suffix The current admin page.
+	 */
+	public function enqueue_admin_assets( $hook_suffix ) {
+		if ( 'edit.php' !== $hook_suffix ) {
+			return;
+		}
+
+		if ( 'shop_order' !== $_GET['post_type'] ) { // phpcs:ignore WordPress.Security.NonceVerification.Recommended
```
felipeelia: [question] Isn't this throwing a notice in logs if post_type is not in the query string?

includes/classes/Feature/WooCommerce/Orders.php
```diff
@@ -0,0 +1,388 @@
...
+			array(
+				'adminUrl'      => admin_url( 'post.php' ),
+				'apiEndpoint'   => $api_endpoint,
+				'apiHost'       => ( 0 !== strpos( $api_endpoint, 'http' ) ) ? esc_url_raw( $api_host ) : '',
+				'authorization' => "Basic $temporary_token",
+				'argsSchema'    => $this->get_args_schema(),
+				'dateFormat'    => wc_date_format(),
+				'statusLabels'  => wc_get_order_statuses(),
+				'timeFormat'    => wc_time_format(),
+			)
```
felipeelia, line 148: thought [non-blocking]: we probably should add some filters here so people can easily change these values.

includes/classes/Feature/WooCommerce/Orders.php
```diff
@@ -0,0 +1,388 @@
...
+	/**
+	 * Save the search template to ElasticPress.io.
+	 *
+	 * @return void
+	 */
+	public function epio_save_search_template() {
+		$endpoint = $this->get_template_endpoint();
+		$template = $this->get_search_template();
+
+		$response = Elasticsearch::factory()->remote_request(
```
felipeelia: It seems `$response` is never used, so can we simply don't assign it to a variable? (Or people will get tempted to add `$response` to the action below but as the request is non-blocking, it will go empty.)

includes/classes/Feature/WooCommerce/Orders.php
```diff
@@ -0,0 +1,388 @@
...
+		add_filter( 'ep_is_integrated_request', [ $this, 'is_integrated_request' ], 10, 2 );
+
+		$query = new \WP_Query(
+			array(
+				'ep_integrate'             => true,
+				'ep_order_search_template' => true,
+				'post_status'              => array_keys( $order_statuses ),
+				'post_type'                => 'shop_order',
+				's'                        => '{{ep_placeholder}}',
+			)
```
felipeelia, line 272: This will also need a filter. If adding it to this PR sounds too early let's capture that need somewhere else.

includes/classes/Feature/WooCommerce/Orders.php
```diff
@@ -0,0 +1,388 @@
...
+		if ( is_wp_error( $response ) ) {
+			return false;
+		}
+
+		$response = wp_remote_retrieve_body( $response );
+		$response = json_decode( $response );
+
+		$token = base64_encode( "$response->username:$response->clear_password" ); // phpcs:ignore WordPress.PHP.DiscouragedPHPFunctions.obfuscation_base64_encode
+
+		update_user_meta( $user_id, 'ep_temporary_token', $token );
```
felipeelia, line 382: We need to delete these during the plugin uninstall. Also, this will eventually expire, right? I didn't find where we are getting a new one if that is the case, sorry :(

felipeelia (changes requested):
@JakePT I left some comments (mostly just thoughts and things we will need to keep track of). Do you mind reaching out to me on slack to share an endpoint I can use to test it? Thanks!

includes/classes/Feature/WooCommerce/Orders.php
JakePT: @felipeelia This query can be filtered using `pre_get_posts` and checking `$query->get( 'ep_order_search_template' )`.

includes/classes/Feature/WooCommerce/Orders.php
JakePT: @felipeelia Yeah I'll check with Dustin how we want to do that. At the moment it's treated as if the tokens are permanent.

includes/classes/Feature/WooCommerce/Orders.php
JakePT: @JakePT So I'm hesitant to allow filtering this for the same reason I was hesitant bout it for Instant Results. I don't think this is mature enough yet. The endpoint does have a filter though, and I believe the host is also filterable (this is largely copied from Instant Results). The Woo formats and statuses also already have their own filters.

includes/classes/Feature/WooCommerce/Orders.php
JakePT: @felipeelia Done.

includes/classes/Feature/WooCommerce/Orders.php
JakePT: @felipeelia Good catch. Fixed.

includes/classes/Feature/WooCommerce/Orders.php
JakePT: @felipeelia Good catch. Fixed.

PR #3196, opened 2022-12-09
#3181 Fix post_parent returning early if set to 0

### LINKED ISSUE #3181: pre_get_posts to only show top level CPT is not working while ep_integrate is true
### Describe your question

I have a CPT called 'food'. On the main `is_post_type_archive( 'food' )` I'm trying to show only top level posts. 

In `pre_get_posts` I have `$query->set( 'post_parent', 0 );`. It only works when I add `$query->set( 'ep_integrate', false );`.

I've re-indexed but still the same issue. Am I missing something else to use `pre_get_posts` with a core archive query and EP?

Thanks :)

### Code of Conduct

JiveDig:
### Description of the Change
This change changes `empty( $args['post_parent'] )` to `! isset( $args['post_parent'] )` so any query or `pre_get_posts` (or similar) filter that sets `post_parent` to `0` will still work as expected.

Closes #3181

### How to test the Change
As seen in #3181, use a filter like the following on a hierarchical CPT to confirm the confirm child posts are actually removed.

```
add_action( 'pre_get_posts', function( $query ) {
	if ( is_admin() ) {
		return;
	}

	if ( ! $query->is_main_query() ) {
		return;
	}

	if ( ! $query->is_post_type_archive( 'things' ) ) {
		return;
	}

	$query->set( 'post_parent', 0 );
});
```

### Changelog Entry

> Fixed - Queries with `post_parent` set to `0` not working correctly

### Credits

Props @JiveDig

### Checklist:

- [] I have updated the documentation accordingly.

### DIFF
--- a/includes/classes/Indexable/Post/Post.php
+++ b/includes/classes/Indexable/Post/Post.php
@@ -1720,7 +1720,7 @@ class Post extends Indexable {
 	 * @return array
 	 */
 	protected function parse_post_parent( $args ) {
-		if ( empty( $args['post_parent'] ) || 'any' === strtolower( $args['post_parent'] ) ) {
+		if ( ! isset( $args['post_parent'] ) || 'any' === strtolower( $args['post_parent'] ) ) {
 			return [];
 		}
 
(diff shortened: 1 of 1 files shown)


felipeelia:
Hey @JiveDig, thanks for the PR! It seems I was wrong and `isset()` was not enough to cover all possible cases. I've opened #3211 with a slight change in your code so we can account for both scenarios. I'm closing this PR and will merge the other one as soon as the tests pass.

PR #3206, opened 2022-12-14
Stop the Sync process if mapping failed

### LINKED ISSUE #3204: Stop the Sync process if mapping failed
### Describe the bug

Currently, if the mapping fails due to an error, the sync process continues and ends with a success message.

This is not ideal, and we should halt the sync operation and display the error message as soon as mapping fails.

### Steps to Reproduce

1. Add this filter `add_filter( 'ep_config_mapping_request', '__return_false' );` to force the mapping failed
2. Try to sync from dashboard

_No response_

### Environment information

_No response_

_No response_

### Code of Conduct

burhandodhy:
### Description of the Change

This PR updates the `put_mapping` method to return the `WP_ERROR` object in case there is issue while doing the mapping. The main reason to doing this is so we can stop the sync if there is an error and also shows the error message in dashboard and WP CLI command

Closes #3204 

### How to test the Change

1. Add this filter to enforce the mapping fail 
```
add_filter(
	'ep_config_mapping_request',
	function() {
		return new \WP_Error( 'test', 'This was forced to fail' );
	}
);
```
2. Run the sync from dashboard and make sure the sync stopped and the message is logged
3. Run `put-mapping` command and it should display the error message

### Changelog Entry

> Changed - put-mapping wpcli returns error message if mapping failed
> Fixed - Stopped Sync if put mapping throws error.

### Credits

Props @burhandodhy @felipeelia 

### Checklist:

### DIFF
--- a/includes/classes/Command.php
+++ b/includes/classes/Command.php
@@ -287,7 +287,13 @@ class Command extends WP_CLI_Command {
 						WP_CLI::success( esc_html__( 'Mapping sent', 'elasticpress' ) );
 					} else {
 						/* translators: Error message */
-						WP_CLI::error( sprintf( esc_html__( 'Mapping failed: %s', 'elasticpress' ), $result->get_error_message() ), false );
+						WP_CLI::error(
+							sprintf(
+								esc_html__( 'Mapping failed: %s', 'elasticpress' ),
+								$result->get_error_message()
+							),
+							false
+						);
 
 						return false;
 					}
@@ -324,7 +330,13 @@ class Command extends WP_CLI_Command {
 					WP_CLI::success( esc_html__( 'Mapping sent', 'elasticpress' ) );
 				} else {
 					/* translators: Error message */
-					WP_CLI::error( sprintf( esc_html__( 'Mapping failed: %s', 'elasticpress' ), $result->get_error_message() ), false );
+					WP_CLI::error(
+						sprintf(
+							esc_html__( 'Mapping failed: %s', 'elasticpress' ),
+							$result->get_error_message()
+						),
+						false
+					);
 
 					return false;
 				}
@@ -362,7 +374,13 @@ class Command extends WP_CLI_Command {
 				WP_CLI::success( esc_html__( 'Mapping sent', 'elasticpress' ) );
 			} else {
 				/* translators: Error message */
-				WP_CLI::error( sprintf( esc_html__( 'Mapping failed: %s', 'elasticpress' ), $result->get_error_message() ), false );
+				WP_CLI::error(
+					sprintf(
+						esc_html__( 'Mapping failed: %s', 'elasticpress' ),
+						$result->get_error_message()
+					),
+					false
+				);
 
 				return false;
 			}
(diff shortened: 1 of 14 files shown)


includes/classes/Command.php
```diff
@@ -283,10 +283,11 @@ private function put_mapping_helper( $args, $assoc_args ) {
...
 
-					if ( $result ) {
+					if ( ! is_wp_error( $result ) ) {
 						WP_CLI::success( esc_html__( 'Mapping sent', 'elasticpress' ) );
 					} else {
-						WP_CLI::error( esc_html__( 'Mapping failed', 'elasticpress' ), false );
+						/* translators: Error message */
+						WP_CLI::error( sprintf( esc_html__( 'Mapping failed: %s', 'elasticpress' ), $result->get_error_message() ), false );
 
 						return false;
```
felipeelia: To make it more readable, can we split these functions into different lines?
```suggestion
						/* translators: Error message */
						WP_CLI::error(
							sprintf(
								esc_html__( 'Mapping failed: %s', 'elasticpress' ),
								$result->get_error_message()
							)
						);
```

includes/classes/Elasticsearch.php
```diff
@@ -810,7 +810,7 @@ public function create_network_alias( $indexes, $network_alias ) {
 	 * @param  string $index Index name.
 	 * @param  array  $mapping Mapping array.
 	 * @since  3.0
-	 * @return boolean
+	 * @return boolean|WP_ERROR
```
felipeelia: The type here is actually `WP_Error`, isn't it?

includes/classes/Elasticsearch.php
```diff
@@ -810,7 +810,7 @@ public function create_network_alias( $indexes, $network_alias ) {
...
 	 */
 	public function put_mapping( $index, $mapping ) {
```
felipeelia: As I said in slack, we can't simply change the return and introduce a breaking change. Let's please add an optional third parameter where devs can choose between receiving `false` or the actual error.

felipeelia (changes requested):

includes/classes/Elasticsearch.php
burhandodhy: @felipeelia  I made the changes as you requested.

includes/classes/IndexHelper.php
```diff
@@ -1208,13 +1210,20 @@ protected function on_error_update_and_clean( $error ) {
...
+		switch ( $context ) {
+			case 'mapping':
 				/* translators: Error message */
-				esc_html__( 'Index failed: %s', 'elasticpress' ),
-				$error['message']
-			)
-		);
+				$message = sprintf( esc_html__( 'Mapping failed: %s', 'elasticpress' ), $error['message'] );
+				/* translators: Line break */
+				$message .= sprintf( esc_html__( '%sMapping has failed, which will cause ElasticPress search results to be incorrect. Please click `Delete all Data and Start a Fresh Sync` to retry mapping.', 'elasticpress' ), "\n" );
```
felipeelia:
```suggestion
				$message  = sprintf( esc_html__( 'Mapping failed: %s', 'elasticpress' ), $error['message'] );
				$message .= "\n";
				$message .= esc_html__( 'Mapping has failed, which will cause ElasticPress search results to be incorrect. Please click `Delete all Data and Start a Fresh Sync` to retry mapping.', 'elasticpress' );
```

felipeelia (changes requested):

PR #3222, opened 2022-12-20
Fix: Prevent using the ES for coupon search queries

### LINKED ISSUE #3221: BUG: Search Coupon Query go through ElasticSearch
### Describe the bug

Currently we have a `blacklist_coupons` function which restrict the query to go through ElasticSearch if the post type is `shop_coupon`. However, the functionality isn't working as expected.  

The main problem is that the `ep_elasticpress_enabled` filter in WooCommerce class is override with the `Search` class  https://github.com/10up/ElasticPress/blob/develop/includes/classes/Feature/Search/Search.php#L99because filter because of the same filter priority 

### Steps to Reproduce

1. Create the `test-coupon`.
2. Search for that coupon via code 

```
add_action( 'init', function() : void {

		$query = new \WP_Query( [
			'post_type' => 'shop_coupon',
			's' => 'test-coupon',
		] );

		var_dump( $query->posts );
}, 999  );

```
3. It will return no result because the query go through ES.

_No response_

### Environment information

_No response_

_No response_

### Code of Conduct

burhandodhy:
### Description of the Change

This PR fixes the issue where the search queries for coupon always go through the ES

Closes #3221

### Changelog Entry

> Fixed - Prevent search queries for coupon to use Elasticsearch

### Credits

Props @burhandodhy 

### Checklist:

### DIFF
--- a/includes/classes/Feature/WooCommerce/WooCommerce.php
+++ b/includes/classes/Feature/WooCommerce/WooCommerce.php
@@ -626,7 +626,11 @@ class WooCommerce extends Feature {
 	 * @return bool
 	 */
 	public function blacklist_coupons( $enabled, $query ) {
-		if ( method_exists( $query, 'get' ) && 'shop_coupon' === $query->get( 'post_type' ) ) {
+		if ( is_admin() ) {
+			return $enabled;
+		}
+
+		if ( 'shop_coupon' === $query->get( 'post_type' ) && empty( $query->query_vars['ep_integrate'] ) ) {
 			return false;
 		}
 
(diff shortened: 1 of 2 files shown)


includes/classes/Feature/WooCommerce/WooCommerce.php
```diff
@@ -626,7 +626,13 @@ public function get_admin_searchable_post_types() {
 	 * @return bool
 	 */
 	public function blacklist_coupons( $enabled, $query ) {
-		if ( method_exists( $query, 'get' ) && 'shop_coupon' === $query->get( 'post_type' ) ) {
+
```
felipeelia: Can we remove this empty line, please?

includes/classes/Feature/WooCommerce/WooCommerce.php
```diff
@@ -626,7 +626,13 @@ public function get_admin_searchable_post_types() {
...
+		// return early if we're in the admin.
+		if ( is_admin() ) {
+			return $enabled;
+		}
+
+		if ( method_exists( $query, 'get' ) && 'shop_coupon' === $query->get( 'post_type' ) && empty( $query->query_vars['ep_integrate'] ) ) {
```
felipeelia: As we are using ep_integrate_search_queries now and that runs inside integrate_search_queries:
1. We can be sure $query is a WP_Query object and that class has the get method since WP 1.5. We don't need to check if that exists.
2. integrate_search_queries already deals with ep_integrate, and we shouldn't test it here again.

felipeelia (changes requested):

includes/classes/Feature/WooCommerce/WooCommerce.php
burhandodhy: The check for the WP_Query is removed. However, the `ep_integrate` check is needed to handle the case where user explicitly set the `ep_integrate` to false in query. 

```
$query = new \WP_Query(
			[
				'post_type'    => 'shop_coupon',
				's'            => 'test-coupon',
				'ep_integrate' => false,
			]
		)
```

PR #3257, opened 2023-01-11
Filter boolean flags strictly 

### LINKED ISSUE #3199: BUG: Evaluate other assoc_args as booleans
### Describe the bug

With #3185 we can validate values strictly. There are other areas of improvement to implement this for example:

	public function get_last_cli_sync( $args, $assoc_args ) {

		$last_sync = Utils\get_option( 'ep_last_cli_index', array() );

		if ( isset( $assoc_args['clear'] ) ) {
			Utils\delete_option( 'ep_last_cli_index' );
		}

		$this->pretty_json_encode( $last_sync, $this->filter_boolean( $assoc_args['pretty'] ) );
	}

As of now, the original problem we were solving for the pretty flag seems to be true for other scenarios like this. 
		

### Steps to Reproduce

1.- Pass any value to the --clear flag for get-last-cli-sync
2.- It should delete the ep_last_cli_index option

_No response_

### Environment information

_No response_

_No response_

### Code of Conduct

oscarssanchez:
### Description of the Change

This PR changes the method filter_boolean() to is_allowed_flag_value() to be in line with the requirements from: https://github.com/10up/ElasticPress/issues/3199#issuecomment-1362868692

This method should account for future restricting values outside of booleans.

Closes https://github.com/10up/ElasticPress/issues/3199

### Changelog Entry

> Added - New feature
> Changed - filter_boolean() to is_allowed_flag_value() 
> Removed - Feature
> Fixed - Evaluate booleans strictly
> Security - Vulnerability

### Credits

Props @oscarssanchez  , @felipeelia  ...

### Checklist:


includes/utils.php
```diff
@@ -531,7 +531,7 @@ function update_option( $option, $value, $autoload = null ) {
  * @since 3.6.0
  * @param string $option        Name of the option to get.
  * @param mixed  $default_value Default value.
- * @return bool
+ * @return bool|array
```
felipeelia, line 534: Good catch, @oscarssanchez, thanks. This should actually be `mixed`, to match [get_option()](https://developer.wordpress.org/reference/functions/get_option/) docs though.

includes/classes/Command.php
```diff
@@ -170,8 +170,20 @@ public function deactivate_feature( $args, $assoc_args ) {
 	 * @param array $assoc_args Associative CLI args.
 	 */
 	public function list_features( $args, $assoc_args ) {
+		$defaults = [
+			'all' => false,
+		];
```
felipeelia, line 175: Perhaps a little late @oscarssanchez but what is the exact problem we are trying to solve with these changes? I gave a look into how other plugins check flag values and it seems the best option is to use `get_flag_value()`, as in `\WP_CLI\Utils\get_flag_value( $assoc_args, $flag, $default = null );`. Is that something you already considered?

Some links I found:
- [The get_flag_value function](https://github.com/wp-cli/wp-cli/blob/1e9ebdf4eb19dbe6f31a4f696f7b210056a8028c/php/utils.php#L949)
- [Usage in WooCommerce](https://github.com/woocommerce/woocommerce/blob/0654c4a7c63728e9c8186830e71cd95d4487a6c2/plugins/woocommerce/includes/cli/class-wc-cli-com-command.php#L150)
- [Usage in a WP-CLI Command itself](https://github.com/wp-cli/entity-command/blob/39d420e0cd660eb353f7034a71d3d4e21395bfe4/src/Taxonomy_Command.php#L152)

felipeelia (changes requested):

PR #3273, opened 2023-01-25
Prevent using the EP if query has unsuported orderby

### LINKED ISSUE #3272: Prevent sending requests to ES if orderby is set to post__in

Elasticsearch only allows asc or desc for the sorting, and if the WordPress query has `orderby` to `post__in`; the ES query ends with the error `No mapping found for [post__in] in order to sort on`.  

As the query never ends successfully, to prevent this error, the suggested solution is to never send the request to ES if the `orderby` is set to `post__in`. 

Related issue #773

### Designs

_No response_

_No response_

### Code of Conduct

burhandodhy:
### Description of the Change

This PR prevents WordPress to use Elasticsearch if the Query does have unsupported orderby value

Closes #3272 

### How to test the Change

1. Set the orderby => 'post__in' in WP_Query.
2. Check the query should not use Elasticsearch.

### Changelog Entry

> Changed - Prevent using the Elasticsearch if query has unsupported orderby

### Credits

Props @burhandodhy 

### Checklist:

### DIFF
--- a/includes/classes/Feature/Search/Search.php
+++ b/includes/classes/Feature/Search/Search.php
@@ -807,7 +807,14 @@ class Search extends Feature {
 			'parent',
 		];
 
-		if ( in_array( $query->get( 'orderby' ), $unsupported_orderby, true ) ) {
+		$orderby = $query->get( 'orderby' );
+		if ( is_array( $orderby ) ) {
+			$orderby = array_keys( $orderby );
+		} else {
+			$orderby = explode( ' ', $orderby );
+		}
+
+		if ( array_intersect( $orderby, $unsupported_orderby ) ) {
 			return true;
 		}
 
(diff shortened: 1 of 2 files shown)


felipeelia (changes requested):
@burhandodhy WP_Query's `orderby` also accepts an array and we need to account for that as well. Instead of checking
```php
if ( in_array( $query->get( 'orderby' ), $unsupported_orderby, true ) ) {
```
we could do something like
```php
if ( array_intersect( (array) $query->get( 'orderby' ), $unsupported_orderby ) ) {
```

includes/classes/Feature/Search/Search.php
```diff
@@ -786,4 +787,37 @@ public function save_exclude_from_search_meta( $post_id, $post ) {
...
 		update_post_meta( $post_id, 'ep_exclude_from_search', $exclude_from_search );
 	}
+
+	/**
+	 * If WP_Query has unsupported orderby, skip ES query integration and use the WP query instead.
+	 *
+	 * @param bool      $skip Whether to skip ES query integration
+	 * @param \WP_Query $query WP_Query object
+	 * @since 4.5
+	 */
```
felipeelia: @burhandodhy can we please document the function return as well?

felipeelia (changes requested):

PR #3307, opened 2023-02-08
New X-ElasticPress-Request-ID header

### LINKED ISSUE #3305: New HTTP Header in all Requests

To better keep track of requests being sent to Elasticsearch, let's add a new HTTP header called `X-Request-Id` with a value equal to a `wp_generate_uuid4()` call prefixed by the endpoint name. This should be added to both backend and frontend requests (autosuggest and Instant Results.)

### Designs

_No response_

_No response_

### Code of Conduct

felipeelia:
### Description of the Change

Closes #3305

### Changelog Entry

> Added - Requests now have a new header called X-ElasticPress-Request-ID to help with debugging.

### Credits

Props @felipeelia, @dustinrue 

### Checklist:

### DIFF
--- a/assets/js/api-search/src/hooks.js
+++ b/assets/js/api-search/src/hooks.js
@@ -38,8 +38,9 @@ export const useFetchResults = (apiHost, apiEndpoint, Authorization, requestIdBa
 			Authorization,
 		};
 
-		if (requestIdBase) {
-			headers['X-ElasticPress-Request-ID'] = generateRequestId(requestIdBase);
+		const requestId = generateRequestId(requestIdBase);
+		if (requestId) {
+			headers['X-ElasticPress-Request-ID'] = requestId;
 		}
 
 		request.current = fetch(url, {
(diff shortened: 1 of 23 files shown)


assets/js/api-search/src/hooks.js
```diff
@@ -3,15 +3,21 @@
  */
 import { useCallback, useRef } from '@wordpress/element';
 
+/**
+ * Internal dependencies.
```
JakePT, line 7: I've been trying to keep `api-search` as self-contained as possible, because I can imagine breaking it out into its own library some day, so I don't love that we're importing this function from outside the api-search directory.

Another approach could be to support passing `requestHeaders` to `ApiSearchProvider`, then that could be defined as a function outside of `api-search` that generates a request ID header.

Does that make sense?

This doesn't need to be a blocker, as api-search is not currently publicly exposed or available as a library so we can always change it later if necessary.

JakePT (approved):

felipeelia:
I am merging this as is, after a conversation with Jacob via slack. The main reason why the `generateRequestId` isn't in the API is that it is also used by autosuggest, which does not use an API-driven approach yet. When it comes, we can bring that function in, and fix the API encapsulation again.

PR #3310, opened 2023-02-10
Add a setting to enable Orders Autosuggest

### LINKED ISSUE #3277: Add toggle to WooCommerce feature for enabling orders autosuggest

- Add an option to enable/disable Orders Autosuggest inside the WooCommerce Feature box
- Introduce logic to do a full sync after enabling it
- Introduce logic to sync orders post type (Sync it even if Protected Content is disabled)

### Designs

_No response_

_No response_

### Code of Conduct

JakePT:
### Description of the Change
Adds a setting to enable Orders Autosuggest. The setting is disabled for non-EPIO users, with a filter to enable it.

Enabling Orders Autosuggest will also enable indexing of Orders, regardless of whether the Protected Content feature is enabled. Therefore a reindex is required when enabling the feature. This includes a change to allow requiring a reindex if a particular feature setting is changed.

Closes #3277

### How to test the Change
- Disable Protected Content feature.
- Enable Orders Autosuggest from the WooCommerce feature. You should be warned that a reindex will be required.
- Visit _WooCommerce > Orders_ and start searching for orders. Suggestions should be offered in a listbox below the search form.
- If not using EP.io the option to enable the feature should be disabled, with an explanatory message.

### Changelog Entry
See #3175

### Credits
@JakePT @felipeelia 

### Checklist:


anjulahettige:
The option will be grayed out to non-EPio users with the following message: 
`Due to the sensitive nature of orders, this autosuggest feature is available only to ElasticPress.io customers. (Learn More)`

JakePT:
@anjulahettige @felipeelia @brandwaffle I've pushed a change with the message, but I still need a URL for the Learn more link.

PR #3313, opened 2023-02-12
Add custom capability for managing ElasticPress

tlovett1:
This adds a new capability for managing ElasticPress and adds it to the admin role.

The point of this change is it allows people to add the capability to non-admins. Right now, that's not possible since everything goes off of `manage_options`.

### Checklist:


tott:
@tlovett1 this is a great suggestion and we will make sure this gets rolled into the 4.5 release currently scheduled for march. We will want to do some changes to the suggested code to ensure backward compatibility and performance integrity.

tlovett1:
Thanks @tott

PR #3319, opened 2023-02-14
Add a download report button

### LINKED ISSUE #3288: Add a new "Download" button in the Status Report page

As the status report copied content can get lengthy, it would be good to provide a way so people can actually get a txt file, making things easier to share with support.

### Designs

After talking to @brandwaffle, we want to make this new button the primary one (blue/primary color background, etc.)

_No response_

### Code of Conduct

felipeelia:
### Description of the Change

Closes #3288

### How to test the Change
1. Go to the Status Report page
2. Click on the Download report button

### Changelog Entry

> Added - New "Download report" button in the Status Report page

### Credits

Props @felipeelia 

### Checklist:


assets/css/status-report.css
```diff
@@ -32,6 +32,10 @@
 	align-items: center;
 	display: inline-flex;
 	margin: 0.5rem 0 1rem;
+
+	& .button + .button {
+		margin-left: 0.5rem;
```
JakePT: I think rather than this, it would be better to change `.ep-copy-button-wrapper` so that it's no longer `flex`. This way the buttons inside will display properly as inline and there'll be a natural space between them.

JakePT (changes requested):

assets/css/status-report.css
felipeelia: That works @JakePT but then the spaces aren't even between spaces and the "Copied!" text:

image

If we remove the space between the button and the text, it seems things will be too close in general:

image

What would you prefer?

JakePT: @felipeelia Honestly, in playing with it I think I prefer the version in your first screenshot. That's just my preference though, I won't hold this up for it.

JakePT (dismissed):

assets/css/status-report.css
anjulahettige: This was the approved version:
image

felipeelia: Changed in 5680a98

JakePT (approved):

PR #3301, opened 2023-02-07
Swap filter bool functions

oscarssanchez:
### Description of the Change

Closes #3199

### Changelog Entry

> Changed - Filter command flags using get_flag_value()

### Credits

Props @oscarssanchez , @felipeelia 

### Checklist:

- [x] I agree to follow this project's [**Code of Conduct**](https://github.com/10up/.github/blob/trunk/CODE_OF_CONDUCT.md).
- [ ] I have updated the documentation accordingly.
- [ ] I have added tests to cover my change.
- [x] All new and existing tests pass.

PR #3320, opened 2023-02-16
Fix: All public methods in Command class are treated as a wpcli command 

### LINKED ISSUE #3193: BUG: All public methods in Command class are treated as a wpcli command
### Describe the bug

The `Command` class has a few methods whose visibility is public, and WP CLI treated these methods as a command. For example, if we run `wp elasticpress custom_get_transient` , WordPress treats this [method](https://github.com/10up/ElasticPress/blob/4.0.0/includes/classes/Command.php#L1214) as a command and eventually ends with the errors. 

### Steps to Reproduce

Run WP CLI command `wp elasticpress custom_get_transient`

_No response_

### Environment information

_No response_

_No response_

### Code of Conduct

burhandodhy:
### Description of the Change

This PR adds a new Helper class for WP CLI commands and move the public method to Helper class so that they will be not treated as CLI command

Closes #3193 

### Changelog Entry

> Fixed - All public methods in Command class are treated as a wpcli command

### Credits

Props @burhandodhy 

### Checklist:


includes/classes/Command.php
```diff
@@ -16,6 +16,7 @@
 use ElasticPress\Utils as Utils;
 use ElasticPress\Elasticsearch as Elasticsearch;
 use ElasticPress\Indexables as Indexables;
+use ElasticPress\Command\Utility as Utility;
```
felipeelia: @burhandodhy you don't need to repeat the class name if you are keeping the same. The two lines above the new one have that same problem. `use ElasticPress\Command\Utility` is already enough.

includes/classes/Command.php
```diff
@@ -663,11 +664,8 @@ private function create_network_alias_helper( Indexable $indexable ) {
 	 * @since  3.3
 	 */
 	public function delete_transient_on_int( $signal_no ) {
-		if ( SIGINT === $signal_no ) {
-			$this->delete_transient();
-			WP_CLI::log( esc_html__( 'Indexing cleaned up.', 'elasticpress' ) );
-			WP_CLI::halt( 0 );
-		}
+		_doing_it_wrong( __METHOD__, esc_html__( 'This method is deprecated and will be removed from 5.0. Use \ElasticPress\Command\Utility::delete_transient_on_int instead.', 'elasticpress' ), '4.5.1' );
```
felipeelia: Instead of `_doing_it_wrong` we should call `_deprecated_function` in all these functions. Also, do you mind reviewing the last parameter of the function? You have it as `'4.5.1'` but that should be 4.5.0.

includes/classes/Command/Utility.php
```diff
@@ -0,0 +1,193 @@
+<?php
+/**
+ * ElasticPress CLI Utility
+ *
+ * @since 4.4.0
```
felipeelia: This should be 4.5.0

includes/classes/Command/Utility.php
```diff
@@ -0,0 +1,193 @@
...
+
+/**
+ * Utility class for WP CLI commands.
+ */
+class Utility {
+
+	/**
+	 * Internal timer.
+	 *
+	 * @since 4.2.0
```
felipeelia: Although these methods exist before, the `@since` tag should be either removed or updated to 4.5.0, as these are not the (exact) same methods. Let's remove them for now, as all of them are `4.5.0`

felipeelia (changes requested):
@burhandodhy asked for some changes on this one. Although I'm not a fan of introducing a whole new class to fix the original problem, I also see it being valid to separate concerns... 🤔

includes/classes/Command.php
burhandodhy: I followed the same convention. However, I removed the alias from all the classes.

includes/classes/Command.php
burhandodhy: Its done.

tests/php/TestCommands.php
```diff
@@ -12,6 +12,7 @@
 use ElasticPress\Command;
 use ElasticPress\Indexables;
 use ElasticPress\Utils;
+use ElasticPress\Command\Utility as Utility;
```
felipeelia: @burhandodhy we don't need the alias here as well

felipeelia (changes requested):
In addition to the small change that still needs to be done here @burhandodhy, do you mind fixing the git conflict that arose? Thanks.

tests/php/TestCommands.php
burhandodhy: Done

includes/classes/Command.php
```diff
@@ -723,9 +721,7 @@ public function delete_transient_on_int( $signal_no ) {
 	 * @param array $assoc_args Associative CLI args.
 	 */
 	public function sync( $args, $assoc_args ) {
-		global $wp_actions;
-
-		$setup_option = \WP_CLI\Utils\get_flag_value( $assoc_args, 'setup', false );
+		$setup_option = isset( $assoc_args['setup'] ) ? $assoc_args['setup'] : false;
```
felipeelia: @burhandodhy this actually should keep the change introduced in #3301

includes/classes/Command.php
```diff
@@ -821,11 +817,11 @@ public function sync( $args, $assoc_args ) {
 
 		\ElasticPress\IndexHelper::factory()->full_index( $index_args );
 
-		remove_action( 'ep_sync_put_mapping', [ $this, 'stop_on_failed_mapping' ] );
-		remove_action( 'ep_sync_put_mapping', [ $this, 'call_ep_cli_put_mapping' ], 10, 2 );
-		remove_action( 'ep_index_batch_new_attempt', [ $this, 'should_interrupt_sync' ] );
+		remove_action( 'ep_sync_put_mapping', [ Utility::class, 'stop_on_failed_mapping' ] );
+		remove_action( 'ep_sync_put_mapping', [ Utility::class, 'ep_cli_put_mapping' ], 10, 2 );
+		remove_action( 'ep_index_batch_new_attempt', [ Utility::class, 'maybe_interrupt_sync' ] );
```
felipeelia: Are you sure about these two method names? Shouldn't they be `call_ep_cli_put_mapping` and `should_interrupt_sync`?

includes/classes/Command.php
```diff
@@ -1135,15 +1124,11 @@ function ( $prefix ) use ( $assoc_args ) {
...
-		if ( $should_interrupt_sync ) {
-			WP_CLI::line( esc_html__( 'Sync was interrupted', 'elasticpress' ) );
-			$this->delete_transient_on_int( 2 );
-			WP_CLI::halt( 0 );
-		}
+		_deprecated_function( __METHOD__, '4.5.0', '\ElasticPress\Command\Utility::should_interrupt_sync' );
+		Utility::should_interrupt_sync();
 	}
 
+
```
felipeelia: It seems a new line was introduced here by accident. Do you mind removing it?

includes/classes/Command.php
```diff
@@ -1337,7 +1297,7 @@ public function index_output( $message, $args, $index_meta, $context ) {
 				break;
 
 			case 'error':
-				$this->clear_sync();
+				Utility::delete_transient();
```
felipeelia: Making this change we are no longer calling the `ep_cli_before_clear_index` and `ep_cli_after_clear_index` actions. Is there a reason why?

felipeelia (changes requested):
@burhandodhy the conflict was not properly fixed (the new code was reverted.) So I took a deeper look into the code and it seems we still have some inconsistencies and breaking changes. Do you mind giving it another thorough review? Are you 100% sure this does not break any existing behavior including hooks? Thanks.

tests/php/TestCommands.php
felipeelia: It seems this wasn't changed @burhandodhy ?

felipeelia (changes requested):
@burhandodhy we are almost there. There is a change that didn't seem to be pushed and also I just remembered we need to change the --remove parameter [here](https://github.com/10up/ElasticPress/blob/develop/.github/workflows/build-docs.yml#L49).

PR #3324, opened 2023-02-20
Ngram fields for orders autosuggest

### LINKED ISSUE #3281: Implement partial word search and highlighting for orders autosuggest

- Add a new field concatenating order number, email, and name (and use it to display the list item)
- Add new ngram field for the above

### Designs

Screenshot 2023-01-28 at 2 15 16 am

_No response_

### Code of Conduct

felipeelia:
### Description of the Change

This PR introduces the code needed so orders autosuggest start using ngram fields. As autosuggest and the orders autosuggest have some intersection between fields and mapping changes, the common code was moved to the indexable and post indexable classes.

It is done by adding a `term_suggest` field to all orders, filled with all fields we want to be searchable.

Closes #3281 

### Changelog Entry

Add this to the orders autosuggest PRs list

### Credits

Props @felipeelia 

### Checklist:

### DIFF
--- a/includes/classes/Feature/WooCommerce/Orders.php
+++ b/includes/classes/Feature/WooCommerce/Orders.php
@@ -440,7 +440,7 @@ class Orders {
 	 * @return array
 	 */
 	public function filter_term_suggest( $post_args ) {
-		if ( 'shop_order' !== $post_args['post_type'] ) {
+		if ( empty( $post_args['post_type'] ) || 'shop_order' !== $post_args['post_type'] ) {
 			return $post_args;
 		}
 
--- a/includes/classes/Feature/WooCommerce/Orders.php
+++ b/includes/classes/Feature/WooCommerce/Orders.php
@@ -40,6 +40,9 @@ class Orders {
 		add_filter( 'ep_after_sync_index', [ $this, 'epio_save_search_template' ] );
 		add_filter( 'ep_saved_weighting_configuration', [ $this, 'epio_save_search_template' ] );
 		add_action( 'rest_api_init', [ $this, 'rest_api_init' ] );
+		add_filter( 'ep_post_sync_args', [ $this, 'filter_term_suggest' ], 10 );
+		add_filter( 'ep_post_mapping', [ $this, 'mapping' ] );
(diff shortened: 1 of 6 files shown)


includes/classes/Feature/WooCommerce/Orders.php
```diff
@@ -429,4 +432,96 @@ public function check_token_permission() {
...
+		 */
+		$post_args['meta']['order_post_id'][] = [
+			'raw'   => $post_args['ID'],
+			'value' => $post_args['ID'],
+		];
+
+		$suggest = [];
+
+		$fields_to_ngram = [
+			'_order_key',
```
JakePT: `_order_key` is the hidden value used in the URL for access to the order status page, and isn't publicly exposed, so it's not the field we'll want to search. The order number is the post ID by default, but should be accessed with `$order->get_order_number()` for compatibility with plugins like https://wordpress.org/plugins/woocommerce-sequential-order-numbers/. We don't seem to be currently indexing this, since it's not saved as meta by default.

includes/classes/Feature/WooCommerce/Orders.php
```diff
@@ -429,4 +432,96 @@ public function check_token_permission() {
...
+			'_billing_email',
+			'_billing_last_name',
+			'_billing_first_name',
+			'_shipping_first_name',
+			'_shipping_last_name',
```
JakePT: We're not currently displaying the shipping names, so do we need to use them for this?

JakePT (changes requested):

anjulahettige:
For now, we are just searching on the fields that are being displayed.

includes/classes/Feature/WooCommerce/Orders.php
felipeelia: I've added some code in d890c3c to use that instead @JakePT. Do you mind checking it again?

includes/classes/Feature/WooCommerce/Orders.php
felipeelia: Changed in d890c3c as well.

JakePT (approved):

PR #3333, opened 2023-02-21
Fix #3325 Set the current post when indexing posts

### LINKED ISSUE #3325: BUG: Blocks that use postId context crash indexing
### Describe the bug

1. I have a dynamic block that requires the `postId` context to render post specific content.
2. This context is only provided if the global `$post` object is set. This is fine because my block is only available in posts, and is inside a post.
3. ElasticPress does not set the current post so `$post` is undefined
4. ElasticPress uses `the_content` filter to retrieve my posts content
5. WP tries to render the block and the callback tries to use `postId` expecting an integer
6. the block renderer recieves a `null` and the application crashes

### Steps to Reproduce

Attempt to index a site that uses the `postId` in a block renderer

_No response_

### Environment information

_No response_

_No response_

### Code of Conduct

tomjn:
### Description of the Change

When indexing a post, the content gets rendered, and shortcodes/blocks/etc may rely on the global `$post` variable to provide context that will be missing when performing an ES index.

In my case it was a block which did not handle the value being missing, but context dependent content would generate PHP warnings/errors or incorrect data

By setting the global `$post` and the current postdata, we can avoid some of these situations. This enabled indexing to continue locally where it would fail previously.

Closes #3325

### How to test the Change

In my case, I did not check the `postId` in the block context for presence and used it without verification. This lead to a fatal error which is avoided when using this PR

### Changelog Entry

> Fixed - Not setting the post context when indexing a post

### Credits

Props @tomjn 

### Checklist:

### DIFF
--- a/includes/classes/Indexable/Post/Post.php
+++ b/includes/classes/Indexable/Post/Post.php
@@ -420,7 +420,9 @@ class Post extends Indexable {
 	 * @return bool|array
 	 */
 	public function prepare_document( $post_id ) {
+		global $post;
 		$post = get_post( $post_id );
+		setup_postdata( $post );
 
 		if ( empty( $post ) ) {
 			return false;


tomjn:
Making this PR via the web interface, I don't have a lot of bandwidth right now to dig into things

PR #3346, opened 2023-02-27
Use WP_Query argument for facetable queries.

JakePT:
### Description of the Change
The current facet blocks use the `ep_is_facetable` filter to enabled faceting for queries used to create the block previews. This can be unreliable if the preview query spawns any other queries as the aggregations from those queries may be used instead. By using an argument on the query instead faceting can be reliably enabled only on the specific queries needed.

### How to test the Change
- Face block previews should continue to work as expected.
- The Facet by Range block should no longer display the "Is this a numeric field?" error if a numeric field is selected.

### Changelog Entry
- Fixed - An issue where Facet block previews may be missing or inaccurate.

### Credits

Props @JakePT

### Checklist:

### DIFF
--- a/includes/classes/Feature/Facets/Facets.php
+++ b/includes/classes/Feature/Facets/Facets.php
@@ -248,6 +248,10 @@ class Facets extends Feature {
 			return true;
 		}
 
+		if ( $query->get( 'ep_is_facetable' ) ) {
+			return true;
+		}
+
 		if ( is_admin() || is_feed() ) {
 			return false;
 		}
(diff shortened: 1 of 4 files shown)


felipeelia:
Although I LOVE the idea of making it a WP_Query level thing, I guess that would be a much bigger change. Introducing this feature as is, would mislead developers to believe ~aggregations~ facets would also be set in a WP_Query level.

For the specific problem we are trying to fix with this, I got https://github.com/10up/ElasticPress/tree/feature/3291-2 working, if you can test, @JakePT .

PR #3348, opened 2023-02-27
Make failed queries notice dismissible

### LINKED ISSUE #3322: Make ElasticPress queries failed notices dismissible

As of now, these notices are not dismissible, they can become disturbing to users if they for example, are aware of them and actively working them, but want to get rid of them in the meantime for the editorial team etc.

### Designs

_No response_

_No response_

### Code of Conduct

oscarssanchez:
### Description of the Change
This PR makes the status report notice dismissible. 

Closes #3322

### How to test the Change
1.- Trigger a failed query
2.- Go to the admin dashboard
3.- Dismiss the notice
4.- Reload page. Notice should not re-appear
5.- Trigger another error
6.- Notice should re-appear

### Changelog Entry

> Fixed - Make notice dismissible

### Credits

Props @oscarssanchez, @felipeelia 

### Checklist:


includes/classes/QueryLogger.php
```diff
@@ -213,6 +214,10 @@ public function clear_logs() {
 	 * @return array
 	 */
 	public function maybe_add_notice( array $notices ) : array {
+		if ( ! current_user_can( 'manage_options' ) ) {
```
felipeelia: @oscarssanchez we now have a `\Utils\get_capability()` function that should be used here, instead of hardcoding `'manage_options'`. Do you mind changing it please?

felipeelia (changes requested):
@oscarssanchez Although changing the time limit and storing the time when the notice was dismissed works, I'd go with the simpler approach we agreed on during the EP sync: in `QueryLogger::update_logs()` delete the `ep_hide_has_failed_queries_notice` option right before updating the transient with the logged queries.

Did you try that and face any issues? If not, I'd prefer that approach as it'll be simpler to read.

PR #3355, opened 2023-03-01
Exclude dist directory from PHP linting

burhandodhy:
### Description of the Change

When we run the `composer lint-fix` command, the php linter tries to fix the PHP files in the `dist` directory. The `dist` directory is already in the gitignore so we don't need to check the linting issue there. 

Closes #

### How to test the Change

Run `composer lint-fix` command and it should not check the PHP files in the `dist` directory

### Changelog Entry

> Changed - Exclude dist directory from the PHP linting

### Credits

Props @burhandodhy 

### Checklist:


burhandodhy:
It appears that there is a problem with my dependencies. Removing the vendor and installing all dependencies resolved the problem. Closing this PR.

PR #3365, opened 2023-03-03
Facet by Meta Range E2E tests

### LINKED ISSUE #3357: Add E2E tests for the Facet by Meta Range block

ElasticPress 4.5.0 will include a new Facet by Meta Range block, but it still needs E2E tests.

### Designs

_No response_

_No response_

### Code of Conduct

JakePT:
### Description of the Change
Adds E2E tests for the Facet by Meta Range block.

Closes #3357

### How to test the Change
Tests should pass in Cypress and on GitHub.

### Changelog Entry
N/A

### Credits

Props @JakePT 

### Checklist:

### DIFF
--- a/tests/cypress/support/index.js
+++ b/tests/cypress/support/index.js
@@ -58,3 +58,16 @@ cy.elasticPress = {
 		},
 	},
 };
+
+/**
+ * Ignore ResizeObserver error.
+ *
+ * @see {@link https://stackoverflow.com/questions/49384120/resizeobserver-loop-limit-exceeded}
+ */
+Cypress.on('uncaught:exception', (err) => {
+	if (err.message?.includes('ResizeObserver loop limit exceeded')) {
+		return false;
+	}
+
+	return err;
+});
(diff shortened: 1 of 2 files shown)


tests/cypress/support/index.js
```diff
@@ -58,3 +58,16 @@ cy.elasticPress = {
...
+ *
+ * @see {@link https://stackoverflow.com/questions/49384120/resizeobserver-loop-limit-exceeded}
+ */
+Cypress.on('uncaught:exception', (err) => {
+	if (err.message?.includes('ResizeObserver loop limit exceeded')) {
+		return false;
+	}
+
+	return err;
+});
```
JakePT, line 73: @felipeelia I was running into this issue, which is described at the link. It was happening in the test before the block was even inserted, but the library used by the block does appear to use ResizeObserver, so it might be related. The problem is that it's in the library so I can't do anything about it, however the various answers at that link seem to suggest that ignoring the issue is safe. I'd be curious to hear your thoughts though.

felipeelia: Yeah, we are already using this solution for Instant Results: https://github.com/10up/ElasticPress/blob/develop/tests/cypress/integration/features/instant-results.cy.js#L241

I wonder if we should merge those two in a more generic function or something. Thoughts?

PR #3373, opened 2023-03-07
Fix: Undefined array key 'displayCount'

burhandodhy:
### Description of the Change

This PR fixes the issue where the Facet by taxonomy throws the error when its added as a classis widget.

Closes #

### How to test the Change

- Add Taxonym Facet in classic widget
- Go to frontend and see there is no warning message. 

### Changelog Entry

> Fixed -  Undefined array key 'displayCount' error for facet

### Credits

Props @burhandodhy 

### Checklist:

### DIFF
--- a/includes/classes/Feature/Facets/Types/Taxonomy/Renderer.php
+++ b/includes/classes/Feature/Facets/Types/Taxonomy/Renderer.php
@@ -48,7 +48,7 @@ class Renderer {
 			$instance,
 			[
 				'title'        => '',
-				'displayCount' => '',
+				'displayCount' => false,
 			]
 		);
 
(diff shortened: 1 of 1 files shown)


includes/classes/Feature/Facets/Types/Taxonomy/Renderer.php
```diff
@@ -47,7 +47,8 @@ public function render( $args, $instance ) {
 		$instance = wp_parse_args(
 			$instance,
 			[
-				'title' => '',
+				'title'        => '',
+				'displayCount' => '',
```
felipeelia: @burhandodhy the doc for the object attribute that will eventually receive this is set to "boolean". Do you mind changing its default to false instead of an empty string? Thanks

felipeelia (changes requested):

PR #3397, opened 2023-03-21
Facets on product category

### LINKED ISSUE #3360: Facet not working on category page.
### Describe your question

I have installed ElasticPress on my website and would like to utilize its facet feature. While it works fine on search results, I'm having difficulty getting it to work properly on category pages that contain multiple products. Specifically, when I select a color facet, such as navy, it redirects me to the homepage and adds "?ep_filter_pa_colour=navy" to the end of the URL.

For example, if I'm on the category page for long-sleeve products:

https://example.com/?product_cat=long-sleeve

and then select the navy color facet, I'm redirected to:

https://example.com/?ep_filter_pa_colour=navy

Is there a way to make ElasticPress facets work on category pages, or are they only designed to work on search result pages?

### Code of Conduct

MARQAS:
### Description of the Change
This PR will fix the issue where facets are not working while the pretty permalink is not being used. 

Closes #3360 

### How to test the Change

- Add facets
- Change the permalink structure to use plain permalinks
- Go to a product category page, click on a facet and notice that it is working as expected

### Changelog Entry

> Fixed - Facets not working on product category pages when pretty permalink is not being used

### Credits

Props @MARQAS 

### Checklist:

### DIFF
--- a/includes/classes/Feature/Facets/Facets.php
+++ b/includes/classes/Feature/Facets/Facets.php
@@ -508,9 +508,21 @@ class Facets extends Feature {
 	 * @since 3.6.0
 	 */
 	public function get_allowed_query_args() {
-		$default_args         = array( 's', 'post_type', 'orderby', 'cat', 'tag' );
-		$indexable_taxonomies = Indexables::factory()->get( 'term' )->get_indexable_taxonomies();
-		$args                 = array_merge( $default_args, $indexable_taxonomies );
+		$default_args = array( 's', 'post_type', 'orderby', 'cat' );
+		// Retrieve all registered query variables for public taxonomies
+		$taxonomies = get_taxonomies(
+			array(
+				'public' => true,
+			),
+			'objects'
+		);
+		$query_vars = array();
+		foreach ( $taxonomies as $taxonomy ) {
+			if ( $taxonomy->query_var ) {
+				$query_vars[] = $taxonomy->query_var;
+			}
+		}
+		$args = array_merge( $default_args, $query_vars );
 
 		/**
 		 * Filter allowed query args
--- a/includes/classes/Feature/Facets/Facets.php
+++ b/includes/classes/Feature/Facets/Facets.php
@@ -508,7 +508,9 @@ class Facets extends Feature {
 	 * @since 3.6.0
 	 */
 	public function get_allowed_query_args() {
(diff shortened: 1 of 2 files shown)


felipeelia (changes requested):
@MARQAS although this code will fix the immediate problem described in the original issue, it doesn't seem to be the optimal way to fix the problem. If you test the same thing but with the native category (instead of product category) you still have the same problem: the `?cat=<id>` will be removed.

That said, what do you think about fixing the problem in the get_allowed_query_args() method, adding all public taxonomies to the allowed args?

includes/classes/Feature/Facets/Facets.php
```diff
@@ -508,7 +508,9 @@ public function output_feature_box_long() {
 	 * @since 3.6.0
 	 */
 	public function get_allowed_query_args() {
-		$args = array( 's', 'post_type', 'orderby' );
+		$default_args         = array( 's', 'post_type', 'orderby', 'cat', 'tag' );
```
felipeelia: @MARQAS are you sure post tags' parameter is `tag`?

includes/classes/Feature/Facets/Facets.php
```diff
@@ -508,7 +508,9 @@ public function output_feature_box_long() {
...
+		$indexable_taxonomies = Indexables::factory()->get( 'term' )->get_indexable_taxonomies();
```
felipeelia: @MARQAS can you please elaborate on the reason why you are getting this from the Terms indexable? Is there a reason why you are not getting all public taxonomies as we've discussed?

includes/classes/Feature/Facets/Facets.php
```diff
@@ -508,7 +508,9 @@ public function output_feature_box_long() {
...
+		$args                 = array_merge( $default_args, $indexable_taxonomies );
```
felipeelia: @MARQAS I'm not sure this will achieve the result we need. One thing is the taxonomy slug, and a different thing is its parameter in the query string. Although they are the same in most cases, it is not 100% guaranteed.

felipeelia (changes requested):
@MARQAS can we please extend the tests suite to be sure this problem is really fixed? Thanks.

felipeelia:
closing in favor of #3422 (pretty much the same code + tests)

PR #3403, opened 2023-03-23
Refine ElasticPress blocks.

### LINKED ISSUE #3384: Refine ElasticPress blocks
ElasticPress currently features 5 blocks:

- Related Posts (ElasticPress)
- Search Comments (ElasticPress)
- Facet by Taxonomy (ElasticPress)
- Facet by Meta (ElasticPress)
- Facet by Meta Range - Beta (ElasticPress)

I think there's a number of things we can do to improve the quality of these blocks:

- [x] **Recategorize blocks.**
Move the blocks into an "ElasticPress" block category.
- [x] **Update block names.**
Remove "(ElasticPress)" from the block name. Consider updating the terminology of the Facet blocks. It's not really correct to say "Facet by", instead "Filter by Taxonomy Facet" would be more correct.
- [x] **Update controls for consistency with core.**
Update "ASC/DESC" radio buttons for sorting with a dropdown with friendlier options like "Newest to oldest".
- [x] **Support customisation.**
Add support for non-experimental [block features](https://developer.wordpress.org/block-editor/reference-guides/block-api/block-supports/):
  - [x] Alignment.
  - [x] Background, text, and link colors.
  - [x] Sticky.
  - [x] Spacing.
  - [x] Font size and line height.
- [x] **Consolidate the Facet blocks.**
Currently there is a bunch of duplicated information and functionality between these blocks which could be reduced by implementing them as variations of a single block.
- [x] **Add new icons.**
Currently the Related Posts block has no icon, so it uses the generic lego block, the search comments block has a generic search icon whose style is inconsistent with core search blocks, and the facet blocks all use the same icon that doesn't represent the difference between the list and range blocks. Unique icons with a consistent style would make the blocks appear more professional.
- [ ] **Support inner blocks.**
Add inner block support to the Facet blocks so that headings etc. can be added that will be automatically hidden when the facet is not available.

### Code of Conduct

### LINKED ISSUE #3489: Add E2E tests for block styles

#3403 implemented support for the following block features:

  - Color
    - Background
    - Link
    - Text
  - Position
    - Sticky
  - Spacing
    - Margin
    - Padding
  - Typography
    - Font size
    - Line height

We should add E2E tests to confirm that these continue to work as expected for each release.

The effect of these options is essentially the same for every block, as they simply add classes or a style attribute to the block wrapper. This means that it should be possible to write most of these tests once while using them for multiple blocks.

### Designs

_No response_

_No response_

### Code of Conduct

JakePT:
### Description of the Change
Introduces the following features, improvements, and refinements to our blocks:

- All blocks have been moved into an "ElasticPress" block category and the "(ElasticPress)" identifier has been removed from their names.
- Descriptions and keywords have been added to all blocks.
- All blocks now support the following features in themes that support them:
  - Color
    - Background
    - Link
    - Text
  - Position
    - Sticky
  - Spacing
    - Margin
    - Padding
  - Typography
    - Font size
    - Line height
 - The Facet blocks have been renamed from "Facet by..." to "Filter by...".
 - References to filtering by "Meta" have been replaced with "Metadata" for consistency with how [WordPress documentation](https://developer.wordpress.org/plugins/metadata/) and #3068 refer to it.
- The _Filter by Taxonomy_ and _Filter by Metadata_ blocks now share a single edit component to avoid code reuse.
- Any functionality shared by the _Filter by Metadata Range_ and the other _Filter by_ blocks has been moved into shared components.
- Custom block previews have been replaced with implementations of [`ServerSideRender`](https://developer.wordpress.org/block-editor/reference-guides/packages/packages-server-side-render/).
- REST API requests in edit components have been replaced with implementations of [`useSelect`](https://make.wordpress.org/core/2019/06/10/introducing-usedispatch-and-useselect/) and a custom [data store](https://developer.wordpress.org/block-editor/reference-guides/packages/packages-data/#registering-a-store). This eliminates repeated API requests for data that has already been retrieved, including filterable taxonomies, meta fields, and meta range previews.
- When adding _Filter by Taxonomy_ or _Filter by Metadata_ blocks, users are now prompted to select a taxonomy or meta field before a preview is shown. This is consistent with several core blocks and the existing functionality of the _Facet by Meta Range_ block.
- The labels for the block inspector panels containing controls for block attributes have all been changed to just "Settings" for consistency with core blocks that only have a single panel.
- Labels and options have been normalized to sentence case for consistency with core.
- The taxonomy options for the _Filter by Taxonomy_ block have been changed to singular and the slug has been included in parentheses. This makes the options consistent with the Facet options in Instant Results, works better with the "Filter by..." label, resolves an inconsistency in the letter case and plurality of WooCommerce taxonomies and product attributes, and more clearly distinguishes different taxonomies with the same name.
- Fixes an issue where the "Preview unavailable" message would not display for filters without values.

Closes #3384

### Changelog Entry

> Added - New feature
> Removed - Feature
> Fixed - Bug fix
> Security - Vulnerability

### Credits

Props @JakePT 

### Checklist:


includes/classes/Indexable.php
```diff
@@ -1210,7 +1210,7 @@ public function get_distinct_meta_field_keys( $blog_id = null ) {
 			$meta_keys = array_values( array_keys( $meta_fields ) );
 			sort( $meta_keys );
 		} catch ( \Throwable $th ) {
-			return new \Exception( 'Meta fields not available.', 0 );
+			throw new \Exception( 'Meta fields not available.', 0 );
```
JakePT, line 1214: @felipeelia This error wasn't being caught in the API endpoint because the API endpoint expected a thrown exception. As far as I'm aware exceptions should be thrown rather than returned, but let me know if there's a better fix.

JakePT:
@felipeelia This includes renaming the facet blocks to "Filter by". Should we be renaming the Facets feature also?

JakePT:
@oscarssanchez While working on these changes the Facet by Post Type block was introduced. To make that new block consistent with the changes made to the existing blocks, the following will need to be done:

- Rename the block to "Filter by Post Type"
- Add a description "Let visitors filter your content by post type."
- Use the elasticpress block category.
- Add keywords and textdomain to block.json.
- Add support for the same features, namely color, position, spacing and typography.
- Use the same filter icon.
- Use the shared components for the order controls and display count control.
- ~Display a placeholder when first inserting the block.~
- ~Create a post type control for selecting post type, based on the FacetMetaControl and FacetTaxonomyControl introduced in this PR.~
- ~To support the post type control you'll need to add actions to the store to retrieve post types.~

oscarssanchez:
@JakePT I've applied the changes here https://github.com/10up/ElasticPress/pull/3553 can you review please?

assets/js/blocks/facets/common/components/empty-response-placeholder.js
```diff
@@ -0,0 +1,31 @@
...
+		<Warning>
+			{type === 'taxonomy'
+				? __(
+						'Preview unavailable. The selected taxonomy has no terms assigned to searchable content.',
+						'elasticpress',
+				  )
+				: __(
+						'Preview unavailable. There is no searchable content available with the selected metadata.',
+						'elasticpress',
+				  )}
```
felipeelia, line 28: Shouldn't this be coming from the facet/filter? It seems weird to have a switch here depending on the meta (if we have a 4th one coming in the future)

assets/js/blocks/facets/common/components/facet-order-control.js
```diff
@@ -0,0 +1,51 @@
...
+			label: __('Highest to lowest count'),
+			value: 'count/desc',
+		},
+		{
+			label: __('Lowest to highest count'),
+			value: 'count/asc',
+		},
+		{
+			/* translators: label for ordering posts by title in ascending order */
+			label: __('A → Z', 'elasticpress'),
```
felipeelia: This should be a _x() with the comment as the context, no?

includes/classes/Feature/Facets/Types/Meta/Block.php
felipeelia, line 1: We can't remove methods in a minor version, so instead of removing them completely, let's keep them as a wrapper for a _doing_it_wrong() or _deprecated_function() call (4.7.0 has a bunch of these that can be used as an example)

includes/classes/Feature/Facets/Types/MetaRange/Block.php
```diff
@@ -114,15 +128,15 @@ public function check_facets_meta_rest_permission() {
 	 * @return string
 	 */
 	public function render_block( $attributes ) {
-		$attributes = $this->parse_attributes( $attributes );
-
```
felipeelia, line 117: $attributes is actually passed to the filter below, so can't be removed.

felipeelia (changes requested):
Overall it looks good @JakePT but we'll need to address a few things before merging it (especially the methods removal instead of deprecation)

assets/js/blocks/facets/common/components/empty-response-placeholder.js
JakePT: @felipeelia So in this case we’re sharing a lot of code between several facet blocks, so there isn’t anywhere to pass it down from. See: [https://github.com/10up/ElasticPress/pull/3403/files/db14e679a9d3718013a9bae9405f60cfdc279fde#diff-e1ffb7036f5d0252e614[…]85cde47d00d7fb1e7df417dR19](https://github.com/10up/ElasticPress/pull/3403/files/db14e679a9d3718013a9bae9405f60cfdc279fde#diff-e1ffb7036f5d0252e6140926216e93a9bf346215a85cde47d00d7fb1e7df417dR19) We share the edit component for the checkbox filters, so we would need to create a couple more files with a bunch of boilerplate just to pass down a different string.

includes/classes/Feature/Facets/Types/MetaRange/Block.php
JakePT: @felipeelia `$attributes` is passed as the first parameter of the `render_block` function, so it will be passed to the filter. We just don't pass it through `parse_attributes` anymore,

includes/classes/Feature/Facets/Types/Meta/Block.php
JakePT: @felipeelia I've restored these methods and deprecated them.

assets/js/blocks/facets/common/components/facet-order-control.js
JakePT: @felipeelia I think I just copied this from core, since I don't usually leave a `translators` unless there's a placeholder, but I've updated it to use `_x` instead.

includes/classes/Feature/Facets/Facets.php
```diff
@@ -188,25 +187,6 @@ public function set_agg_filters( $args, $query_args, $query ) {
 		return $args;
 	}
 
-	/**
-	 * Output scripts for widget admin
-	 *
-	 * @param  string $hook WP hook
-	 * @since  2.5
-	 */
-	public function admin_scripts( $hook ) {
```
felipeelia, line 197: @JakePT this should not be removed but wrap a _doing_it_wrong() call.

includes/classes/Feature/Facets/Types/Meta/Block.php
```diff
@@ -224,23 +179,12 @@ function ( $meta_fields ) use ( $attributes ) {
...
-				'searchPlaceholder' => esc_html_x( 'Search', 'Facet by meta search placeholder', 'elasticpress' ),
-				'facet'             => '',
-				'displayCount'      => false,
-				'orderby'           => 'count',
-				'order'             => 'desc',
-
-			]
+		_doing_it_wrong(
+			__METHOD__,
+			esc_html__( 'Since the Meta facets is only available as a block, attribute parsing is now left to block.json.', 'elasticpress' ),
```
felipeelia: I'm assuming this mechanic is compatible with WordPress 5.6, right?

includes/classes/Feature/Facets/Types/MetaRange/Block.php
felipeelia: my bad, thanks!

includes/classes/Feature/Facets/Types/MetaRange/Block.php
```diff
@@ -195,27 +209,4 @@ public function get_rest_registered_metakeys() {
...
 		return $meta_keys;
 	}
-
-	/**
-	 * Utilitary method to set default attributes.
-	 *
-	 * @param array $attributes Attributes passed
-	 * @return array
-	 */
-	protected function parse_attributes( $attributes ) {
```
felipeelia, line 204: this should be a wrapper for a _doing_it_wrong() call

includes/classes/Feature/Facets/Types/Taxonomy/Block.php
```diff
@@ -163,89 +160,21 @@ public function render_block( $attributes ) {
...
-		return ob_get_clean();
-	}
-
-	/**
-	 * Outputs the block preview
-	 *
-	 * @param \WP_REST_Request $request REST request
-	 * @return string
-	 */
-	public function render_block_preview( $request ) {
```
felipeelia, line 192: this one too

includes/classes/Feature/Facets/Types/Taxonomy/Block.php
```diff
@@ -163,89 +160,21 @@ public function render_block( $attributes ) {
...
-	}
+		$wrapper_attributes = get_block_wrapper_attributes( [ 'class' => 'wp-block-elasticpress-facet' ] );
 
-	/**
-	 * Utilitary method to set default attributes.
-	 *
-	 * @param array $attributes Attributes passed
-	 * @return array
-	 */
-	protected function parse_attributes( $attributes ) {
```
felipeelia, line 244: same here

felipeelia (changes requested):
a few more things @JakePT (i could be missing a few things, sorry in advance)

PR #3407, opened 2023-03-27
Require the utils file in the uninstaller

felipeelia:
### Description of the Change

Fixes an error while uninstalling the plugin:

Fatal error: Uncaught Error: Call to undefined function ElasticPress\Utils\get_capability() in /var/www/html/wp-content/plugins/elasticpress/uninstall.php:225

Closes #

### How to test the Change

Install and activate the plugin, then run `wp plugin uninstall elasticpress`

### Changelog Entry

> Fixed - Fatal error during plugin uninstall

### Credits

Props @felipeelia 

### Checklist:

### DIFF
--- a/.github/workflows/cypress-tests.yml
+++ b/.github/workflows/cypress-tests.yml
@@ -76,7 +76,8 @@ jobs:
       if: always()
       run: |
         PLUGIN_NAME=$(basename "$PWD")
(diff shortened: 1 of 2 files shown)


burhandodhy (approved):
LGTM

PR #3425, opened 2023-04-03
Feature: Did you mean

### LINKED ISSUE #2089: Feature: Did you mean....?
Everyone knows the "did you mean ...?" functionality big search engines offer. 

Adding such a feature into ElasticPress that would index successful search queries and the amount of results they return (in a safe and performant way) and compare current search queries against this list of known successful queries and suggest alternative searches would be a great feature that can help visitors find relevant content.

burhandodhy:
### Description of the Change

This PR adds the 'Did you mean' functionality just like other search engines have. The algorithm on ES will detect any typos in the search term and based on the result, the most preferred term will be disabled. To add the functionality to the search page, the user has to call the function `\ElasticPress\Features::factory()->get_registered_feature( 'did-you-mean' )->get_suggestion();`

Closes #2089

### How to test the Change

- Enable the "Did You Mean" Feature
- Add the method `\ElasticPress\Features::factory()->get_registered_feature( 'did-you-mean' )->get_suggestion();` on theme search template.
- Search a term with typo

This feature includes a setting that applies when the search results yield no matches.

1) "Display all the suggestions": When this option is enabled, ElasticPress will display all suggested terms.
2) "Automatically redirects user to top suggestion": Enabling this option will automatically redirect the user to the top suggested term when there are no search results.

### Changelog Entry

> Added - 'Did you mean' functionality when there is typo in search term

### Credits

Props @burhandodhy 

### Checklist:

### DIFF
--- a/elasticpress.php
+++ b/elasticpress.php
@@ -117,6 +117,10 @@ function register_indexable_posts() {
 	);
 
 	Features::factory()->register_feature(
+		new Feature\DidYouMean\DidYouMean()
+	);
+
+	Features::factory()->register_feature(
 		new Feature\WooCommerce\WooCommerce()
 	);
 
@@ -164,10 +168,6 @@ function register_indexable_posts() {
 
 	Features::factory()->register_feature(
 		new Feature\Terms\Terms()
-	);
-
-	Features::factory()->register_feature(
-		new Feature\DidYouMean\DidYouMean()
 	);
 
 	/**
--- a/elasticpress.php
+++ b/elasticpress.php
@@ -166,6 +166,10 @@ function register_indexable_posts() {
 		new Feature\Terms\Terms()
 	);
 
+	Features::factory()->register_feature(
+		new Feature\DidYouMean\DidYouMean()
+	);
+
 	/**
 	 * Register search algorithms
 	 */
--- a/includes/classes/Feature/DidYouMean/DidYouMean.php
+++ b/includes/classes/Feature/DidYouMean/DidYouMean.php
@@ -170,6 +170,7 @@ class DidYouMean extends Feature {
 		 * Filter the search analyzer use for the did you mean feature.
 		 *
 		 * @hook ep_search_suggestion_analyzer
+		 * @since 4.6.0
 		 * @param {array} $search_analyzer Search analyzer
 		 * @param {array} $formatted_args Formatted Elasticsearch query
 		 * @param {array} $args WP_Query arguments
--- a/includes/classes/Feature/DidYouMean/DidYouMean.php
+++ b/includes/classes/Feature/DidYouMean/DidYouMean.php
@@ -334,12 +334,23 @@ class DidYouMean extends Feature {
 	/**
 	 * Return a message to the user when the original search term has no results and the user is redirected to the suggested term.
 	 *
+	 * @param WP_Query $query WP_Query object
+	 *
 	 * @return string|void
 	 */
-	public function get_original_search_term() {
+	public function get_original_search_term( $query = null ) {
 		global $wp_query;
 
-		if ( ! $wp_query->is_main_query() || ! $wp_query->is_search() ) {
+		$settings = $this->get_settings();
+		if ( empty( $settings['active'] ) ) {
+			return false;
+		}
+
+		if ( ! $query && $wp_query->is_main_query() && $wp_query->is_search() ) {
+			$query = $wp_query;
+		}
+
+		if ( ! is_a( $query, '\WP_Query' ) ) {
 			return;
 		}
 
@@ -361,7 +372,7 @@ class DidYouMean extends Feature {
 			<span class="no-result">%s</span><strong>%s</strong>
 			</div>',
 			esc_html__( 'Showing results for: ', 'elasticpress' ),
-			esc_html( $wp_query->query_vars['s'] ),
+			esc_html( $query->query_vars['s'] ),
 			esc_html__( 'No results for: ', 'elasticpress' ),
 			esc_html( $original_term )
 		);
@@ -374,16 +385,16 @@ class DidYouMean extends Feature {
 		 * @param {string} $html HTML output
 		 * @param {string} $search_term Suggested search term
 		 * @param {string} $original_term Original search term
-		 * @param {WP_Query} $wp_query WP_Query object
+		 * @param {WP_Query} $query WP_Query object
 		 * @return {string} New HTML output
 		 */
-		return apply_filters( 'ep_suggestion_original_search_term_html', $html, $wp_query->query_vars['s'], $original_term, $wp_query );
+		return apply_filters( 'ep_suggestion_original_search_term_html', $html, $query->query_vars['s'], $original_term, $query );
 	}
 
 	/**
 	 * Returns the suggestion
 	 *
-	 * @param WP_Query $query WP_Query object. Default set to false.
+	 * @param WP_Query $query WP_Query object
 	 * @return void
 	 */
 	public function the_output( $query = null ) {
--- a/includes/classes/Feature/DidYouMean/DidYouMean.php
+++ b/includes/classes/Feature/DidYouMean/DidYouMean.php
@@ -46,6 +46,7 @@ class DidYouMean extends Feature {
 		add_filter( 'ep_post_formatted_args', [ $this, 'add_query_args' ], 10, 3 );
 		add_filter( 'ep_integrate_search_queries', [ $this, 'set_ep_suggestion' ], 10, 2 );
 		add_action( 'template_redirect', [ $this, 'automatically_redirect_user' ] );
+		add_action( 'ep_suggestions', [ $this, 'the_output' ] );
 	}
 
 	/**
@@ -128,22 +129,22 @@ class DidYouMean extends Feature {
 			return false;
 		}
 
-		$html = sprintf( '<span class="ep-suggested-spell-term">%s: <a href="%s">%s</a>?</span>', esc_html__( 'Did you mean', 'elasticpress' ), get_search_link( $term ), $term );
+		$html = sprintf( '<span class="ep-spell-suggestion">%s: <a href="%s">%s</a>?</span>', esc_html__( 'Did you mean', 'elasticpress' ), get_search_link( $term ), $term );
 
 		$html .= $this->get_alternatives_terms( $query );
 		$terms = $query->suggested_terms['options'] ?? [];
 
 		/**
-		 * Filter the did you mean suggested term HTML.
+		 * Filter the did you mean suggested HTML.
 		 *
 		 * @since 4.6.0
-		 * @hook ep_did_you_mean_suggested_term_html
+		 * @hook ep_suggestion_html
 		 * @param {string}   $html The HTML output.
 		 * @param {array}    $terms All suggested terms.
 		 * @param {WP_Query} $query The WP_Query object.
 		 * @return {string}  New $html value
 		 */
-		return apply_filters( 'ep_did_you_mean_suggestion_html', $html, $terms, $query );
+		return apply_filters( 'ep_suggestion_html', $html, $terms, $query );
 	}
 
 	/**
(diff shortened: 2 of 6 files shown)


elasticpress.php
```diff
@@ -166,6 +166,10 @@ function register_indexable_posts() {
 		new Feature\Terms\Terms()
 	);
 
+	Features::factory()->register_feature(
+		new Feature\DidYouMean\DidYouMean()
+	);
```
felipeelia: Let's ask @brandwaffle where in the list he wants to add this new feature. It could be one of the most visible ones instead of the last in the list.

includes/classes/Feature/DidYouMean/DidYouMean.php
```diff
@@ -0,0 +1,308 @@
+<?php
+/**
+ * Did You Mean feature.
+ *
+ * @since   4.6
```
felipeelia: Let's make it 4.6.0 (with the final .0) everywhere, please :)

includes/classes/Feature/DidYouMean/DidYouMean.php
```diff
@@ -0,0 +1,308 @@
...
+		$this->slug = 'did-you-mean';
+
+		$this->title = esc_html__( 'Did You Mean', 'elasticpress' );
+
+		$this->summary = __( '"Did You Mean" search feature provides alternative suggestions for misspelled or ambiguous search queries, enhancing search accuracy and user experience.', 'elasticpress' );
+
+		$this->requires_install_reindex = true;
+
+		$this->default_settings = [
+			'force_inactive'  => true,
```
felipeelia: Do we need this to be true?

includes/classes/Feature/DidYouMean/DidYouMean.php
```diff
@@ -0,0 +1,308 @@
...
+		parent::__construct();
+	}
+
+	/**
+	 * Setup search functionality.
+	 *
+	 * @return void
+	 */
+	public function setup() {
+		add_filter( 'ep_post_mapping', [ $this, 'add_mapping' ], 10 );
```
felipeelia:
```suggestion
		add_filter( 'ep_post_mapping', [ $this, 'add_mapping' ] );
```

includes/classes/Feature/DidYouMean/DidYouMean.php
```diff
@@ -0,0 +1,308 @@
...
+			'type'      => 'custom',
+			'tokenizer' => 'standard',
+			'filter'    => [
+				'lowercase',
+				'shingle_filter',
+			],
+		];
+
+		// Mapping.
+		$mapping['mappings']['properties']['post_content']['fields'] = [
```
felipeelia: Depending on the ES version this might need to be different (with a `'post'` between properties and post_content), no?

includes/classes/Feature/DidYouMean/DidYouMean.php
```diff
@@ -0,0 +1,308 @@
...
+		/**
+		 * Filter the Did You Mean HTML output.
+		 *
+		 * @since 4.6.0
+		 * @hook ep_did_you_mean_suggestion_html
+		 * @param {string}   $html The HTML output.
+		 * @param {WP_Query} $query The WP_Query object.
+		 * @return {string}  New $html value
+		 */
+		return apply_filters( 'ep_did_you_mean_suggestion_html', $html, $query );
```
felipeelia: Let's pass the term(s) here as well.

includes/classes/Feature/DidYouMean/DidYouMean.php
```diff
@@ -0,0 +1,308 @@
...
+		];
+
+		/**
+		 * Filter the search analyzer use for the did you mean feature.
+		 *
+		 * @hook ep_search_suggestion_analyzer
+		 * @param {array} $search_analyzer Search analyzer
+		 * @param {array} $formatted_args Formatted Elasticsearch query
+		 * @param {array} $args WP_Query arguments
+		 * @param {array} $wp_query WP_Query object
```
felipeelia: If $wp_query is a WP_Query object, the type of this `@param` needs to reflect that.

includes/classes/Feature/DidYouMean/DidYouMean.php
```diff
@@ -0,0 +1,308 @@
...
+				<label><input name="settings[search_behavior]" type="radio" <?php checked( $settings['search_behavior'], 'redirect' ); ?> value="redirect"><?php esc_html_e( 'Automatically redirects user to top suggestion', 'elasticpress' ); ?></label><br>
+			</div>
+		</div>
+		<?php
+	}
+
+	/**
+	 * Returns the list of other suggestions
+	 *
+	 * @param WP_Query $query WP_Query object
```
felipeelia, line 245: It needs to document the return as well (and as it can return `false` it can be signed as `: string`

includes/classes/Feature/DidYouMean/DidYouMean.php
```diff
@@ -0,0 +1,308 @@
...
+
+	/**
+	 * Set the ep_suggestion flag to true if the query is a search query.
+	 *
+	 * @param bool     $enabled Whether to enable the search queries integration.
+	 * @param WP_Query $query   The WP_Query object.
+	 */
+	public function set_ep_suggestion( $enabled, $query ) : bool {
+		if ( $query->is_search() && ! empty( $query->query_vars['s'] ) ) {
+			$query->set( 'ep_suggestion', true );
```
felipeelia, line 201: I think we will need different names here, no? One flag to say the query should search for suggestions and a different one to actually store suggestions?

includes/classes/Indexable/Post/QueryIntegration.php
```diff
@@ -359,6 +359,7 @@ public function get_es_posts( $posts, $query ) {
 			$query->found_posts           = $found_documents;
 			$query->num_posts             = $query->found_posts;
 			$query->max_num_pages         = ceil( $found_documents / $query->get( 'posts_per_page' ) );
+			$query->ep_suggestion         = isset( $ep_query['suggest']['ep_suggestion'] ) && isset( $ep_query['suggest']['ep_suggestion'][0] ) ? $ep_query['suggest']['ep_suggestion'][0] : [];
```
felipeelia: What about multiple suggestions?

felipeelia (commented):
@burhandodhy identified some changes we'll need to do but happy to talk about it next week.

includes/classes/Feature/DidYouMean/DidYouMean.php
burhandodhy: It depends; if we don't set it to true, the feature will be automatically enabled upon enabling the search feature.

burhandodhy: Found the root cause of the problem, and I removed my changes.

includes/classes/Feature/DidYouMean/DidYouMean.php
burhandodhy: Done.

includes/classes/Indexable/Post/QueryIntegration.php
burhandodhy: The 0 index does have all the suggestions.

includes/classes/Feature/DidYouMean/DidYouMean.php
```diff
@@ -0,0 +1,325 @@
...
+						'field' => 'post_content.shingle',
+					],
+				],
+			],
+		];
+
+		/**
+		 * Filter the search analyzer use for the did you mean feature.
+		 *
+		 * @hook ep_search_suggestion_analyzer
```
felipeelia, line 174: Can we add a `@since` here @burhandodhy ?

includes/classes/Feature/DidYouMean/DidYouMean.php
```diff
@@ -0,0 +1,325 @@
...
+		 * @param {array} $search_analyzer Search analyzer
+		 * @param {array} $formatted_args Formatted Elasticsearch query
+		 * @param {array} $args WP_Query arguments
+		 * @param {WP_Query} $wp_query WP_Query object
+		 *
+		 * @return  {array} New search analyzer
+		 */
+		$search_analyzer = apply_filters( 'ep_search_suggestion_analyzer', $search_analyzer, $formatted_args, $args, $wp_query );
```
felipeelia, line 181: Can we add a simple unit test for this filter?

felipeelia (changes requested):
@burhandodhy Can we also please add some documentation about how people can use this feature?

includes/classes/Feature/DidYouMean/DidYouMean.php
burhandodhy: Done. From now `suggested_terms` would have all the suggestions.

includes/classes/Feature/DidYouMean/DidYouMean.php
```diff
@@ -0,0 +1,395 @@
...
+		);
+
+		wp_safe_redirect( $url );
+		exit;
+	}
+
+	/**
+	 * Return a message to the user when the original search term has no results and the user is redirected to the suggested term.
+	 *
+	 * @return string
```
felipeelia: It can also return void @burhandodhy

includes/classes/Feature/DidYouMean/DidYouMean.php
```diff
@@ -0,0 +1,395 @@
...
+	 * Returns the suggestion
+	 *
+	 * @param WP_Query $query WP_Query object. Default set to false.
+	 * @return void
+	 */
+	public function the_output( $query = false ) {
+		$html  = $this->get_original_search_term();
+		$html .= $this->get_suggestion( $query );
+
+		echo $html; // phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped
```
felipeelia: Why can't we sanitize it with wp_kses_post or similar?

includes/classes/Feature/DidYouMean/DidYouMean.php
```diff
@@ -0,0 +1,395 @@
...
+	 *
+	 * @return void
+	 */
+	public function output_feature_box_settings() {
+		$settings = $this->get_settings();
+		?>
+		<div class="field">
+			<div class="field-name status"><?php esc_html_e( 'Search behavior when no result is found', 'elasticpress' ); ?></div>
+			<div class="input-wrap">
+				<label><input name="settings[search_behavior]" type="radio" <?php checked( ! (bool) $settings['search_behavior'] ); ?> value="0"><?php esc_html_e( 'Display a top suggestion', 'elasticpress' ); ?></label><br>
```
felipeelia:
```suggestion
				<label><input name="settings[search_behavior]" type="radio" <?php checked( ! (bool) $settings['search_behavior'] ); ?> value="0"><?php esc_html_e( 'Display the top suggestion', 'elasticpress' ); ?></label><br>
```

includes/classes/Feature/DidYouMean/DidYouMean.php
```diff
@@ -0,0 +1,395 @@
...
+	}
+
+	/**
+	 * Returns the suggestion
+	 *
+	 * @param WP_Query $query WP_Query object. Default set to false.
+	 * @return void
+	 */
+	public function the_output( $query = false ) {
+		$html  = $this->get_original_search_term();
```
felipeelia: Any reason why sometimes we consider it possible to work with a custom WP_Query and on this function it only uses the main_query?

includes/classes/Feature/DidYouMean/DidYouMean.php
```diff
@@ -0,0 +1,395 @@
...
+
+	/**
+	 * Initialize feature, setting it's config.
+	 */
+	public function __construct() {
+		$this->slug = 'did-you-mean';
+
+		$this->title = esc_html__( 'Did You Mean', 'elasticpress' );
+
+		$this->summary = __( '"Did You Mean" search feature provides alternative suggestions for misspelled or ambiguous search queries, enhancing search accuracy and user experience.', 'elasticpress' );
```
felipeelia:
```suggestion
		$this->summary = __( 'Recommend alternative search terms for misspelled queries or terms with no results', 'elasticpress' );
```

per https://github.com/10up/ElasticPress/issues/2089#issuecomment-1579014160

felipeelia (changes requested):
@burhandodhy some final changes we'll need to make before this can be merged. We also need to add somewhere that code is required to display the suggestions. Thoughts?

Also, I just noticed that the feature is auto-activating on existing installations and that is not something we can do on a minor release. Steps to reproduce:
- 10updocker wp option patch delete ep_feature_settings did-you-mean
- git checkout develop
- git checkout feature/2089

includes/classes/Feature/DidYouMean/DidYouMean.php
```diff
@@ -0,0 +1,395 @@
...
+	 */
+	public function requirements_status() : FeatureRequirementsStatus {
+		$features = Features::factory();
+		$search   = $features->get_registered_feature( 'search' );
+
+		if ( ! $search->is_active() ) {
+			return new FeatureRequirementsStatus( 2, esc_html__( 'This feature requires the "Post Search" feature to be enabled', 'elasticpress' ) );
+		}
+
+		return parent::requirements_status();
```
felipeelia: I thought we've talked about it before but I might be wrong. This feature should NOT return a FeatureRequirementsStatus->code 0, it should behave exactly like Instant Results, with a code 1.

includes/classes/Feature/DidYouMean/DidYouMean.php
burhandodhy: Yes, the logic behind this is to display the original search term only for the main search query.

includes/classes/Feature/DidYouMean/DidYouMean.php
burhandodhy: Done

includes/classes/Feature/DidYouMean/DidYouMean.php
burhandodhy: I missed that comment. updated the summary.

includes/classes/Feature/DidYouMean/DidYouMean.php
burhandodhy: Good catch. Done

includes/classes/Feature/DidYouMean/DidYouMean.php
burhandodhy: Done

PR #3432, opened 2023-04-05
PostSyncManager: Add missing kill_sync() to action_edited_term

rebeccahum:
### Description of the Change
We're missing `kill_sync()` in the `action_edited_term`, resulting in unexpected actions for it since it is not triggered (as we have a hook that calls it), like in the other ones e.g. `action_queue_meta_sync`

### Changelog Entry
Fixed action_edited_term to call kill_sync in SyncManager for post Indexable

### Credits
Props @rebeccahum 

### Checklist:

### DIFF
--- a/includes/classes/Indexable/Post/SyncManager.php
+++ b/includes/classes/Indexable/Post/SyncManager.php
@@ -554,6 +554,10 @@ class SyncManager extends SyncManagerAbstract {
 	 * @since  4.0.0
 	 */
 	public function action_edited_term( $term_id, $tt_id, $taxonomy ) {
+		if ( $this->kill_sync() ) {
+			return;
+		}
+
 		global $wpdb;
 
 		/**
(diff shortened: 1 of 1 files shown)


includes/classes/Indexable/Post/SyncManager.php
```diff
@@ -554,6 +554,10 @@ public function action_set_object_terms( $post_id, $terms, $tt_ids, $taxonomy, $
 	 * @since  4.0.0
 	 */
 	public function action_edited_term( $term_id, $tt_id, $taxonomy ) {
+		if ( $this->kill_sync() ) {
+			return;
+		}
+
```
felipeelia, line 562: @rebeccahum This should actually come after the global declaration. Do you mind making the change? Thanks

felipeelia (changes requested):

includes/classes/Indexable/Post/SyncManager.php
rebeccahum: @felipeelia I made the change. How come the global declaration affects it?

felipeelia: It does not affect it functionality-wise, but we always try to keep global declarations as the first line, just to make it easier to find/read.

PR #3462, opened 2023-05-16
WooCommerce weight by date

### LINKED ISSUE #3177: Weight results by date and WooCommerce Products

If "Weight results by date" is enabled, the search results for WooCommerce products are often not what is intended, as usually the publish date of a product should not interfere with its position in search.

### Designs

_No response_

We could disable that behavior in product-only WP Queries. To do that we could 
1. Create a new setting in the WooCommece feature
2. Change the existent setting in the Post Search feature to Enabled/Disabled/Disabled in product queries (to be displayed only when the WC feature is enabled)
3. Disable it via code and add a filter to change that (following the ["decisions, not options" philosophy](https://wordpress.org/about/philosophy/#decisions))

### Code of Conduct

MARQAS:
### Description of the Change
This PR will add two options to control the weighting in the search results for products. One will disable the weighting only for the product queries and the other will disable the weighting on any query that includes products.

Closes #3177 

### How to test the Change

- Enable the WooCommerce feature
- Notice that the Weight results by date settings will now have two additional options in it
- Select the third option (Disable for products) and search for a product on the products page(admin) and note that it will not have weighting. Normal search will have weighting even if it contains products in it. 
- Select the 4th option(Disable Weighting in any query that includes products) and search for a product and note that it will not have weighting. Any query that has a post-type product in it will not have the weighting args.
-  Selecting the 2nd option will disable it for any search you make.

### Changelog Entry

> Added - Two new options for Weight results by date to manage weighting for product post-types

### Credits

Props @MARQAS @felipeelia 

### Checklist:


includes/classes/Feature/Search/Search.php
```diff
@@ -9,6 +9,7 @@
 namespace ElasticPress\Feature\Search;
 
 use ElasticPress\Feature as Feature;
+use ElasticPress\Features as Features;
```
felipeelia: @MARQAS you do not need an alias of a class if you are using the original name. `use ElasticPress\Features;` should be enough (you can fix all the others as well.)

includes/classes/Feature/Search/Search.php
```diff
@@ -428,19 +429,24 @@ public function filter_query_post_type_for_search( $post_type, $query ) {
...
 			]
 		);
-
-		return (bool) $settings['decaying_enabled'];
+		$decaying = (bool) $settings['decaying_enabled'];
+		// Check decaying based on WooCommerce decay settings
+		if ( ( is_array( $post_type ) && in_array( 'product', $post_type, true ) && 'disabled_products_all' === $settings['decaying_enabled'] ) || ( 'product' === $post_type && in_array( $settings['decaying_enabled'], [ 'disabled_products', 'disabled_products_all' ], true ) ) ) {
+			$decaying = false;
+		}
+		return $decaying;
```
felipeelia: The Search class should not be aware of a "product" post type, that is something to be dealt with by the WooCommerce class.

includes/classes/Feature/Search/Search.php
```diff
@@ -629,7 +637,17 @@ public function output_feature_box_settings() {
 			<div class="field-name status"><?php esc_html_e( 'Weight results by date', 'elasticpress' ); ?></div>
 			<div class="input-wrap">
 				<label><input name="settings[decaying_enabled]" type="radio" <?php checked( (bool) $settings['decaying_enabled'] ); ?> value="1"><?php esc_html_e( 'Enabled', 'elasticpress' ); ?></label><br>
-				<label><input name="settings[decaying_enabled]" type="radio" <?php checked( ! (bool) $settings['decaying_enabled'] ); ?> value="0"><?php esc_html_e( 'Disabled', 'elasticpress' ); ?></label>
+				<label><input name="settings[decaying_enabled]" type="radio" <?php checked( ! (bool) $settings['decaying_enabled'] ); ?> value="0"><?php esc_html_e( 'Disabled', 'elasticpress' ); ?></label><br>
+				<?php
+				/**
+				 * Fires after the default Weight results by date settings
+				 *
+				 * @since  5.0.0
```
felipeelia: Let's please change this to `4.6.0`

includes/classes/Feature/Search/Weighting.php
```diff
@@ -8,6 +8,7 @@
 namespace ElasticPress\Feature\Search;
 
 use ElasticPress\Features;
+use ElasticPress\Feature as Feature;
```
felipeelia: Same as above

includes/classes/Feature/Search/Weighting.php
```diff
@@ -245,6 +246,9 @@ public function render_settings_page() {
 				$current_values = $this->get_weighting_configuration();
 
 				foreach ( $post_types as $post_type ) :
+					if ( 'product' === $post_type && ! $search->is_decaying_enabled( $post_type ) ) {
+						continue;
+					}
```
felipeelia: This class should not be dealing with specific post types as well.

includes/classes/Feature/Search/Weighting.php
```diff
@@ -568,6 +572,15 @@ public function post_type_has_fields( $post_type, $args = [] ) {
 	 * @return array Formatted ES args
 	 */
 	public function do_weighting( $formatted_args, $args ) {
+		$feature_settings = Utils\get_option( 'ep_feature_settings', [] );
```
felipeelia: Is this variable being used?

includes/classes/Feature/Search/Weighting.php
```diff
@@ -568,6 +572,15 @@ public function post_type_has_fields( $post_type, $args = [] ) {
...
+
+		/** Features Class @var Features $features */
+		$features = Features::factory();
+		/** Search Feature @var Feature\Search\Search $search */
+		$search = $features->get_registered_feature( 'search' );
```
felipeelia: Not sure why the docblock with the var type here but we generally go with `Features::factory()->get_registered_feature( 'search' );`

includes/classes/Feature/Search/Weighting.php
```diff
@@ -568,6 +572,15 @@ public function post_type_has_fields( $post_type, $args = [] ) {
...
+		if ( ! $search->is_decaying_enabled( $args['post_type'] ?? [] ) ) {
+			return $formatted_args;
+		}
```
felipeelia: Is there any reason why we would not apply date decaying AND regular weighting? I don't think one thing excludes the other.

includes/classes/Feature/WooCommerce/WooCommerce.php
```diff
@@ -886,6 +886,9 @@ public function setup() {
 		if ( $this->is_orders_autosuggest_enabled() ) {
 			$this->orders->setup();
 		}
+
+		// Add WooCommerce Settings for Weight results by date
+		add_action( 'ep_weight_settings_after_search', [ $this, 'add_weight_settings_search' ], 10, 1 );
```
felipeelia:
```suggestion
		add_action( 'ep_weight_settings_after_search', [ $this, 'add_weight_settings_search' ] );
```

includes/classes/Feature/WooCommerce/WooCommerce.php
```diff
@@ -1365,4 +1368,17 @@ public function is_orders_autosuggest_available() : bool {
 	public function is_orders_autosuggest_enabled() : bool {
 		return $this->is_orders_autosuggest_available() && '1' === $this->get_setting( 'orders' );
 	}
+
+	/**
+	 * Add weight results settings for WooCommerce settings
+	 *
+	 * @since 5.0.0
```
felipeelia: Let's change the `@since` tag here as well.

tests/php/features/TestSearch.php
```diff
@@ -227,6 +227,103 @@ public function testDecayingDisabled() {
 		);
 	}
 
+	/**
+	 * Test if decaying is disabled on products.
+	 *
+	 * @since 5.0
```
felipeelia: Same here

tests/php/features/TestSearch.php
```diff
@@ -227,6 +227,103 @@ public function testDecayingDisabled() {
...
+			! isset(
+				$this->fired_actions['ep_formatted_args']['query']['function_score'],
+				$this->fired_actions['ep_formatted_args']['query']['function_score']['functions'],
+				$this->fired_actions['ep_formatted_args']['query']['function_score']['functions'][0],
+				$this->fired_actions['ep_formatted_args']['query']['function_score']['functions'][0]['exp'],
+				$this->fired_actions['ep_formatted_args']['query']['function_score']['functions'][0]['exp']['post_date_gmt'],
+				$this->fired_actions['ep_formatted_args']['query']['function_score']['functions'][0]['exp']['post_date_gmt']['scale'],
+				$this->fired_actions['ep_formatted_args']['query']['function_score']['functions'][0]['exp']['post_date_gmt']['decay'],
+				$this->fired_actions['ep_formatted_args']['query']['function_score']['functions'][0]['exp']['post_date_gmt']['offset']
+			)
```
felipeelia: This is a very nice usage of isset()!

tests/php/features/TestSearch.php
```diff
@@ -227,6 +227,103 @@ public function testDecayingDisabled() {
...
+		ElasticPress\Elasticsearch::factory()->refresh_indices();
+
+		add_filter( 'ep_formatted_args', array( $this, 'catch_ep_formatted_args' ) );
+
+		$query = new \WP_Query(
+			[
+				's'         => 'test',
+				'post_type' => 'product',
+			]
+		);
```
felipeelia: Tests will need some improvement here @MARQAS. We need to be sure that `disabled_products` only interferes with queries that has only products, so we need to test it against queries that have products and something else. Same for disabled_products_all.

felipeelia (changes requested):
We'll need to separate concerns a bit better here before we can merge it @MARQAS. Perhaps a filter in Search that is used by WooCommerce is all you really need. I'm also not sure how this change interferes with regular weighting.

includes/classes/Feature/Search/Search.php
```diff
@@ -428,19 +429,32 @@ public function filter_query_post_type_for_search( $post_type, $query ) {
 	/**
 	 * Returns true/false if decaying is/isn't enabled
 	 *
+	 * @param string|array $post_type Post type
+	 *
 	 * @return bool
 	 */
-	public function is_decaying_enabled() {
+	public function is_decaying_enabled( $post_type = [] ) {
```
felipeelia: @MARQAS I think it makes more sense to pass all the args. People using the new filter could use more than the post type(s) to decide if the decay should or not be applied.

includes/classes/Feature/Search/Weighting.php
```diff
@@ -245,6 +246,9 @@ public function render_settings_page() {
 				$current_values = $this->get_weighting_configuration();
 
 				foreach ( $post_types as $post_type ) :
+					if ( ! $search->is_decaying_enabled( $post_type ) ) {
+						continue;
+					}
```
felipeelia: I guess I asked this before but what is the relationship between fields weighting and date decay @MARQAS ?

includes/classes/Feature/WooCommerce/WooCommerce.php
```diff
@@ -1365,4 +1370,33 @@ public function is_orders_autosuggest_available() : bool {
...
+	/**
+	 * Add weight results settings for WooCommerce settings
+	 *
+	 * @since 4.6.0
+	 * @param {array} $settings Current settings.
+	 */
+	public function add_weight_settings_search( $settings ) {
+		?>
+		<label><input name="settings[decaying_enabled]" type="radio" <?php checked( $settings['decaying_enabled'], 'disabled_products' ); ?> value="disabled_products"><?php esc_html_e( 'Disabled for products', 'elasticpress' ); ?></label><br>
+		<label><input name="settings[decaying_enabled]" type="radio" <?php checked( $settings['decaying_enabled'], 'disabled_products_all' ); ?> value="disabled_products_all"><?php esc_html_e( 'Disable Weighting in any query that includes products', 'elasticpress' ); ?></label>
```
felipeelia: Let's make these values a bit more intuitive. What do you think about `disabled_only_products` and `disabled_includes_products`? Also, let's run those options by @brandwaffle.

includes/classes/Feature/WooCommerce/WooCommerce.php
```diff
@@ -1365,4 +1370,33 @@ public function is_orders_autosuggest_available() : bool {
...
+	 * Enable/Disable Decaying based on WooCommerce Decay settings.
+	 *
+	 * @since 4.6.0
+	 * @param bool         $decaying Decaying
+	 * @param array        $settings  Settings
+	 * @param string|array $post_type Post type
+	 * @return bool
+	 */
+	public function ep_decaying_enabled_woocommerce( $decaying, $settings, $post_type ) {
+		if ( ( is_array( $post_type ) && in_array( 'product', $post_type, true ) && 'disabled_products_all' === $settings['decaying_enabled'] ) || ( 'product' === $post_type && in_array( $settings['decaying_enabled'], [ 'disabled_products', 'disabled_products_all' ], true ) ) ) {
```
felipeelia: Let's split this into more than one big statement. Also, remember to use early returns.

tests/php/features/TestSearch.php
```diff
@@ -227,6 +227,169 @@ public function testDecayingDisabled() {
...
+				$this->fired_actions['ep_formatted_args']['query']['function_score']['functions'][0]['exp']['post_date_gmt']['decay'],
+				$this->fired_actions['ep_formatted_args']['query']['function_score']['functions'][0]['exp']['post_date_gmt']['offset']
+			)
+		);
+		$this->assertTrue(
+			! isset(
+				$this->fired_actions['ep_formatted_args']['query']['bool'],
+				$this->fired_actions['ep_formatted_args']['query']['bool']['should']
+			)
+		);
```
felipeelia: Let's create some custom methods like `assertDecayEnabled` to abstract all this a bit. It is becoming hard to understand what exactly we want to make sure here.

felipeelia (changes requested):

includes/classes/Feature/Search/Weighting.php
felipeelia: This was not addressed but I made a new comment about it anyway.

burhandodhy (approved):
LGTM

Issue #3440, opened 2023-04-13
BUG: WooCommerce Custom Ordering + Name not working


lfinsaas:
### Describe the bug

WooCommerce Default Sort (Custom Ordering + Name) does not work. We expect to be able to sort the products by menu_order, but when ElasticPress is enabled, the shop does not respect the menu_order and sorts the products alphabetically.

When we disable ElasticPress, the Shop sorts by menu_order, as expected.

### Steps to Reproduce

We cannot let the shop remain purely alphabetical, nor can we disable ElasticPress. Thus, we're sorting the shop by Popularity until this is resolved (meaning we can't show steps to reproduce). Here is the site: https://jaxengrey.com/shop/ 

### WordPress and ElasticPress information

Wordpress: 6.2
ElasticPress: 4.5.1 
WooCommerce: 7.5.1

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

felipeelia:
Hi, @lfinsaas. Please install [Debug Bar ElasticPress](https://10up.github.io/ElasticPress/tutorial-debugging.html), visit the page where you want to sort products that way, and share with us the Requests' Info. Also, can you please share the Report Status content? Thanks.

lfinsaas:
Here are the outputs:

Here's the setting in Customizer:

<img width="815" alt="Screenshot 2023-04-17 at 12 07 42" src="https://user-images.githubusercontent.com/36778320/232454203-5834e619-62fe-4148-b326-c333f3095028.png">


REQUEST:
shop

QUERY STRING:
post_type=product

MATCHED REWRITE RULE:
shop/?$

MATCHED REWRITE QUERY:
post_type=product

<img width="1280" alt="Screenshot 2023-04-17 at 12 04 40" src="https://user-images.githubusercontent.com/36778320/232453415-a4a00cc3-77e1-4df9-86dd-7ef4d7f6b174.png">


## QUERIES INFO ##

### HTTPS://JAXENGREY.COM/SHOP (2023-04-17 04:55:20) ###
ES_REQ: POST ..../_SEARCH
REQUEST_ID: ....
QUERY_TIME: 285.449028015
WP_ARGS: ARRAY (
  'POST_TYPE' => 'PRODUCT',
  'ERROR' => '',
  'M' => '',
  'P' => 0,
  'POST_PARENT' => '',
  'SUBPOST' => '',
  'SUBPOST_ID' => '',
  'ATTACHMENT' => '',
[...]

PR #3468, opened 2023-05-22
Allow menu_order to be negative

felipeelia:
### Description of the Change
As pointed out in [this comment](https://github.com/10up/ElasticPress/issues/3440#issuecomment-1545446291), ElasticPress is not storing negative menu_order values but making them positive instead. This PR addresses this issue.

### How to test the Change
Create a product (or any post), give it a negative menu_order value, check Debug Bar EP and see the menu_order stored as a positive number

### Credits

Props @felipeelia and @navidabdi

### Checklist:


burhandodhy (approved):
LGTM

PR #3478, opened 2023-06-01
composer update + requiring composer installers

### LINKED ISSUE #3332: BUG: composer.json is missing in the zipped up release builds
### Describe the bug

If you include this plugin via composer it will download it as a generic library and is unaware it's a WP plugin.

The installed version no longer contains a `composer.json`

### Steps to Reproduce

Require `10up/elasticpress` with a stable released version

_No response_

### Environment information

_No response_

_No response_

### Code of Conduct

felipeelia:
### Description of the Change
Update composer packages and add installers

Closes #3332

### Changelog Entry

> Added - Installers added to composer.json, so `installer-paths` works without any additional requirement.

### Credits

Props @felipeelia @tomjn

### Checklist:

- [X] I agree to follow this project's [**Code of Conduct**](https://github.com/10up/.github/blob/trunk/CODE_OF_CONDUCT.md).

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -28,7 +28,8 @@
     }
   ],
   "require": {
-    "php": ">=7.0"
+    "php": ">=7.0",
+    "composer/installers": "^1.0 || ^2.0"
   },
   "require-dev": {
     "10up/phpcs-composer": "dev-master",
--- a/includes/classes/Command/Utility.php
+++ b/includes/classes/Command/Utility.php
@@ -158,21 +158,19 @@ class Utility {
 			*/
 			$should_interrupt_sync = wp_cache_get( $transient, 'transient', true );
 		} else {
-			$options = $wpdb->options;
-
-			// phpcs:disable
+			// phpcs:disable WordPress.DB.DirectDatabaseQuery
 			$should_interrupt_sync = $wpdb->get_var(
 				$wpdb->prepare(
 					"
 						SELECT option_value
-						FROM $options
+						FROM $wpdb->options
 						WHERE option_name = %s
 						LIMIT 1
 					",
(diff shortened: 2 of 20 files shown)


composer.json
```diff
@@ -28,7 +28,8 @@
     }
   ],
   "require": {
-    "php": ">=7.0"
+    "php": ">=7.0",
+    "composer/installers": "^1.0 || ^2.0"
```
tomjn, line 32: I literally just added this line to a different repo and noticed CI started failing because the allowed plugins wasn't added too, make sure you're doing this with an up to date Composer v2 as not all v2 releases require this but it's important when they do for certain things that it be accounted for

includes/classes/Command/Utility.php
```diff
@@ -160,8 +160,8 @@ public static function custom_get_transient( $pre_transient, $transient ) {
 		} else {
 			$options = $wpdb->options;
 
+			// phpcs:disable
```
burhandodhy: Can we use `// phpcs:disable WordPress.DB.DirectDatabaseQuery,` instead of ignoring everything for the block?

burhandodhy (dismissed):

composer.json
felipeelia: Thanks for jumping in @tomjn. In fact, we already have that code there.

includes/classes/Command/Utility.php
felipeelia: good catch @burhandodhy, thanks. I've addressed this in [bbc71be](https://github.com/10up/ElasticPress/pull/3478/commits/bbc71be4d98f0da8b79ec548d148969812593775).

PR #3495, opened 2023-06-12
Ensure get_elasticsearch_error_reason returns a string

felipeelia:
The title says it all. This is really just a safety measure for any unknown return format.


burhandodhy (approved):
LGTM

PR #3499, opened 2023-06-18
Implement abstract Block facet feature class

### LINKED ISSUE #3472: Enhancement: Default methods for Facet Block class

I've found similar blocks of code like the following in the Block classes from different facets:

	public function check_facets_meta_rest_permission() {
		if ( ! current_user_can( 'edit_theme_options' ) ) {
			return new \WP_Error( 'ep_rest_forbidden', esc_html__( 'Sorry, you cannot view this resource.', 'elasticpress' ), array( 'status' => 401 ) );
		}

		return true;
	}

In other facet types that seem to be repeating the same code. I think we can improve this by having an abstract Block class that can be extended. If the rest permissions need to change the method can be redeclared without a problem.

### Designs

_No response_

_No response_

### Code of Conduct

oscarssanchez:
### Description of the Change

This PR implements two abstract classes for the Facets feature. It tries to gather as many commonalities as possible from the facets so in the future, if we need to add subsequent Facet types, we can use the abstract classes as a guide to reduce developing times.

Closes https://github.com/10up/ElasticPress/issues/3472

### Changelog Entry

> Changed - Moved methods to abstract facet class

### Credits

Props @oscarssanchez, @felipeelia 

### Checklist:

### DIFF
--- a/includes/classes/Feature/Facets/Block.php
+++ b/includes/classes/Feature/Facets/Block.php
@@ -2,8 +2,8 @@
 /**
  * Abstract Facet Block class.
  *
- * @since  4.7
- * @package  elasticpress
+ * @since 4.7.0
+ * @package elasticpress
  */
 
 namespace ElasticPress\Feature\Facets;
--- a/includes/classes/Feature/Facets/Block.php
+++ b/includes/classes/Feature/Facets/Block.php
@@ -1,0 +1,60 @@
+<?php
+/**
+ * Abstract Facet Block class.
+ *
+ * @since  4.7
+ * @package  elasticpress
+ */
+
+namespace ElasticPress\Feature\Facets;
+
+/**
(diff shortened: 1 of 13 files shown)


oscarssanchez:
As a next step, we could think about `get_facet_item_value_html()` and make a decision if it makes sense to take out the "is_selected" values and merge them into one array, or if we just want to handle them separately. 

Both have implications specially in the original Facet feature block. We'll need to deprecate those parameters accordingly.

felipeelia (changes requested):
@oscarssanchez I've pushed a commit changing a few things but it seems we still need to change the `check_facets_rest_permission` call in other Block.php files. Also, do you mind calling me on slack to explain more about that $is_selected parameter? I remember we talked about it but now that I looked into it, it seems unnecessary.

PR #3500, opened 2023-06-19
Sync command stop-on-error flag

### LINKED ISSUE #2000: Create a new --stop-on-errors parameter for the WP-CLI index command

Currently, WP-CLI index command will only stop if the response to the request is an error. Otherwise, it will record all errors (like "too many fields" errors) and present them at the end of the whole process.

It would be nice to have a way to stop everything on the first sign of a problem. 

**Additional context**
Related lines of code:
Line where the command will stop if the response is an error: https://github.com/10up/ElasticPress/blob/develop/includes/classes/Command.php#L889
Line where the command will add the "soft errors" to an array:
https://github.com/10up/ElasticPress/blob/develop/includes/classes/Command.php#L917
Line where all errors are finally displayed:
https://github.com/10up/ElasticPress/blob/develop/includes/classes/Command.php#L946

oscarssanchez:
### Description of the Change

This PR introduces a new flag `stop-on-error` to stop sync the moment we find an error. We accomplish this by returning an error logged by `output()` that later kills the sync.

Closes #2000

### Changelog Entry

> Added - Flag --stop-on-error

### Credits

Props @oscarssanchez 

### Checklist:

### DIFF
--- a/includes/classes/Command.php
+++ b/includes/classes/Command.php
@@ -1311,7 +1311,7 @@ class Command extends WP_CLI_Command {
 		static $time_elapsed = 0, $counter = 0;
 
 		// Special treatment if stop-on-error flag is set.
-		if ( ! empty( $args['stop_on_error'] ) && in_array( $message['status'], [ 'error', 'warning' ] ) ) {
+		if ( ! empty( $args['stop_on_error'] ) && in_array( $message['status'], [ 'error', 'warning' ], true ) ) {
 			$this->clear_sync();
 			WP_CLI::error( $message['message'] );
 		}
--- a/includes/classes/Command.php
+++ b/includes/classes/Command.php
@@ -689,6 +689,9 @@ class Command extends WP_CLI_Command {
 	 * [--show-nobulk-errors]
 	 * : Display the error message returned from Elasticsearch when a post fails to index while not using the /_bulk endpoint
 	 *
+	 * [--stop-on-error]
+	 * : Stop indexing if an error is encountered and display the error.
+	 *
 	 * [--offset=<offset_number>]
 	 * : Skip the first n posts (don't forget to remove the `--setup` flag when resuming or the index will be emptied before starting again).
 	 *
@@ -800,8 +803,14 @@ class Command extends WP_CLI_Command {
 			'static_bulk'    => $static_bulk,
 		];
 
-		if ( isset( $assoc_args['show-errors'] ) || ( isset( $assoc_args['show-bulk-errors'] ) && ! $no_bulk ) || ( isset( $assoc_args['show-nobulk-errors'] ) && $no_bulk ) ) {
+		$show_errors = isset( $assoc_args['show-errors'] ) || ( isset( $assoc_args['show-bulk-errors'] ) && ! $no_bulk ) || ( isset( $assoc_args['show-nobulk-errors'] ) && $no_bulk );
+
+		if ( $show_errors ) {
 			$index_args['show_errors'] = true;
+		}
+
+		if ( isset( $assoc_args['stop-on-error'] ) && $show_errors ) {
+			$index_args['stop_on_error'] = true;
 		}
 
 		if ( ! empty( $assoc_args['post-ids'] ) ) {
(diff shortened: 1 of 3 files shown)


felipeelia:
@oscarssanchez if I add a snippet like this in a mu-plugin and run the sync, it will only stop if --show-errors is also present:

```php
add_filter(
	'ep_prepare_meta_data',
		}
		return $post_meta;
	},
	10,
	2
);
```

Is that intended?

includes/classes/Command.php
```diff
@@ -800,10 +803,16 @@ public function sync( $args, $assoc_args ) {
...
 		];
 
-		if ( isset( $assoc_args['show-errors'] ) || ( isset( $assoc_args['show-bulk-errors'] ) && ! $no_bulk ) || ( isset( $assoc_args['show-nobulk-errors'] ) && $no_bulk ) ) {
+		$show_errors = isset( $assoc_args['show-errors'] ) || ( isset( $assoc_args['show-bulk-errors'] ) && ! $no_bulk ) || ( isset( $assoc_args['show-nobulk-errors'] ) && $no_bulk );
+
+		if ( $show_errors ) {
 			$index_args['show_errors'] = true;
 		}
 
+		if ( isset( $assoc_args['stop-on-error'] ) && $show_errors ) {
```
felipeelia: @oscarssanchez isn't this one of those cases where it would be better to use `\WP_CLI\Utils\get_flag_value( $assoc_args, 'stop-on-error', false );` like you did a while ago? As we are changing $show_errors, probably a good place to change it too.

tests/php/TestCommands.php
```diff
@@ -929,6 +929,57 @@ public function testRequest() {
 		$this->assertStringContainsString( 'Response:', $output );
 	}
 
+	/**
+	 * Test sync command with stop-on-error flag.
+	 * Expect an error message that stops the sync instead of a warning.
+	 */
```
felipeelia: @oscarssanchez can we add a `@since` tag here please?

tests/php/TestCommands.php
```diff
@@ -929,6 +929,57 @@ public function testRequest() {
...
+	public function testSyncStopOnError() {
```
felipeelia: Not a blocker but for newer tests we are using snake-case for test names. Do you mind renaming it to `test_sync_stop_on_error()`? We'll have a follow up issue to rename all test methods.

tests/php/TestCommands.php
```diff
@@ -929,6 +929,57 @@ public function testRequest() {
...
+					$fake_request->items[0] = $fake_item;
+
+					$request['body'] = wp_json_encode( $fake_request );
+
+					return $request;
+				}
+
+				return $request;
+			}
+		);
```
felipeelia, line 966: Instead of faking the return via a filter, it would be faster to mock the remote_request response (similar to what I did here: https://github.com/10up/ElasticPress/blob/develop/tests/php/TestElasticsearch.php#L156) Do you mind making that change?

tests/php/TestCommands.php
```diff
@@ -929,6 +929,57 @@ public function testRequest() {
...
+
+		Indexables::factory()->get( 'post' )->delete_index();
+
+		$this->ep_factory->post->create_many( 10 );
```
felipeelia: With the above, I think this can go away. If mocking the method return is not doable, can we avoid deleting the index at least?

tests/php/TestCommands.php
```diff
@@ -929,6 +929,57 @@ public function testRequest() {
...
+
+		$this->expectExceptionMessage( '10 (Post): [my dummy error type] my dummy error reason' );
+
+		$this->command->sync(
+			[],
+			[
+				'setup'         => true,
+				'show-errors'   => true,
+				'stop-on-error' => true,
+				'yes'           => true,
```
felipeelia: Just to make sure it already works with the minimal case, can we remove the setup, show-errors, and the yes flags?

felipeelia (changes requested):

PR #3503, opened 2023-06-20
Multisite context switch support to SyncManager

### LINKED ISSUE #3471: BUG: Sync not working properly with multisite
### Describe the bug

The bug has been detected with multisite using Multilingualpress although it might occur in other multisite environment.

Multilingualpress provides a mechanism to replicate a post  from one site to other sites. When we are editing a post, we select one or more sites where the current post will be copied. When the replication process is launched, it changes the context to the target site and order the creation/update of the post. 

In this very moment, when the update operation is triggered, if the post is allowed to be indexed, ElasticPress stores the post ID in a queue in order to be processed. 

What happens? the queue does not store any data indicating which site each ID belongs to. This produces a queue with several post IDs belonging to different sites, but all of them will be processed against the same site. This way, Elastic will index records that should not be indexing.

### Steps to Reproduce

1. Go to the edit page of a post (e.g. Site 1 - ID 21259)
2. Using multilingualpress, select one site where the current post will be copied (e.g Site1 to Site3)
3. Update the post

The current post (21259) will be updated in site 1. This ID will be stored in a queue.
The target post (20782) will be created in the site 3 and this ID will be stored in a queue.

The queue will process the ID 21259 and 20782, both in site 1. If 20782 is an attachment, it will be indexed. But the post 20782 of site 3 will not be synced.

_No response_

Tested in:
ElasticPress 4.3.1 and 4.5.2
MultilingualPress 4.1.1

### Code of Conduct

sathyapulse:
This commit adds support for Multisite context switching to the SyncManager class. This allows it to index the content to the correct site when indexing in the shutdown hook.

The commit introduces two methods in the SyncManager class: `get_sync_queue` and `reset_sync_queue`. The `get_sync_queue` method returns the sync queue for the current site or the `$blog_id`. The `reset_sync_queue` method resets the sync queue for the current site or the `$blog_id`.

This commit also updates all the dependent tests to use the new methods.

Fixes #3471 issue.

### Description of the Change

Closes #3471

### Changelog Entry

> Added - New feature
> Removed - Feature
> Fixed - Bug fix
> Security - Vulnerability

### Credits

### Checklist:

### DIFF
--- a/includes/classes/Indexable/Post/SyncManager.php
+++ b/includes/classes/Indexable/Post/SyncManager.php
@@ -297,9 +297,7 @@ class SyncManager extends SyncManagerAbstract {
 		 * Make sure to remove this post from the sync queue in case an shutdown happens
 		 * before a redirect when a redirect has already been triggered.
 		 */
-		if ( isset( $this->sync_queue[ $post_id ] ) ) {
-			unset( $this->sync_queue[ $post_id ] );
-		}
+		$this->remove_from_queue( $post_id );
 	}
 
 	/**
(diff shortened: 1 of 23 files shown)


felipeelia:
Closing in favor of #3689. Thanks @sathyapulse!

PR #3506, opened 2023-06-21
Add ep_facet_special_slug_taxonomies filter

### LINKED ISSUE #3467: Enhancement: Move Woocommerce logic from facet by taxonomy to the Woocommerce feature

In facet by taxonomy, we have this code:
```

		// Account for taxonomies that should be woocommerce attributes, if WC is enabled
		$attribute_taxonomies = [];
		if ( function_exists( 'wc_attribute_taxonomy_name' ) ) {
			$all_attr_taxonomies = wc_get_attribute_taxonomies();

			foreach ( $all_attr_taxonomies as $attr_taxonomy ) {
				$attribute_taxonomies[ $attr_taxonomy->attribute_name ] = wc_attribute_taxonomy_name( $attr_taxonomy->attribute_name );
			}
		}

```

This checks if the function exists, but doesn't take into account if the Woocommerce feature is activated, for example. It also polludes the "abstraction" incorporating an edge case for another plugin. We are already doing this in the Woocommerce feature for example, it could help to move things there to keep them organized. 

I think we could solve both by adding a filter or an action, and we can move the code snippet to the Woocommerce feature.  It could also solve future implementation problems for similar features that want to achieve the same. 

### Designs

_No response_

_No response_

### Code of Conduct

oscarssanchez:
### Description of the Change
This PR introduces a new filter to account for the edge case from https://github.com/10up/ElasticPress/pull/3506/files#diff-64cabb18869a398887abf9cc35592a9636381971dedfc6e1ee1b3ada27188edfL175 and moves the code to the Woocommerce feature.

If we ever need to do something like this again, we can reuse this filter.

Closes #3467

### Changelog Entry

> Added - ep_facet_tax_special_slug_taxonomies filter

### Credits

Props @oscarssanchez 

### Checklist:

### DIFF
--- a/includes/classes/Feature/Facets/Types/Taxonomy/FacetType.php
+++ b/includes/classes/Feature/Facets/Types/Taxonomy/FacetType.php
@@ -176,11 +176,11 @@ class FacetType extends \ElasticPress\Feature\Facets\FacetType {
 		 * Filter for treatment special slugs in taxonomies. This is used in case you need to change the default taxonomy slug
 		 *
 		 * @since 4.7.0
-		 * @hook ep_facet_special_slug_taxonomies
+		 * @hook ep_facet_tax_special_slug_taxonomies
 		 * @param  {array} $special_taxonomies Taxonomies with special slugs
 		 * @return {array} $selected_filters Selected filters
 		 */
-		$special_taxonomies = apply_filters( 'ep_facet_special_slug_taxonomies', [], $selected_filters );
+		$special_taxonomies = apply_filters( 'ep_facet_tax_special_slug_taxonomies', [], $selected_filters );
 
 		foreach ( $selected_filters['taxonomies'] as $taxonomy => $filter ) {
 			$tax_query[] = [
@@ -222,11 +222,11 @@ class FacetType extends \ElasticPress\Feature\Facets\FacetType {
 		 * Filter for treatment special slugs in taxonomies. This is used in case you need to change the default taxonomy slug
 		 *
 		 * @since 4.7.0
-		 * @hook ep_facet_special_slug_taxonomies
+		 * @hook ep_facet_tax_special_slug_taxonomies
 		 * @param  {array} $special_taxonomies Taxonomies with special slugs
 		 * @return {array} $selected_filters Selected filters
 		 */
-		$special_taxonomies = apply_filters( 'ep_facet_special_slug_taxonomies', [], $selected_filters );
+		$special_taxonomies = apply_filters( 'ep_facet_tax_special_slug_taxonomies', [], $selected_filters );
 
 		$match_type = $feature->get_match_type();
 
--- a/includes/classes/Feature/Facets/Types/Taxonomy/FacetType.php
+++ b/includes/classes/Feature/Facets/Types/Taxonomy/FacetType.php
@@ -216,7 +216,7 @@ class FacetType extends \ElasticPress\Feature\Facets\FacetType {
 		 * Filter for treatment special slugs in taxonomies. This is used in case you need to change the default taxonomy slug.
 		 *
 		 * @since 4.7.0
-		 * @hook ep_facet_tax_special_slug_taxonomies.
+		 * @hook ep_facet_tax_special_slug_taxonomies
 		 * @param  {array} $special_taxonomies Taxonomies with special slugs.
 		 * @param  {array} $selected_filters Selected filters.
 		 * @return {array} New taxonomies with special slugs.
--- a/includes/classes/Feature/Facets/Types/Taxonomy/FacetType.php
+++ b/includes/classes/Feature/Facets/Types/Taxonomy/FacetType.php
@@ -207,23 +207,27 @@ class FacetType extends \ElasticPress\Feature\Facets\FacetType {
 		}
 
 		$selected_filters = $feature->get_selected();
+
+
 		if ( empty( $selected_filters ) || empty( $selected_filters[ $this->get_filter_type() ] ) ) {
 			return;
 		}
 
 		/**
-		 * Filter for treatment special slugs in taxonomies. This is used in case you need to change the default taxonomy slug
+		 * Filter for treatment special slugs in taxonomies. This is used in case you need to change the default taxonomy slug.
 		 *
 		 * @since 4.7.0
-		 * @hook ep_facet_tax_special_slug_taxonomies
-		 * @param  {array} $special_taxonomies Taxonomies with special slugs
-		 * @return {array} $selected_filters Selected filters
+		 * @hook ep_facet_tax_special_slug_taxonomies.
+		 * @param  {array} $special_taxonomies Taxonomies with special slugs.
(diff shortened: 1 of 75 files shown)


tests/php/features/TestFacetTypeTaxonomy.php
```diff
@@ -322,4 +322,32 @@ public function testAddQueryParams() {
 
 		$this->assertSame( $new_query_params, $expected_query_params );
 	}
+
+	/**
+	 * Test the ep_facet_tax_special_slug_taxonomies filter runs.
+	 *
+	 * @return void
+	 */
```
felipeelia: Can we add a `@since` tag here, please?

includes/classes/Feature/Facets/Types/Taxonomy/FacetType.php
```diff
@@ -172,19 +172,19 @@ public function facet_query( $query ) {
...
-			}
-		}
+		/**
+		 * Filter for treatment special slugs in taxonomies. This is used in case you need to change the default taxonomy slug
+		 *
+		 * @since 4.7.0
+		 * @hook ep_facet_tax_special_slug_taxonomies
+		 * @param  {array} $special_taxonomies Taxonomies with special slugs
+		 * @return {array} $selected_filters Selected filters
+		 */
```
felipeelia: We don't need to repeat the full docblock here, it is enough to change this to
```suggestion
		/** This filter is documented below */
```
like WordPress core does (in WP they mention the same file name, which I think is a redundancy.)

felipeelia (changes requested):

tests/php/features/TestFacetTypeTaxonomy.php
oscarssanchez: This is now added here: https://github.com/10up/ElasticPress/pull/3506/files#diff-d36fb3af9a833294f34cf95964cbdc054fb4d401805400564aa71bc8ce01db90R329

includes/classes/Feature/Facets/Types/Taxonomy/FacetType.php
oscarssanchez: I've removed the docblock @felipeelia

includes/classes/Feature/Facets/Types/Taxonomy/FacetType.php
```diff
@@ -218,20 +211,20 @@ public function add_query_filters( $filters ) {
...
-				$attribute_taxonomies[ $attr_taxonomy->attribute_name ] = wc_attribute_taxonomy_name( $attr_taxonomy->attribute_name );
-			}
-		}
+		/**
+		 * Filter for treatment special slugs in taxonomies. This is used in case you need to change the default taxonomy slug
+		 *
+		 * @since 4.7.0
+		 * @hook ep_facet_tax_special_slug_taxonomies
+		 * @param  {array} $special_taxonomies Taxonomies with special slugs
+		 * @return {array} $selected_filters Selected filters
```
felipeelia: @oscarssanchez in this docblock, $selected_filters is actually a parameter (then we need to add the return)

includes/classes/Feature/WooCommerce/WooCommerce.php
```diff
@@ -648,4 +649,19 @@ public function maybe_disable_decaying( $is_decaying_enabled, $settings, $args )
 		_deprecated_function( __METHOD__, '4.7.0', "\ElasticPress\Features::factory()->get_registered_feature( 'woocommerce' )->products->maybe_disable_decaying()" );
 		return $this->products->maybe_disable_decaying( $is_decaying_enabled, $settings, $args );
 	}
+
+	/**
+	 * Add taxonomies that should be woocommerce attributes.
+	 *
+	 * @param array $attribute_taxonomies  Attribute taxonomies.
+	 * @return array $attribute_taxonomies Attribute taxonomies.
+	 */
```
felipeelia: Can we add a `@since` here please, @oscarssanchez ?

includes/classes/Feature/WooCommerce/WooCommerce.php
```diff
@@ -648,4 +649,19 @@ public function maybe_disable_decaying( $is_decaying_enabled, $settings, $args )
...
+	public function add_taxonomy_attributes( $attribute_taxonomies ) {
```
felipeelia:
```suggestion
	public function add_taxonomy_attributes( array $attribute_taxonomies ) : array {
```
Especially for methods we have more control on, we are trying to specify types, @oscarssanchez :)

tests/php/features/TestFacetTypeTaxonomy.php
```diff
@@ -322,4 +322,33 @@ public function testAddQueryParams() {
...
+		);
+
+		$facet_feature = Features::factory()->get_registered_feature( 'facets' );
+		$facet_type    = $facet_feature->types['taxonomy'];
+
+		parse_str( 'ep_filter_taxonomy=dolor,amet', $_GET );
+
+		$facet_type->add_query_filters( [] );
+
+		$this->assertGreaterThanOrEqual( 1, did_filter( 'ep_facet_tax_special_slug_taxonomies' ) );
```
felipeelia, line 355: @oscarssanchez can you add an assertion to check if `testmyfilter` was really added to the final return?

felipeelia (changes requested):
@oscarssanchez do you mind putting together a small test covering the new add_taxonomy_attributes method, please?

oscarssanchez:
@felipeelia I've added the test here https://github.com/10up/ElasticPress/pull/3506/files#diff-a0c161d42d6543c88327f86bccbc9be735eacb277cf60c3a35848775b03c24a0R364

includes/classes/Feature/Facets/Types/Taxonomy/FacetType.php
oscarssanchez: This is now fixed here https://github.com/10up/ElasticPress/pull/3506/files @felipeelia

includes/classes/Feature/WooCommerce/WooCommerce.php
oscarssanchez: Added here https://github.com/10up/ElasticPress/pull/3506/files#diff-5d9d9db6657b73dc4a922c343a4c5a48dc38584aa8cae7123c90a415900e456aR656 @felipeelia

includes/classes/Feature/WooCommerce/WooCommerce.php
oscarssanchez: Changed here @felipeelia

tests/php/features/TestFacetTypeTaxonomy.php
oscarssanchez: Added here @felipeelia  https://github.com/10up/ElasticPress/pull/3506/files#diff-d36fb3af9a833294f34cf95964cbdc054fb4d401805400564aa71bc8ce01db90R354

PR #3521, opened 2023-06-30
Avoid empty meta fields

### LINKED ISSUE #3490: avoid empty meta fields that bloat post_meta table

I have a very large Wordpress site with millions of posts in it (you can imagine the need for ElasticPress!) and I noticed that I have a lot of posts that have the `ep_exclude_from_search` meta field saved in it with no value (representing that these posts were saved with the "Exclude from search results" box unchecked).  When you're trying to manage a post meta table in such a large site, little pieces of meta that are unnecessary add up and bloat that database.  Could we instead set up the plugin to only store the field if there's a value?

We could easily change the `save_exclude_from_search_meta()` function in [/includes/classes/Feature/Search/Search.php](https://github.com/10up/ElasticPress/blob/develop/includes/classes/Feature/Search/Search.php#L788) from what it currently is: 

```php
$exclude_from_search = isset( $_POST['ep_exclude_from_search'] ) ? true : false;
update_post_meta( $post_id, 'ep_exclude_from_search', $exclude_from_search );
```

to:

```php
if ( isset( $_POST['ep_exclude_from_search'] ) ) {
    update_post_meta( $post_id, 'ep_exclude_from_search', true );
} else {
    delete_post_meta( $post_id, 'ep_exclude_from_search' );
}
```

Maybe a small enhancement to some, but every little bit of efficiency makes the plugin a little bit better!

### Designs

_No response_

_No response_

### Code of Conduct

MARQAS:
### Description of the Change
This PR will delete the meta fields when it is not set instead of updating the meta with false or empty value

Closes #3490 

### How to test the Change
- Create a post without setting the Exclude from search option in the post
- Check the post meta and notice that it should be empty if it is not set

### Changelog Entry

> Changed - Update post metas only if they are set or have some value

### Credits

Props @MARQAS 

### Checklist:

### DIFF
--- a/includes/classes/Feature/SearchOrdering/SearchOrdering.php
+++ b/includes/classes/Feature/SearchOrdering/SearchOrdering.php
@@ -533,12 +533,8 @@ class SearchOrdering extends Feature {
 			}
 		}
 
+		update_post_meta( $post_id, 'pointers', $final_order_data );
 		update_post_meta( $post_id, 'search_term', $post->post_title );
-		if ( ! empty( $final_order_data ) ) {
-			update_post_meta( $post_id, 'pointers', $final_order_data );
-		} else {
-			delete_post_meta( $post_id, 'pointers' );
-		}
 	}
 
 	/**
--- a/includes/classes/Feature/SearchOrdering/SearchOrdering.php
+++ b/includes/classes/Feature/SearchOrdering/SearchOrdering.php
@@ -533,8 +533,12 @@ class SearchOrdering extends Feature {
 			}
 		}
 
-		update_post_meta( $post_id, 'pointers', $final_order_data );
 		update_post_meta( $post_id, 'search_term', $post->post_title );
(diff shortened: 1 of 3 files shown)


includes/classes/Feature/SearchOrdering/SearchOrdering.php
```diff
@@ -533,8 +533,12 @@ public function save_post( $post_id, $post ) {
 			}
 		}
 
-		update_post_meta( $post_id, 'pointers', $final_order_data );
 		update_post_meta( $post_id, 'search_term', $post->post_title );
+		if ( ! empty( $final_order_data ) ) {
+			update_post_meta( $post_id, 'pointers', $final_order_data );
+		} else {
+			delete_post_meta( $post_id, 'pointers' );
+		}
```
felipeelia: do we have tests covering this change @MARQAS? Also, you are sure this does not break the functionality, right?

PR #3568, opened 2023-07-29
Add isFirstSearch prop to Instant Results

### LINKED ISSUE #3508: BUG: Instant Results shows 0 results or previous results while loading
### Describe the bug

When performing a search with the instant results feature, if the request takes some time to be returned to the client there will be a screen that shows zero results or the previous results set if a search was performed. There is no indication to the user that results are in the process of updating. We should provide some sort of indicator to this effect.

### Steps to Reproduce

1.  input a search term and hit enter to open instant results modal
2.  if there is a lag, "0 results" will appear until the request completes

_No response_

### Environment information

_No response_

EP 4.6

### Code of Conduct

oscarssanchez:
### Description of the Change

This PR adds a new `isFirstSearch` to keep track of whether we are on the first time searching in instant results. If it is set to true instant results will show a "Loading results" text until we have set the results from Elastic.

Closes https://github.com/10up/ElasticPress/issues/3508

### How to test the Change

1.- With instant results activated, search for a term
2.- You should see the text "Loading results" instead of "0 results" for a brief moment

### Changelog Entry

> Fixed - Display "Loading results" instead of "0 results" on first search.

### Credits

Props @oscarssanchez, @JakePT, @felipeelia 

### Checklist:

### DIFF
--- a/assets/js/api-search/index.js
+++ b/assets/js/api-search/index.js
@@ -105,7 +105,6 @@ export const ApiSearchProvider = ({
 		isOn: defaultIsOn,
 		isPoppingState: false,
 		searchResults: [],
-		searchedTerm: '',
 		totalResults: 0,
 	});
 
--- a/assets/js/api-search/index.js
+++ b/assets/js/api-search/index.js
@@ -105,8 +105,8 @@ export const ApiSearchProvider = ({
 		isOn: defaultIsOn,
 		isPoppingState: false,
 		searchResults: [],
-		searchedTerm: '',
 		totalResults: 0,
+		isFirstSearch: true,
 	});
 
 	/**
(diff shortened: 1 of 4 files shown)


assets/js/api-search/index.js
```diff
@@ -105,8 +105,8 @@ export const ApiSearchProvider = ({
 		isOn: defaultIsOn,
 		isPoppingState: false,
 		searchResults: [],
-		searchedTerm: '',
 		totalResults: 0,
+		isFirstSearch: true,
```
JakePT, line 109: @oscarssanchez Can we keep `searchedTerm` but just rename it to `searchTerm`. Current spelling is a mistake, but it should still be there.

JakePT (changes requested):

JakePT (approved):

PR #3616, opened 2023-08-25
Update Composer Dependencies to fix Fatal Error

### LINKED ISSUE #3615: BUG: Issue with very outdated PSR\Container dependency
### Describe the bug

I get the following error when trying to use ElasticPress with some custom code which uses a more recent version of `Psr\Container`

```
...
must be compatible with Psr\Container\ContainerInterface::get($id)
...
```

### Steps to Reproduce

1. Include `psr/container:1.1` or greater into your project using Composer
2. Enable the ElasticPress plugin
3. You will see a fatal error

_No response_

### Environment information

_No response_

_No response_

### Code of Conduct

junaidbhura:
### Description of the Change

This bumps up the platform to `7.4` - which is now the minimum supported version for WordPress as well: https://en-au.wordpress.org/about/requirements/

Closes #3615

### How to test the Change

N/A - just updated a few dependencies.

### Changelog Entry

> Updated PHP version to `7.4`

### Checklist:

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -56,7 +56,7 @@
       "dealerdirect/phpcodesniffer-composer-installer": true
     },
     "platform": {
-      "php": "7.0"
+      "php": "7.4"
     }
   }
 }
--- a/composer.lock
+++ b/composer.lock
@@ -4,43 +4,41 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
-    "content-hash": "93f6532f7aa0270c6df2f16091b31b4f",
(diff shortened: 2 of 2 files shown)


felipeelia:
Thanks for opening the PR @junaidbhura but unfortunately we can't update that yet.

As stated on that page you linked: _WordPress also works with PHP 7.0+_, and that was updated in WP 6.3, and ElasticPress minimum WP version is still 5.6. Also, our minimum PHP required version is 7.0.

That said, we are planning to update the minimum PHP version to 7.4 when we release ElasticPress 5.0 in the next few months. Until then we'll have to find an alternative solution to the problem outlined in #3615.

felipeelia:
Closing in favor of #3668

PR #3635, opened 2023-09-06
Improving phpDoc

renatonascalves:
### Description of the Change

phpDoc improvements.

### How to test the Change

You can run [phpstan](https://phpstan.org/) in the project and confirm the errors.

### Changelog Entry

> Changed - Updated phpDoc entries.

### Credits

@renatonascalves 

### Checklist:

### DIFF
--- a/includes/classes/Indexable.php
+++ b/includes/classes/Indexable.php
@@ -12,7 +12,6 @@ namespace ElasticPress;
 
 use ElasticPress\Elasticsearch as Elasticsearch;
 use ElasticPress\SyncManager as SyncManager;
-use ElasticPress\QueryIntegration as QueryIntegration;
 
 if ( ! defined( 'ABSPATH' ) ) {
 	exit; // Exit if accessed directly.
@@ -49,7 +48,7 @@ abstract class Indexable {
 	 * Instance of QueryIntegration. This should handle integrating with a default
 	 * WP query.
 	 *
-	 * @var QueryIntegration
+	 * @var object
 	 * @since  3.0
 	 */
 	public $query_integration;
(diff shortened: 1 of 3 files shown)


includes/classes/Indexable.php
```diff
@@ -49,7 +48,7 @@ abstract class Indexable {
 	 * Instance of QueryIntegration. This should handle integrating with a default
 	 * WP query.
 	 *
-	 * @var QueryIntegration
+	 * @var object
```
renatonascalves, line 51: None of the core features extends this class. So this object here is incorrect.

One gets this error while extending the class.

````
Property ElasticPress\Indexable::$query_integration (ElasticPress\QueryIntegration) does not accept Project\QueryIntegration. 
```

includes/classes/Indexables.php
```diff
@@ -152,7 +152,7 @@ public function get_all( $global = null, $slug_only = false, $status = 'active'
 	/**
 	 * Return singleton instance of class
 	 *
-	 * @return object
+	 * @return self
```
renatonascalves, line 155: Intellisense tools will make better use of this with `self`. Currently, one gets this:
```
Call to an undefined method object::get().       
```

includes/classes/SyncManager.php
```diff
@@ -69,8 +69,9 @@ public function __construct( $indexable_slug ) {
 	/**
 	 * Add an object to the sync queue.
 	 *
-	 * @param  id $object_id object ID to sync
-	 * @since  3.1.2
+	 * @since 3.1.2
+	 *
+	 * @param int $object_id Object ID to sync.
```
renatonascalves, line 74: Fixes this error:
```
Parameter #1 $object_id of method ElasticPress\SyncManager::add_to_queue() expects ElasticPress\id, int given.  
```

PR #3643, opened 2023-09-19
Replace Sync AJAX callback with REST API endpoint.

### LINKED ISSUE #3627: Replace Sync AJAX callback with REST API endpoint

To support more features on the Sync page, we should replace the admin-ajax.php callbacks with a REST API endpoint for syncing, with support for more at least these arguments supported by the CLI:

- `offset`
- `indexables`
- `post-type`
- `include`
- `upper-limit-object-id`
- `lower-limit-object-id`

The endpoint should support `POST` requests for triggering or continuing a sync, and `GET` for getting the status of a sync in progress. Both requests should return the sync status, including:

- The sync trigger (eg. manual, CLI).
- Log messages.
- Total items to be synced.
- Total items synced.
- Sync status.

### Designs

_No response_

_No response_

### Code of Conduct

JakePT:
### Description of the Change
Replaces the admin-ajax.php callbacks used by the Sync page with a REST API. This also adds support for additional arguments that are passed to the index helper:

- `include`
- `indexables`
- `lower_limit_object_id`
- `offset`
- `post_type`
- `put_mapping`
- `upper_limit_object_id`

This PR updates the Sync page to use the REST API, but there is no UI for the additional parameters at this stage.

Closes #3627

### How to test the Change
- Sync should function as normal.
- `POST` requests to `wp-json/elasticpress/v1/sync` with the above parameters should perform the expected sync.
- `GET` requests to `wp-json/elasticpress/v1/sync` should return the status of an in progress sync.
- `DELETE` requests to `wp-json/elasticpress/v1/sync` should cancel an in progress sync.

### Credits
Props @JakePT 

### Checklist:

### DIFF
--- a/includes/classes/Screen/Sync.php
+++ b/includes/classes/Screen/Sync.php
@@ -141,6 +141,7 @@ class Sync {
 	/**
 	 * Register REST API routes.
 	 *
+	 * @since 5.0.0
 	 * @return void
 	 */
 	public function register_rest_routes() {
--- a/includes/classes/Screen/Sync.php
+++ b/includes/classes/Screen/Sync.php
@@ -8,11 +8,12 @@
 
 namespace ElasticPress\Screen;
 
-use ElasticPress\IndexHelper;
-use ElasticPress\Screen;
-use ElasticPress\Utils;
-use ElasticPress\Stats;
 use ElasticPress\Elasticsearch;
+use ElasticPress\IndexHelper;
+use ElasticPress\REST;
+use ElasticPress\Screen;
+use ElasticPress\Stats;
+use ElasticPress\Utils;
 
 if ( ! defined( 'ABSPATH' ) ) {
 	exit; // Exit if accessed directly.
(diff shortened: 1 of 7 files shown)


includes/classes/Screen/Sync.php
```diff
@@ -230,24 +139,12 @@ public function admin_enqueue_scripts() {
...
-			case 'error':
-				wp_send_json_error( $message );
-				break;
-
-			default:
-				wp_send_json( [ 'data' => $message ] );
-				break;
-		}
-		exit;
+	public function register_rest_routes() {
```
felipeelia, line 147: This will need a `@since` tag @JakePT

felipeelia (changes requested):
@JakePT I added a small change that we will need. IMHO initializing the REST Sync class in Screen Sync seems out of place... now that we will have it available, that should be a standalone thing, so people who want to build on top of it would be able to (and not depend on any "screen" entity.) Thoughts?

includes/classes/Screen/Sync.php
JakePT: Done.

PR #3647, opened 2023-09-21
Store additional syncs info

### LINKED ISSUE #3628: Log more Syncs

Currently the status of only 1 previous manual sync, and 1 previous CLI sync, are recorded. To support displaying a more detailed sync history on the Sync page we should log X (5, 10, 15?) more syncs, including their:

- Status (eg. success/failure).
- Date and time.
- Trigger (eg. manual/CLI).
- No. of errors.

### Designs

_No response_

_No response_

### Code of Conduct

felipeelia:
### Description of the Change

Closes #

### Changelog Entry

> Added - New feature
> Removed - Feature
> Fixed - Bug fix
> Security - Vulnerability

### Credits

### Checklist:

### DIFF
--- a/includes/classes/IndexHelper.php
+++ b/includes/classes/IndexHelper.php
@@ -144,6 +144,7 @@ class IndexHelper {
 			'start_date_time'   => $start_date_time ? $start_date_time->format( DATE_ATOM ) : false,
 			'starting_indices'  => $starting_indices,
 			'messages_queue'    => [],
+			'trigger'           => 'manual',
 			'totals'            => [
 				'total'      => 0,
 				'synced'     => 0,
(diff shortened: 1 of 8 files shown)


includes/classes/IndexHelper.php
```diff
@@ -870,24 +871,64 @@ protected function index_cleanup() {
...
+		 * @hook ep_syncs_to_keep_info
+		 * @param {int} $number Number of past syncs to keep info
+		 * @return {int} New number
+		 */
+		$syncs_to_keep = (int) apply_filters( 'ep_syncs_to_keep_info', 5 - 1 );
+
+		$last_syncs = array_slice( $last_syncs, 0, $syncs_to_keep );
+		array_unshift( $last_syncs, $last_sync_info );
+
+		Utils\update_option( 'ep_last_syncs', $last_syncs, false );
```
JakePT: @felipeelia Nitpick, but can we use something like `ep_previous_syncs` or `ep_sync_history` or something like that? "last" implies the final sync, so there wouldn't be more than one "last" sync.

includes/classes/IndexHelper.php
```diff
@@ -870,24 +871,64 @@ protected function index_cleanup() {
...
+
+		/**
+		 * Filter the number of past syncs to keep info
+		 *
+		 * @since  5.0.0
+		 * @hook ep_syncs_to_keep_info
+		 * @param {int} $number Number of past syncs to keep info
+		 * @return {int} New number
+		 */
+		$syncs_to_keep = (int) apply_filters( 'ep_syncs_to_keep_info', 5 - 1 );
```
JakePT: @felipeelia Should we just pass the `5` there and then `-1` later in `array_slice`? Otherwise you need to pass 1 less than the actual number of syncs you want to keep to the filter.

includes/classes/IndexHelper.php
```diff
@@ -144,6 +144,7 @@ protected function build_index_meta() {
 			'start_date_time'   => $start_date_time ? $start_date_time->format( DATE_ATOM ) : false,
 			'starting_indices'  => $starting_indices,
 			'messages_queue'    => [],
+			'trigger'           => 'manual',
```
JakePT, line 147: @felipeelia Since a sync could be done manually either by the dashboard or WP CLI, could we use `dashboard` as the default? In the UI we would need to infer that this is a dashboard sync anyway, so it would be clearer if the key matched.

Also, would this be a good time to set `trigger` to `cli` for WP CLI syncs?

JakePT (changes requested):

JakePT (approved):

PR #3655, opened 2023-09-26
Update Features to declare settings as JSON

felipeelia:
### Description of the Change

Closes #

### Changelog Entry

> Added - New feature
> Removed - Feature
> Fixed - Bug fix
> Security - Vulnerability

### Credits

### Checklist:

### DIFF
--- a/includes/classes/Feature.php
+++ b/includes/classes/Feature.php
@@ -519,16 +519,20 @@ abstract class Feature {
 	 * @return string
 	 */
 	public function get_json() {
+		$requirements_status = $this->requirements_status();
+
 		$feature_desc = [
-			'slug'            => $this->slug,
-			'title'           => $this->get_title(),
-			'shortTitle'      => $this->get_short_title(),
-			'summary'         => $this->summary,
-			'docsUrl'         => $this->docs_url,
-			'defaultSettings' => $this->default_settings,
-			'order'           => $this->order,
-			'isAvailable'     => $this->is_available(),
-			'settingsSchema'  => $this->get_settings_schema(),
+			'slug'              => $this->slug,
+			'title'             => $this->get_title(),
+			'shortTitle'        => $this->get_short_title(),
+			'summary'           => $this->summary,
+			'docsUrl'           => $this->docs_url,
+			'defaultSettings'   => $this->default_settings,
+			'order'             => $this->order,
+			'isAvailable'       => $this->is_available(),
+			'reqStatusCode'     => $requirements_status->code,
+			'reqStatusMessages' => $requirements_status->message,
+			'settingsSchema'    => $this->get_settings_schema(),
 		];
 		return wp_json_encode( $feature_desc );
 	}
--- a/includes/classes/Feature.php
+++ b/includes/classes/Feature.php
@@ -119,6 +119,14 @@ abstract class Feature {
 	protected $is_visible = true;
 
 	/**
+	 * Settings description
+	 *
+	 * @since 5.0.0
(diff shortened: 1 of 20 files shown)


includes/classes/Feature/DidYouMean/DidYouMean.php
```diff
@@ -415,4 +417,35 @@ public function the_output( $query = null ) {
...
+		$this->settings_schema = [
+			[
+				'default' => false,
+				'key'     => 'search_behavior',
+				'label'   => __( 'Search behavior when no result is found', 'elasticpress' ),
+				'options' => [
+					[
+						[
+							'label' => __( 'Display the top suggestion', 'elasticpress' ),
+							'key'   => 0,
```
JakePT: @felipeelia I think this should be `value`? That's what's used elsewhere and it makes it easier to pass into the various core components that accept options.

includes/classes/Feature.php
```diff
@@ -503,4 +511,60 @@ public function is_available() : bool {
...
+	public function get_json() {
+		$feature_desc = [
+			'slug'            => $this->slug,
+			'title'           => $this->get_title(),
+			'shortTitle'      => $this->get_short_title(),
+			'summary'         => $this->summary,
+			'docsUrl'         => $this->docs_url,
+			'defaultSettings' => $this->default_settings,
+			'order'           => $this->order,
+			'isAvailable'     => $this->is_available(),
```
JakePT: @felipeelia Just want to double check that a boolean `isAvailable` is sufficient for what we need on the front-end, or whether we need the full requirements status. I believe there's currently a different status color and message based on the requirements code, which wouldn't be possible with just this boolean.

includes/classes/Feature/Autosuggest/Autosuggest.php
```diff
@@ -855,6 +857,76 @@ public function intercept_remote_request() {
...
+
+		if ( Utils\is_epio() ) {
+			$this->maybe_add_epio_settings_schema();
+			return;
+		}
+
+		$set_in_wp_config = defined( 'EP_AUTOSUGGEST_ENDPOINT' ) && EP_AUTOSUGGEST_ENDPOINT;
+
+		$this->settings_schema[] = [
+			'editable' => ! $set_in_wp_config,
```
JakePT: @felipeelia Can we use `readonly` for this? It maps a little better to how the field will be treated on the front end.

JakePT (changes requested):
Added a few comments.

includes/classes/Feature/DidYouMean/DidYouMean.php
felipeelia: Thanks @JakePT, fixed in [c105e34](https://github.com/10up/ElasticPress/pull/3655/commits/c105e34d5787be58532feb7c132ca9c75fc00231).

includes/classes/Feature/Autosuggest/Autosuggest.php
felipeelia: Thanks @JakePT, adjusted in [c105e34](https://github.com/10up/ElasticPress/pull/3655/commits/c105e34d5787be58532feb7c132ca9c75fc00231).

includes/classes/Feature.php
felipeelia: I kept that (imagining it'll make your life a bit easier) but I'm now also including the code and the message(s) from the requirement_status method.

JakePT (approved):

PR #3657, opened 2023-09-26
New Sync page with advanced options

### LINKED ISSUE #3640: Allow users to set sync parameters on the Sync page

ElasticPress syncing supports many additional arguments when using the CLI, such as which content types to index, which post types to index, whether to skip certain posts etc. It should be possible to configure some of these parameters using a UI on the Sync page.

### Designs

_No response_

_No response_

### Code of Conduct

JakePT:
### Description of the Change
Introduces a refreshed Sync page with an Advanced Options section for setting additional parameters for a sync, bringing some of the functionality from the WP CLI command to the UI.

elasticpress test_wp-admin_admin php_page=elasticpress-sync (1)

This PR includes a number things:

  - Updates the `SyncProvider` component to support additional arguments when syncing. Auto syncing and initial sync management have been removed so that they can be managed at the UI level.
  - Adds a `SettingsScreenProvider` component that sets a common layout for ElasticPress settings screens and provides a method for displaying toast/snackbar type notifications.
  - Adds a `SyncSettingsProvider` component for managing the state of selected sync settings across components.
  - Combines the separate Sync and Delete and Sync panels into a single panel with a checkbox to perform a Delete and Sync.
  - Adds the date and time to log messages.
  - Adds Clear log and Copy log to clipboard buttons.
  - Adds options for Indexables, Post Types, Object IDs, and an Offset.

The only regression against the current Sync page is that the previous sync status has been removed. The sync history will be added in a separate PR, now that #3647 has been merged.

This PR updates existing E2E tests for Syncing, but tests for new options will come separately.

Closes #3640

### How to test the Change
1. Visit _ElasticPress > Sync_.
2. It should be possible to Start, Pause, Resume, and Stop a sync as before.
3. Checking _Delete all data and start fresh sync_ should delete all data and send mapping when syncing.
4. The sync results and log should reflect the choices made in the Advanced Options section.

### Changelog Entry
Changed - Combined the separate Sync and Delete and Sync panels into a single panel with a checkbox to perform a Delete and Sync.
Added - Added the date and time to log messages on the Sync Page.
Added - Added Clear log and Copy log to clipboard buttons to the Sync page.
Added - Added Advanced Options to the Sync page for setting Content Types, Post Types, Object IDs, and an Offset for a sync.

### Credits

Props @JakePT @felipeelia Apurv Ray @brandwaffle @anjulahettige 

### Checklist:


assets/js/sync-ui/components/include.js
```diff
@@ -0,0 +1,60 @@
...
+		const transformedValue = parseInt(value, 10);
+
+		return transformedValue ? transformedValue.toString() : '';
+	};
+
+	return (
+		<div className="ep-sync-advanced-control">
+			<FormTokenField
+				disabled={isSyncing}
+				help={__('Help!', 'elasticpress')}
```
felipeelia: @JakePT Where would this "Help!" be displayed?

includes/classes/Screen/Sync.php
```diff
@@ -55,85 +56,34 @@ public function admin_enqueue_scripts() {
...
-		wp_enqueue_style( 'wp-edit-post' );
-
 		wp_enqueue_style(
 			'ep_sync_style',
-			EP_URL . 'dist/css/sync-styles.css',
-			Utils\get_asset_info( 'sync-styles', 'dependencies' ),
-			Utils\get_asset_info( 'sync-styles', 'version' )
+			EP_URL . 'dist/css/sync-script.css',
+			[ 'wp-components', 'wp-edit-post' ],
+			Utils\get_asset_info( 'sync-script', 'version' )
```
felipeelia, line 63: mind explaining this change?

assets/js/sync-ui/components/include.js
JakePT: @felipeelia Good catch. I put that there to find out, and the answer is nowhere, so I'll remove it.

includes/classes/Screen/Sync.php
JakePT: @felipeelia So when I created `SettingsScreenProvider` it needed to come with some styles. Since that component will be used by multiple screens I had to figure out what to do with those styles. I could have enqueued it as its own stylesheet, but that seemed excessive since it was very short, and would require maintaining the enqueued stylesheet so that it was enqueued on the right screens, which means that it would need to either be enqueued separately by each screen, or registered somewhere else entirely and enqueued as a style dependency. It would all be rather messy.

The solution I went with was importing the stylesheet into to the script. This meant that any other component/app that imported `SettingsScreenProvider` would import the necessary styles, which worked nicely. When you import styles into a script, toolkit will automatically build a stylesheet with the same name as the script, with any imported styles. Since this was being generated anyway, it made sense to use that as the main stylesheet for the screen, so I removed the sync styles as a separate package.json entry and imported them into the Sync UI script. So now toolkit creates a stylesheet with all the imported styles, including the `SettingsScreenProvider` styles. That's what's being enqueued here.

PR #3664, opened 2023-09-28
Display sync history

JakePT:
### Description of the Change
Displays the sync history on the sync page. Currently limited to Completed syncs (with errors or not).

elasticpress test_wp-admin_admin php_page=elasticpress-sync

### How to test the Change
- Visit _ElasticPress > Sync_. A _Sync history_ panel should be visible with at least one previous sync (if coming from < 5.0.0).
- When a new sync is completed it should be added to the list automatically without a refresh.
- If a a sync is triggered via WP CLI that should be reflected in the the sync title.
- The history should not appear for the initial sync on a site, or after deleting an index.
- The _Delete and sync..._ and _Advanced options_ panels should only be visible if there is a sync history.

### Changelog Entry
Added - Added a sync history to the Sync page.

### Credits
Props @JakePT @felipeelia @brandwaffle @anjulahettige Apurv Ray

### Checklist:


felipeelia:
@JakePT I gave this a try and it seems we are not capturing if a sync failed or not. Adding this snippet to a mu-plugin makes the `[illegal_argument_exception] Limit of total fields [5000] in index [elasticpresstest-post-1] has been exceeded` error to happen:

```php
add_filter(
	'ep_prepare_meta_data',
		}
		return $post_meta;
	},
	10,
	2
);
```

I can see the errors in the log. If I wait for it to complete and reload the page I can see the new entry but it says it was successful.

felipeelia (approved):

felipeelia:
@JakePT we talked about this and it is enough to say it was triggered due to a setting change, we do not need to specify the change for now.

PR #3676, opened 2023-10-03
Add feature settings REST API

JakePT:
### Description of the Change
Adds a REST API endpoint at `/wp-json/elasticpress/v1/features` for updating feature settings. The endpoint uses the settings schema added to features by #3655 to provide validation for updating the `ep_feature_settings` via a `PUT` request.

This PR also creates a `Dashboard` screen class to register the endpoint (for consistency with Sync). The name dashboard was for consistency with the `dashboard-page.php` file used to render it.

### How to test the Change
1. Make an `OPTIONS` request to `/wp-json/elasticpress/v1/features`. The feature settings and their types and possible values should be shown as supported `args`.
2. Copy the saved `ep_feature_settings` option from the database and unserialize it as JSON. Make modifications and send it as the body in a `PUT` request to `/wp-json/elasticpress/v1/features`. Valid changes according to the definitions in the `OPTIONS` response should update the feature settings, and be reflected in the admin, while changes that are invalid should result in a 400 error.

### Changelog Entry
Added - Added a REST API endpoint at `elasticpress/v1/features` for updating feature settings.

### Credits
Props @JakePT 

### Checklist:

### DIFF
--- a/includes/classes/Screen/Dashboard.php
+++ b/includes/classes/Screen/Dashboard.php
@@ -1,0 +1,41 @@
+<?php
+/**
+ * Dashboard screen class.
+ *
+ * @since  3.6.0
+ * @package elasticpress
+ */
+
+namespace ElasticPress\Screen;
+
+use ElasticPress\REST;
+
+if ( ! defined( 'ABSPATH' ) ) {
+	exit; // Exit if accessed directly.
+}
+
+/**
+ * Dashboard screen.
+ *
+ * @since 5.0.0
+ * @package ElasticPress
+ */
+class Dashboard {
+	/**
+	 * Initialize class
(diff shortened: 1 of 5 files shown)


includes/classes/Screen/Dashboard.php
```diff
@@ -0,0 +1,41 @@
+<?php
+/**
+ * Dashboard screen class.
+ *
+ * @since  3.6.0
```
felipeelia:
```suggestion
 * @since 5.0.0
```

felipeelia (changes requested):
Everything looks good @JakePT, the only thing I think we need to change is a `@since` tag.

PR #3693, opened 2023-10-10
Index a full srcset for the post thumbnail

### LINKED ISSUE #3449: Index a full srcset for the post thumbnail

To support the Instant Results feature we currently we index a single image URL (and dimensions) for the post thumbnail, whose size is determined by the `ep_post_thumbnail_image_size` filter. The problem with this is that it's a single size that is not necessarily appropriate for use by other features, such as Autosuggest (particularly when using WooCommerce and the larger thumbnail size is used). This was basically because it was the only cropped size whose layout could be predictable.

However, with support for `object-fit` now widely available we can crop images with CSS. Therefore I think it makes more sense to index a `srcset` of all the uncropped sizes of the image and use the `sizes` attribute of the `img` tag to let the browser choose the optimal size. By setting a proper `sizes` attribute a developer can get the browser to use the size of image that they want, rather than needing to filter the image size so that it's changed for all uses.

### Designs

_No response_

_No response_

### Code of Conduct

burhandodhy:
### Description of the Change

This PR indexes the post srcset. It also makes changes to the mapping. By default, the `dynamic` property is set to `true`, which means the srcset will be automatically indexed even if the index has the old mapping.

Closes #3449

### How to test the Change

1. Index the post.
2. Verify if thumbnail has `srcset` in Elasticsearch

### Changelog Entry

> Changed - Index `srcset` for post thumbnail

### Credits

Props @burhandodhy @JakePT 

### Checklist:


KalleVuorjoki:
This seems very useful feature, is there something missing here or is rebase all that is needed?

felipeelia:
@burhandodhy wouldn't this need a change in the mapping too?

PR #3700, opened 2023-10-17
Add Date Filter

### LINKED ISSUE #3634: Filter by Date

Filtering by date has become one of the commonly asked features for the plugin. We can add a new filter type to implement it. 

For the initial version we can have some initial fixed options like "Last 3 months", "Last 6 months", and "Last 12 months", filterable via code. In addition to those options, the block should also have a "Custom" option, that would display two date fields ("From" and "To"), giving the final user the option to choose a completely different range.

In an eventual second phase, we can add the option of changing the date field (published vs. modified, or a meta field), and also controls to change the default options.

### Designs

_No response_

_No response_

### Code of Conduct

burhandodhy:
### Description of the Change

This PR adds a new Filter block that is the "Date Filter"

Closes #3634

### Changelog Entry

> Added - New Date Filter Block

### Credits

Props @burhandodhy 

### Checklist:


includes/classes/Feature/Facets/Facets.php
```diff
@@ -219,6 +221,19 @@ public function front_scripts() {
...
+		/**
+		 * Filter the data passed to the facets script.
+		 * This is used by facet types to add their own data to the script.
+		 *
+		 * @hook ep_facets_script_data
+		 * @since 5.0.0
+		 * @param  {array} $data Data passed to the script.
+		 * $return {array} New data passed to the script.
+		 */
+		$data = apply_filters( 'ep_facets_script_date', [] );
```
felipeelia: Is it related to date or data, @burhandodhy?

includes/classes/Feature/Facets/Types/Date/FacetType.php
```diff
@@ -0,0 +1,235 @@
...
+
+		/**
+		 * Filter the options for the date facet.
+		 *
+		 * @since 5.0.0
+		 *
+		 * @param {array} $options The options for the date facet.
+		 * @return {array} The options for the date facet.
+		 */
+		return apply_filters( 'ep_facet_date_options', $options );
```
felipeelia, line 231: It would be great if we could add a code example to this docblock @burhandodhy.

felipeelia (commented):
@burhandodhy I gave it a quick look and overall it looks good, just added a few comments. Let's also ask @JakePT to give it a look.

assets/css/facets.css
```diff
@@ -146,3 +146,41 @@
...
+	max-width: 400px;
+}
+
+.facet-date-form.hide {
+	display: none;
+}
+
+.ep-date-range-picker label {
+	margin-right: 0.5rem;
+}
```
JakePT: Can we try and match the styles of the other filter blocks a little more?

Screenshot 2023-10-18 at 5 38 14 pm

In particular:

 - The colour of the unselected radio button.
 - In the selected state can we include a white dot in the same style as the checkbox:
   Screenshot 2023-10-18 at 5 40 00 pm
 - The size of the radio button and the space between it and the label.
 - The Custom option seems to have an additional space in front of the label.

assets/js/blocks/facets/date/Edit.js
```diff
@@ -0,0 +1,50 @@
...
+	const blockProps = useBlockProps();
+	const { attributes, name, setAttributes } = props;
+	const { displayCustomDate } = attributes;
+
+	return (
+		<>
+			<InspectorControls>
+				<PanelBody>
+					<ToggleControl
+						label={__('Display custom date', 'elasticpress')}
```
JakePT: Nitpick, but can we be a little more specific here, eg. "Display custom date option" or "Display custom option"?

assets/js/blocks/facets/date/index.js
```diff
@@ -0,0 +1,20 @@
+/**
+ * WordPress dependencies.
+ */
+import { registerBlockType } from '@wordpress/blocks';
+
+/**
+ * Internal dependencies.
+ */
+import icon from '../common/icon';
+import edit from './edit';
```
JakePT, line 10: It looks like the Edit.js file has been committed with an uppercase E. Can you commit it with the lowercase filename? This might help: https://stackoverflow.com/a/20907647

package.json
```diff
@@ -67,6 +67,7 @@
       "facets-meta-range-block-script": "./assets/js/blocks/facets/meta-range/index.js",
       "facets-post-type-block-script": "./assets/js/blocks/facets/post-type/index.js",
       "facets-meta-range-block-view-script": "./assets/js/blocks/facets/meta-range/view.js",
+      "facets-date-script": "./assets/js/blocks/facets/date/index.js",
```
JakePT: Just for consistency with the other blocks, can we call this `facets-date-block-script`?

assets/js/facets.js
```diff
@@ -50,3 +50,62 @@ facets.forEach((facet) => {
 		}, 200),
 	);
 });
+
+/**
+ * Initializes the date facet functionality.
+ *
+ */
+const initDateFacet = () => {
```
JakePT: Since this is only relevant to the Date block, can we put this in a viewScript, like the Meta Range block?

includes/classes/Feature/Facets/Types/Date/Renderer.php
```diff
@@ -0,0 +1,168 @@
...
+								<input type="date" name="ep-end-date" value="<?php echo esc_attr( $custom_dates[1] ?? '' ); ?>">
+							</div>
+						</div>
+						<button type="submit">Filter</button>
+					</form>
+				</li>
+				<?php if ( ! empty( $selected_filters[ $facet_type->get_filter_type() ] ) ) : ?>
+					<li>
+						<a aria-label="Clear" href="<?php echo esc_url( $this->get_clear_filter_url() ); ?>" rel="nofollow" class="ep-clear-filters">
+							<?php esc_html_e( 'Clear', 'elasticpress' ); ?>
```
JakePT: Could we lay this out similarly to the Meta Range block, for consistency?

Screenshot 2023-10-18 at 5 51 27 pm

Although, in the case of the meta range block the Filter button is always visible, so I'm wondering if it wouldn't be better to have an "all" radio button option, instead of a clear link, since the other list filters don't have one. What do you think @felipeelia?

Also, can we add the `wp-element-button` class to the button, for themes that style this element, and for consistency with the meta range block, which uses it.

includes/classes/Feature/Facets/Facets.php
```diff
@@ -219,6 +221,19 @@ public function front_scripts() {
...
+
+		wp_localize_script( 'elasticpress-facets', 'epFacets', $data );
```
JakePT: If we move the script to its own file, as I suggest elsewhere, we should make this date block specific, like `epDateFacet`.

assets/js/blocks/facets/date/Edit.js
```diff
@@ -0,0 +1,50 @@
...
+
+const FacetDate = (props) => {
+	const blockProps = useBlockProps();
+	const { attributes, name, setAttributes } = props;
+	const { displayCustomDate } = attributes;
+
+	return (
+		<>
+			<InspectorControls>
+				<PanelBody>
```
JakePT: Can we add a "Settings" title here, for consistency with the other filter blocks?

includes/classes/Feature/Facets/Types/Date/Renderer.php
```diff
@@ -0,0 +1,168 @@
...
+				<li>
+					<a aria-label="Apply filter: <?php esc_html_e( 'Custom', 'elasticpress' ); ?>" href="#" rel="nofollow" class="ep-date-range-custom <?php echo $is_custom_date ? 'checked' : ''; ?>">
+						<div class="ep-radio" role="presentation"></div>
+						<?php esc_html_e( 'Custom', 'elasticpress' ); ?>
+					</a>
+
+					<?php
+					$custom_dates = isset( $selected_filters[ $facet_type->get_filter_type() ]['terms'] ) ? array_keys( $selected_filters[ $facet_type->get_filter_type() ]['terms'] ) : [];
+					?>
+					<form class="facet-date-form <?php echo ! $is_custom_date ? 'hide' : ''; ?>">
```
JakePT: Can we prefix our classes with `ep-`. For state classes, can we use `is-hidden`, instead of `hide`.

includes/classes/Feature/Facets/Types/Date/Renderer.php
```diff
@@ -0,0 +1,168 @@
...
+					// phpcs:disable WordPress.Security.EscapeOutput.OutputNotEscaped
+					echo $this->get_facet_item_value_html( $item, $feature->build_query_url( $field_filters ) );
+					// phpcs:enable WordPress.Security.EscapeOutput.OutputNotEscaped
+					?>
+				</li>
+			<?php endforeach ?>
+
+			<?php if ( $this->display_custom_date ) : ?>
+				<li>
+					<a aria-label="Apply filter: <?php esc_html_e( 'Custom', 'elasticpress' ); ?>" href="#" rel="nofollow" class="ep-date-range-custom <?php echo $is_custom_date ? 'checked' : ''; ?>">
```
JakePT: I think this will need some changes for accessibility, only when the filter is not applied:

 - Add `role="button"`
 - Add an `aria-expanded` attribute that changes when the date form is shown.
 - Change the aria-label to something like "Select date range".

When it is already applied it should just be a link and the label should read "Remove filter", like the checkbox filters.

assets/css/facets.css
JakePT: Also, the layout looks a little messed up in Safari:

Screenshot 2023-10-18 at 6 52 21 pm

includes/classes/Feature/Facets/Types/Date/FacetType.php
```diff
@@ -0,0 +1,235 @@
...
+
+			if ( empty( $date_string ) || ! strtotime( $date_string ) ) {
+				$formatted_dates[] = '';
+				continue;
+			}
+
+			$date = new \DateTime();
+			$date->modify( $date_string );
+
+			$formatted_dates[] = $date->format( 'Y-m-d 00:00:00' );
```
JakePT: The "To" date should probably be 23:59:59, otherwise if I filter 2023-01-01 to 2023-01-04 any posts made on the 4th won't be included.

JakePT (changes requested):

includes/classes/Feature/Facets/Types/Date/Renderer.php
felipeelia: the "all" radio button makes more sense, yeah.

felipeelia (changes requested):
@burhandodhy I gave this a try and if I uncheck the "Display custom date option" and save, the block doesn't seem to work anymore. (I see a `Cannot read properties of null (reading 'classList')` error in the console.)

PR #3712, opened 2023-10-18
Updated feature settings UI

JakePT:
### Description of the Change
Makes the following changes to features classes:

- Add full feature description to `summary` property.
- For features that depend on another feature, use `requires_feature` property instead of requirements status.
- Remove ElasticPress.io logo from feature titles, and replace with `is_powered_by_epio` property.
- Returns feature visibility in returned data for JSON.

And adds the updated UI.

elasticpress test_wp-admin_admin php_page=elasticpress (1)

### Changelog Entry
Changed - New feature settings screen

### Credits

Props @JakePT, @felipeelia, Apurv Ray

### Checklist:


assets/js/features/apps/features.js
```diff
@@ -0,0 +1,271 @@
...
+
+	/**
+	 * Notice when syncing now.
+	 */
+	const syncNowNotice = __('Feature settings saved. Starting sync…', 'elasticpress');
+
+	/**
+	 * Success notice.
+	 */
+	const successNotice = __('Feature settings saved.');
```
felipeelia: this should have a domain @JakePT

includes/classes/Screen.php
```diff
@@ -78,13 +78,13 @@ public function setup() {
 		$this->sync_screen        = new Screen\Sync();
 		$this->health_info_screen = new Screen\HealthInfo();
 		$this->status_report      = new Screen\StatusReport();
-		$this->dashboard          = new Screen\Dashboard();
+		$this->features           = new Screen\Features();
```
felipeelia, line 81: You'll need to remove the dashboard attribute and include a new $features one before this method.

JakePT:
@felipeelia The last few test errors seem to be are:

- An error in the Instant Results feature caused by "Category (product_cat)" not being a filter option. This is caused by set_settings_schema() being called before the taxonomy is registered.
- Documents is sometimes failing for some reason. They are passing for me locally, and I think it might just be because not enough time is taken between syncing and searching the PDF.
- Comments seems to occasionally failing because there's 26 comments instead of 27 (or vice verse). That doesn't seem to be specific to this PR.

PR #3720, opened 2023-10-24
Save a draft version of feature settings

felipeelia:
And apply them before a full sync

### Description of the Change

Closes #

### Changelog Entry

> Changed - Depending on the requirements, some feature settings are now saved to be applied after a full sync

### Credits

Props @felipeelia @JakePT 

### Checklist:


includes/classes/REST/Features.php
```diff
@@ -101,42 +101,84 @@ public function check_permission() {
...
 
-			FeaturesStore::factory()->update_feature( $slug, $settings[ $slug ] );
+		foreach ( $current_settings as $slug => $feature ) {
+			FeaturesStore::factory()->update_feature( $slug, $feature );
 		}
 
-		Utils\update_option( 'ep_feature_settings', $settings );
+		if ( $current_settings !== $new_settings ) {
+			FeaturesStore::factory()->save_feature_settings_draft( $new_settings );
+		}
```
JakePT: @felipeelia We aren't deleting draft settings or updating them if this condition is false, which means that, for example, if I enable Instant Results and Sync Later, it creates draft settings with Instant Results active. If I then disable Instant Results without syncing the draft settings still exist with Instant Results enabled.

includes/classes/Features.php
```diff
@@ -316,6 +322,51 @@ public function setup_features() {
...
+	 * @since 5.0.0
+	 */
+	public function apply_draft_feature_settings() {
+		$draft_settings = Utils\get_option( 'ep_feature_settings_draft', false );
+		if ( ! $draft_settings ) {
+			return;
+		}
+
+		Utils\update_option( 'ep_feature_settings', $draft_settings );
+		Utils\delete_option( 'ep_feature_settings_draft' );
```
JakePT: @felipeelia A potential problem here is that this approach means that the draft settings never go through `update_feature`, which means they don't go through the `ep_sanitize_feature_settings` filter or trigger the `post_acivation()` method.

JakePT (changes requested):

includes/classes/Features.php
felipeelia: Changed in [96c160f](https://github.com/10up/ElasticPress/pull/3720/commits/96c160f68434a85bbb7b3d8856986230426b8b07)

includes/classes/REST/Features.php
felipeelia: Changed in [96c160f](https://github.com/10up/ElasticPress/pull/3720/commits/96c160f68434a85bbb7b3d8856986230426b8b07)

PR #3725, opened 2023-10-27
Fix: HTML escape issue on the install page

burhandodhy:
### Description of the Change

This PR fixes the issue where the content of the tooltip on the install page is not showing properly.

Before
image

After fix
image

Closes #

### Changelog Entry

> Fixed - Content was not showing properly on the tooltop on install page
 

### Credits

Props @burhandodhy 

### Checklist:

### DIFF
--- a/includes/partials/install-page.php
+++ b/includes/partials/install-page.php
@@ -122,7 +122,7 @@ $skip_index_url = remove_query_arg( 'ep-skip-features', $skip_install_url );
 													</span>
 												</a>
 												<span role="tooltip" class="a11y-tip__help a11y-tip__help--top">
-													<?php echo wp_kses_post( $feature->summary ); ?>
+													   <?php echo wp_kses( $feature->summary, 'ep-html' ); ?>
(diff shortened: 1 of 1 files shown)


includes/partials/install-page.php
```diff
@@ -122,7 +122,7 @@
 													</span>
 												</a>
 												<span role="tooltip" class="a11y-tip__help a11y-tip__help--top">
-													<?php echo esc_html( $feature->summary ); ?>
+													<?php echo wp_kses_post( $feature->summary ); ?>
```
felipeelia:
```suggestion
													<?php echo wp_kses( $feature->summary, 'ep-html' ); ?>
```

felipeelia (changes requested):
@burhandodhy do you mind applying EP's set of tags instead of posts? Thanks!

includes/partials/install-page.php
burhandodhy: Nice. Its done

PR #3765, opened 2023-11-15
moved index offset reset before message output in indexer cleanup process

### LINKED ISSUE #3764: BUG: Index cleanup process - offset being zeroed too late
I think that in indexer cleanup process (I am testing it with 4.7.2 version), offset resetting happens too late here:

https://github.com/10up/ElasticPress/blob/1844e3186227e0812f2ff70d4a8f80d33331fe58/includes/classes/IndexHelper.php#L845

In case there are any errors during one indexable type indexing (in my case full reindex called from admin dashboard) it will try to output warning before resetting offset. 

The problem is, that output method in the end calls `wp_send_json` which calls `die()` / `wp_die()` and effectively stops further execution. This means that offset will not be zeroed. 

image

Then, when next indexable type indexing starts it picks up offset from where previous indexing finished. So for example when I had 1465 posts to index, terms will start indexing from offset 1465, which in my case leads to terms index being empty. 

image

image

I believe that offset reset should be moved upper, probably right after `$this->index_meta['current_sync_item'] = null;`

pknap:
### Description of the Change

Closes #3764 

### How to test the Change

Run full reindex with multiple indexables when first indexable returns warnings. It should still index further indexables properly.

Before:

image

After:

image

### Changelog Entry

> Fixed - Bug fix #3764 

### Credits

@pknap

### Checklist:

### DIFF
--- a/includes/classes/IndexHelper.php
+++ b/includes/classes/IndexHelper.php
@@ -822,6 +822,7 @@ class IndexHelper {
 		$current_sync_item = $this->index_meta['current_sync_item'];
 
 		$this->index_meta['current_sync_item'] = null;
+		$this->index_meta['offset']            = 0;
 
 		if ( $current_sync_item['failed'] ) {
 			if ( ! empty( $current_sync_item['blog_id'] ) && defined( 'EP_IS_NETWORK' ) && EP_IS_NETWORK ) {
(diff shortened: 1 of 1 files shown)


felipeelia:
Hi @pknap, do you mind fixing the lint issue in 825 so we can merge this PR? (It should be just a matter of aligning the equal sign with line 824.) Thanks!

pknap:
> Hi @pknap, do you mind fixing the lint issue in 825 so we can merge this PR? (It should be just a matter of aligning the equal sign with line 824.) Thanks!

Hi @felipeelia sure thing, should be done now.

PR #3814, opened 2024-01-17
Updated Synonyms UI

JakePT:
### Description of the Change
elasticpress test_wp-admin_admin php_page=elasticpress-synonyms (4)

Introduces a new component-powered settings screen for Synonyms matching the style and functionality of the Features, Sync, Status Report and Weighting screens. The new screen also adds support for Hyponyms, for term with a hierarchical  relationship, and support for many-to-many replacements.

This PR also updates the default example synonyms to reflect the new categories:

```
# Defined synonyms.

runner, running shoe, sneaker, tennis shoe, trainer

# Defined hyponyms.

blue => blue, aqua, azure, cerulean, cyan, ultramarine

# Defined replacements.

supposably => supposedly
flustrated => flustered, frustrated
intensive purposes => intents and purposes
```

### How to test the Change
1. Visit _ElasticPress > Synonyms_.
2. Add a synonym rule. A search for any synonym should return results including itself and any of the other synonyms.
3. Add a hyponym rule. A search for the hypernym should return results including itself and any of its hyponyms, but a search for any of the hyponyms should only return results for itself.
4. Add a replacement rule. A search for any term should return results including any of its replacements, but not itself.
5. Clicking the trash icon should delete a rule.
6. Clicking the pencil icon should open the rule for editing. Changes should be reflected in the list after saving.
7. Clicking multiple checkboxes and then the trash icon at the top of the table should delete all checked rules.
8. After clicking multiple checkboxes, and then trashing a single rule, the same items should be checked.

### Changelog Entry
Added - A new settings screen for synonyms, with the the ability to bulk delete synonyms, support for many-to-many replacements, and a new type of synonym for terms with a hierarchical relationship, called hyponyms.

### Credits

Props @JakePT, @apurvrdx1

### Checklist:


includes/classes/Feature/Search/Synonyms.php
```diff
@@ -818,4 +863,14 @@ private function update_synonym_post( $content ) {
 			true
 		);
 	}
+
+	/**
+	 * Setup REST endpoints
+	 *
+	 * @since  3.2
```
felipeelia: Shouldn't this be 5.1.0?

includes/classes/REST/Synonyms.php
```diff
@@ -0,0 +1,119 @@
+<?php
+/**
+ * Features REST API Controller
+ *
+ * @since 5.0.0
```
felipeelia: Shouldn't this be 5.1.0?

felipeelia (changes requested):

includes/classes/Feature/Search/Synonyms.php
JakePT: @felipeelia Fixed.

includes/classes/REST/Synonyms.php
JakePT: @felipeelia Fixed.

PR #3863, opened 2024-03-17
Increment block API versions.

### LINKED ISSUE #3853: Adopt Block API version 3

As of WordPress 6.3:

>The post editor will be iframed if all registered blocks have a Block API version 3 or higher and the editor has no classic meta boxes below the blocks. Adding version 3 support means that the block should work inside an iframe, though the block may still be rendered outside the iframe if not all blocks support version 3.

Since ElasticPress' blocks use version 2, it will prevent this behaviour for sites that are otherwise compatible.

The only work involved is increasing the version number and verifying that the blocks work as expected in an iframe. ElasticPress' blocks do not do anything that wouldn't work in an iframe, so I don't anticipate any more work than that.

### Designs

_No response_

_No response_

### Code of Conduct

JakePT:
### Description of the Change
Increments the Block API version of blocks to 3, to indicate support for working in an iframed editor.

Closes #3853

### How to test the Change
All blocks should function as normal.

### Changelog Entry
Changed - Increments the Block API version of blocks to 3, to indicate support for working in an iframed editor.

### Credits

Props @JakePT 

### Checklist:


felipeelia:
@JakePT it seems all e2e tests failed in WP latest. Do you mind giving it a look?

PR #3870, opened 2024-03-20
Replace `no-response` with `stale` action

jeffpaul:
### Description of the Change
This PR replaces the [lee-dohm/no-response](https://github.com/lee-dohm/no-response) action with the [actions/stale](https://github.com/actions/stale) action. I had noticed recently that there were lots of No Response action failures, and after looking into it appears that the former action was no longer supported, throwing node 12/16 warnings, and the hope in changing to a GitHub first-party action will be improved stability (aka action runs not failing randomly) and proper node version support.

### How to test the Change
Check the updated run on this PR

### Credits
Props @jeffpaul.

### Checklist:


jeffpaul:
I added back the custom EP close message.  Note that the updated action waits 3 days to mark an issue as stale and then another 3 days to close.  You can configure those timelines in the YML file as well as the related messages for each stale/close comment.

PR #3875, opened 2024-03-22
Apply ngram_fields on search query when ajax and weighting enable

### LINKED ISSUE #3832: BUG: Inconsistent search results when calling the same function via PHP and Ajax
### Describe the bug

We came across an issue where calling the same function via PHP and AJAX resulted in different results. This only happens when the "Autosuggest" feature is activated.

So further investigation revealed that Autosuggest only adds the `ngram` field to public request https://github.com/10up/ElasticPress/blob/develop/includes/classes/Feature/Autosuggest/Autosuggest.php#L213

We need to pass the `ajax` as a type in `is_integrated_request` function. So the ngram field would also be added for the ajax request.  

### Steps to Reproduce

1. Add `add_filter( 'ep_ajax_wp_query_integration', '__return_true' );` and `add_filter( 'ep_enable_do_weighting', '__return_true' );`
2. Call the function on page load and same function via the ajax call
3. Compare the query body

_No response_

### Environment information

_No response_

_No response_

### Code of Conduct

burhandodhy:
### Description of the Change

This PR adds the ngram fields to query when ajax integration and weighting is enabled. 

Closes #3832 

### Changelog Entry

> Fixed - Inconsistent search results when calling the same function via PHP and Ajax

### Credits

Props @burhandodhy 

### Checklist:

### DIFF
--- a/includes/classes/Feature/Autosuggest/Autosuggest.php
+++ b/includes/classes/Feature/Autosuggest/Autosuggest.php
@@ -195,7 +195,7 @@ class Autosuggest extends Feature {
 	 * @return array $query adjusted ES Query arguments
 	 */
 	public function adjust_fuzzy_fields( $query, $post_type, $args ) {
-		if ( ! Utils\is_integrated_request( $this->slug, [ 'public' ] ) || empty( $args['s'] ) ) {
+		if ( ! Utils\is_integrated_request( $this->slug, [ 'public', 'ajax' ] ) || empty( $args['s'] ) ) {
 			return $query;
 		}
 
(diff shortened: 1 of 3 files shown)


includes/classes/Feature/Autosuggest/Autosuggest.php
```diff
@@ -195,7 +195,7 @@ public function set_fuzziness( $fuzziness, $search_fields, $args ) {
 	 * @return array $query adjusted ES Query arguments
 	 */
 	public function adjust_fuzzy_fields( $query, $post_type, $args ) {
-		if ( ! Utils\is_integrated_request( $this->slug, [ 'public' ] ) || empty( $args['s'] ) ) {
+		if ( ! Utils\is_integrated_request( $this->slug, [ 'public', 'ajax' ] ) || empty( $args['s'] ) ) {
```
felipeelia: @burhandodhy looking at this and the `set_fuzziness` method, I think this could be a breaking change hard to work around. Can we create a new `get_contexts` method that would return `[ 'public', 'ajax' ]` by default, but people could use a filter to change it however they want?

felipeelia (changes requested):

felipeelia (approved):

PR #3937, opened 2024-06-12
Fix remove from index on update

### LINKED ISSUE #3936: BUG: Removal from index on update may fail
### Describe the bug

Setting a password on a post and updating it from the admin page does not always remove it from the ES index. It may require hitting the update button a second time for the post to be removed. The behavior depends on details in the implementation of `wp_insert_post()` in WordPress, which varies between versions.

Explanation of why this might happen follows. Some versions of WordPress trigger the `set_post_terms` hook _before_ the post is updated in the database, while the `wp_insert_post` hook is triggered *after* the database update. You can check this in the implementation of `wp_insert_post()` of WordPress 6.4.3 for example. In other WordPress versions (6.5.2) both hooks are invoked _after_ the database update.

ElasticPress uses both of these hooks. As a result, `SyncManager::should_reindex_post()` is called twice while updating a post. Because it fetches the post from the database, this method may first see the old post, then the new post.

Currently, when any of the filters adds the post to the `SyncManager` update queue, the post is indexed even when other filters determine that the post should _not_ be indexed. I think this is not correct. Regardless of whether or not this is a WordPress bug, I think the last filter that runs should have the final say about indexing the post or not.

### Steps to Reproduce

1. Create and publish a post
2. Check that the post is in de index
3. Change visibility, set a password
4. Update the post
5. Check that the post is no longer in the index

_No response_

### Environment information

Ubuntu 22.04 , Chromium 125.0.6422.141

_No response_

### Code of Conduct

dtakken:
This change fixed an issue that I came across: Setting a password on a post and updating it from the admin page did _not_ remove it from the index, unless I hit the update button a second time.

Details about the bug can be found here: #3936

This PR makes the last running filter have the final say about adding the post to the index or not.

I found this issue on WordPress 6.4.3 using ElasticPress 4.7.2. As far as I can see the relevant code has not changed since. I did not test this change on the latest ElasticPress version yet.

I tried to whip up a Cypress test to demonstrate the issue but it does not seem to work correctly. I am not familiar with Cypress, some help is appreciated.

Closes #3936

### How to test the Change
1. Create and publish a post
2. Check that the post is in de index
3. Change visibility, set a password
4. Update the post
5. Check that the post is no longer in the index

### Changelog Entry

> Fixed - Failure to remove post from index during update for some WordPress versions

### Credits
Props @dtakken

### Checklist:


felipeelia:
Hi @dtakken, thanks for sending the PR! We'll have it reviewed and get back to you with our findings/approval. It can take a while though, so I'll ask you some patience :)

includes/classes/SyncManager.php
```diff
@@ -131,6 +131,10 @@ public function remove_from_queue( $object_id ) {
 			return false;
 		}
 
+		if ( ! isset( $this->sync_queue[ $object_id ] ) ) {
+			return false;
+		}
+
```
felipeelia, line 137: Since EP 5.0.0 the sync_queue is indexed by the blog ID, so this `if` can go away (we check the same thing a bit further down.)

tests/cypress/integration/features/protected-content-removal.cy.js
felipeelia, line 1: This test should either go to the protected-content.cy.js file or (preferably) to indexables/post.cy.js, but we'll need to make sure the test actually fails if we remove those additional `remove_from_queue` calls added to SyncManager.php.

tests/cypress/support/commands.js
```diff
@@ -202,6 +202,26 @@ Cypress.Commands.add('publishPost', (postData, viewPost) => {
 	cy.wait(2000);
 });
 
+Cypress.Commands.add('postSetPassword', (id, password) => {
```
felipeelia, line 205: Can we rename this to `setPostPassword` instead, please?

felipeelia (changes requested):
Hey @dtakken, thanks for putting this together! We'll need a couple of changes before merging this one.
Also, regarding Cypress tests, do you have any specific questions we can help you with? Were you able to set it up locally?

felipeelia:
Hi @dtakken! Any chance you can make the changes we need on this PR? Thanks in advance!

felipeelia:
Closing this in favor of #4108.

PR #4019, opened 2024-11-26
Add new filter to disable query logging: ep_disable_query_logging

### LINKED ISSUE #4006: BUG: query logging memory hog
### Describe the bug

Big bulk operations in the WP-CLI context on an environment where `WP_DEBUG` is `true` can, in some use cases, result in an unsustainable growth of this query logging:

https://github.com/10up/ElasticPress/blob/72e9dc0debc7a7f37d4164362f1024e64d7f3750/includes/classes/Elasticsearch.php#L1738-L1740

A filter should be added to the conditional for this query logging so the logging can be explicitly disabled for these bulk operations without having to change the `WP_DEBUG` constant, ex:

```diff
public function add_elasticpress_version_to_user_agent( $user_agent ) {
	 * @param array $query Query to log.
	 */
	protected function add_query_log( $query ) {
-		if ( ( defined( 'WP_DEBUG' ) && WP_DEBUG ) || ( defined( 'WP_EP_DEBUG' ) && WP_EP_DEBUG ) ) {
+		$disable_logging = apply_filters( 'ep_disable_query_logging', false );
+		if ( ! $disable_logging && ( ( defined( 'WP_DEBUG' ) && WP_DEBUG ) || ( defined( 'WP_EP_DEBUG' ) && WP_EP_DEBUG ) ) ) {
			$this->queries[] = $query;
		}
```

### Steps to Reproduce

This doesn't happen on out of the box EP installs, but can happen in certain use cases. For us at VIP it happens out of the box for how we're using EP (more on this later), to replicate:

- setup local wp site [using `vip dev-env`](https://docs.wpvip.com/vip-local-development-environment/), ensure "Enable Elasticsearch"
- set vip-config.php to

```php
define( 'WP_DEBUG', true );
define( 'VIP_ENABLE_VIP_SEARCH', true ); // Enable Enterprise Search
define( 'VIP_ENABLE_VIP_SEARCH_QUERY_INTEGRATION', true ); // Integrate search queries with Enterprise Search
```

- have some posts on the site
- create following wpcli cmd

```php
if ( defined('WP_CLI') && WP_CLI ) {
    class Test_CLI_Command {

        public function ing( $args, $assoc_args ) {
            global $wpdb;

            wp_defer_term_counting( true );
            $counter = 0;
            // loop forever to replicate a huge sites bulk operation
            while ( true ) {
                $posts = get_posts( [
                    'post_type' => 'post',
                    'posts_per_page' => -1,
                    'fields' => 'ids',
                ] );
                foreach ( $posts as $post_id ) {
                    update_post_meta( $post_id, 'test', rand() );

                    \WP_CLI::line( "MEMORY: " . round(memory_get_peak_usage() / 1024 / 1024, 2) . "MB" );
                    
                    $counter++;
                    if ($counter % 100 === 0 ) {
                        \WP_CLI::line("batch of 100 done");

                        // memory cleanup 
	                $wpdb->queries = array();
                    }
                }
            }
        }
    }
    WP_CLI::add_command( 'test', 'Test_CLI_Command' );
}
```

- run the cmd `wp test ing`
- notice the slow growth of memory usage (about +0.01mb every handful of posts)
- comment out this https://github.com/10up/ElasticPress/blob/72e9dc0debc7a7f37d4164362f1024e64d7f3750/includes/classes/Elasticsearch.php#L1739
- run cmd again, notice no slow growth of memory

This is happening specifically on VIP's use case because:

- [updated_post_meta in custom cli script triggers this](https://github.com/Automattic/vip-go-mu-plugins-built/blob/4d1d391fdcc962d295b335bf2df9ff154ff13826/search/elasticpress/includes/classes/Indexable/Post/SyncManager.php#L59)
- [EP does a kill_sync check on that callback](https://github.com/Automattic/vip-go-mu-plugins-built/blob/4d1d391fdcc962d295b335bf2df9ff154ff13826/search/elasticpress/includes/classes/Indexable/Post/SyncManager.php#L151)
- [kill_sync runs this filter](https://github.com/Automattic/vip-go-mu-plugins-built/blob/4d1d391fdcc962d295b335bf2df9ff154ff13826/search/elasticpress/includes/classes/SyncManager.php#L233) 
- [which VIP taps into here](https://github.com/Automattic/vip-go-mu-plugins/blob/87ffe8c418151b0fa041395be1b967cbcd0681d6/search/includes/classes/class-search.php#L633)
- [where VIP does an index_exists check](https://github.com/Automattic/vip-go-mu-plugins/blob/87ffe8c418151b0fa041395be1b967cbcd0681d6/search/includes/classes/class-search.php#L2221)
- [index_exists makes a remote request here](https://github.com/Automattic/vip-go-mu-plugins-built/blob/4d1d391fdcc962d295b335bf2df9ff154ff13826/search/elasticpress/includes/classes/Elasticsearch.php#L1029)
- [which then gets unsustainably logged here](https://github.com/Automattic/vip-go-mu-plugins-built/blob/4d1d391fdcc962d295b335bf2df9ff154ff13826/search/elasticpress/includes/classes/Elasticsearch.php#L1252
)

We reviewed how we've hooked into this for the index check, but we believe it would be best to have a filter for this query logging so this logging can be disabled entirely in the WP-CLI context without using the wider-use WP_DEBUG constant.

see above

### Environment information

_No response_

WP 6.6
EP 4.2.2

### Code of Conduct

- [ ] I agree to follow this project's Code of Conduct

rebeccahum:
### Description of the Change
Adds a new filter called `ep_disable_query_logging` to disable query logging due to potential memory leaks.

Closes #4006

### How to test the Change
`add_filter( 'ep_disable_query_logging', '__return_true' );`

### Credits
Props @davidsword

### Checklist:

### DIFF
--- a/includes/classes/Elasticsearch.php
+++ b/includes/classes/Elasticsearch.php
@@ -1736,7 +1736,7 @@ class Elasticsearch {
 	 */
 	protected function add_query_log( $query ) {
 		$wp_debug      = defined( 'WP_DEBUG' ) && WP_DEBUG;
-		$wp_ep_debug = defined( 'WP_EP_DEBUG' ) && WP_EP_DEBUG );
+		$wp_ep_debug   = defined( 'WP_EP_DEBUG' ) && WP_EP_DEBUG;
 		
 		/**
 		 * Filter query logging. Don't log anything to the queries property when true.
@@ -1746,7 +1746,7 @@ class Elasticsearch {
 		 * @return {bool} New value
 		 * @since  5.2.0
 		 */
-		$disable_query_logging = apply_filters( 'ep_disable_query_logging', false )
+		$disable_query_logging = apply_filters( 'ep_disable_query_logging', false );
 
 		if ( ! $disable_query_logging && ( $wp_debug || $wp_ep_debug ) ) {
 			$this->queries[] = $query;
--- a/includes/classes/Elasticsearch.php
+++ b/includes/classes/Elasticsearch.php
@@ -1730,14 +1730,14 @@ class Elasticsearch {
 	 * Query logging. Don't log anything to the queries property when
 	 * WP_DEBUG is not enabled. Calls action 'ep_add_query_log' if you
 	 * want to access the query outside of the ElasticPress plugin. This
-	 * runs regardless of debufg settings.
+	 * runs regardless of debug settings.
 	 *
 	 * @param array $query Query to log.
 	 */
 	protected function add_query_log( $query ) {
-		$wp_debug      = defined( 'WP_DEBUG' ) && WP_DEBUG;
-		$wp_ep_debug   = defined( 'WP_EP_DEBUG' ) && WP_EP_DEBUG;
-		
+		$wp_debug    = defined( 'WP_DEBUG' ) && WP_DEBUG;
+		$wp_ep_debug = defined( 'WP_EP_DEBUG' ) && WP_EP_DEBUG;
+
 		/**
 		 * Filter query logging. Don't log anything to the queries property when true.
 		 *
--- a/includes/classes/Elasticsearch.php
+++ b/includes/classes/Elasticsearch.php
@@ -1735,6 +1735,9 @@ class Elasticsearch {
 	 * @param array $query Query to log.
 	 */
 	protected function add_query_log( $query ) {
+		$wp_debug      = defined( 'WP_DEBUG' ) && WP_DEBUG;
+		$wp_ep_debug = defined( 'WP_EP_DEBUG' ) && WP_EP_DEBUG );
+		
 		/**
 		 * Filter query logging. Don't log anything to the queries property when true.
 		 *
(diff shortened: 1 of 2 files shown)


felipeelia:
Thanks @rebeccahum. Any chance you can add a docblock to that new filter following the pattern we have for the other filters?

includes/classes/Elasticsearch.php
```diff
@@ -1735,7 +1735,16 @@ public function add_elasticpress_version_to_user_agent( $user_agent ) {
...
+		/**
+		 * Filter query logging. Don't log anything to the queries property when true.
+		 *
+		 * @hook ep_disable_query_logging
+		 * @param  {bool} Whether to log to the queries property. Defaults to false.
+		 * @return {bool} New value
+		 * @since  5.2.0
+		 */
+		if ( apply_filters( 'ep_disable_query_logging', false ) && ( ( defined( 'WP_DEBUG' ) && WP_DEBUG ) ||
+		( defined( 'WP_EP_DEBUG' ) && WP_EP_DEBUG ) ) ) {
```
felipeelia:
```suggestion
		$wp_debug    = defined( 'WP_DEBUG' ) && WP_DEBUG;
		$wp_ep_debug = defined( 'WP_EP_DEBUG' ) && WP_EP_DEBUG );
		
		/**
		 * Filter query logging. Don't log anything to the queries property when true.
		 *
		 * @hook ep_disable_query_logging
		 * @param  {bool} Whether to log to the queries property. Defaults to false.
		 * @return {bool} New value
		 * @since  5.2.0
		 */
		$disable_query_logging = apply_filters( 'ep_disable_query_logging', false )

		if ( ! $disable_query_logging && ( $wp_debug || $wp_ep_debug ) ) {
```

@rebeccahum I didn't test it, but it seems we need to negate the filter value to have it working the way we expect, no? I've expanded the code a bit to make it easier to read.

Any chance you can add a unit test too?

rebeccahum: Sure, that approach makes sense to me for it to override, actually yeah.

rebeccahum:
@felipeelia Done! Thank you

PR #4029, opened 2024-12-05
Update our blocks `apiVersion` from 2 to 3

### LINKED ISSUE #3853: Adopt Block API version 3

dkotter:
### Description of the Change

This is a follow on from the work started in #3863. We update the `apiVersion` in all of our blocks from 2 to 3, to allow the iframed editor to work correctly.

This change causes some of our E2E tests to start failing, as we now need to interact with the block editor in an iframe state, something Cypress has issues with without changes. The following changes have been made:

1. Introduce a new `getBlockEditor` command that will return the block editor iframe
2. Anywhere we need to interact with the editor, use that new command
3. Introduce a new `getIframe` function that the `getBlockEditor` command uses to properly get the block editor iframe
4. Introduce a new `closeBlockInserter` command to close the block inserter. In running tests locally, was seeing some issues where after adding a block, if the block inserter wasn't closed, Cypress was complaining about elements not being ready to interact with
5. Set `chromeWebSecurity` in our Cypress config to `false` to allow Cypress to properly interact with iframes
6. Add some `wait` commands to ensure things save properly. Not sure if that will be needed in our pipelines but running tests locally, sometimes things fire too quickly and changes we're making end up not getting saved (and causes tests to fail)

Closes #3853

### How to test the Change

Main thing to test here is that all blocks still work as expected.

Also ensure all E2E tests pass.

### Changelog Entry

> Changed - Update all of our blocks `apiVersion` from 2 to 3, to indicate support for working in an iframed editor
> Developer - Update E2E tests to work properly with the iframed block editor

### Credits

Props @dkotter, @JakePT

### Checklist:

### DIFF
--- a/.github/workflows/cypress-tests.yml
+++ b/.github/workflows/cypress-tests.yml
@@ -42,7 +42,6 @@ jobs:
         key: npm-${{ env.NODE_VERSION }}-${{ hashFiles('**/package-lock.json') }}
         restore-keys: |
           npm-${{ env.NODE_VERSION }}-
-
     - name: Prepare composer cache
       uses: actions/cache@v4
       with:
@@ -131,7 +130,6 @@ jobs:
         key: npm-${{ env.NODE_VERSION }}-${{ hashFiles('**/package-lock.json') }}
         restore-keys: |
           npm-${{ env.NODE_VERSION }}-
-
     - name: Prepare composer cache
       uses: actions/cache@v4
       with:
--- a/assets/css/ordering.css
+++ b/assets/css/ordering.css
@@ -57,6 +57,10 @@
 			color: #0073aa;
 		}
 
+		& .removed .title {
+			color: #999;
+		}
+
 		& .pointer-actions {
 			float: right;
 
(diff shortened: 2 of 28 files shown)


dkotter:
@fabiankaegy @felipeelia I've updated all of the custom blocks from `apiVersion` 2 to 3 and also have updated E2E tests to work with the iframed editor. All tests are passing though note I didn't do much in the way of formal QA on the blocks so probably worth some manual testing here to ensure those all still work as expected

felipeelia:
Thank you VERY much @dkotter. I'm merging this one now, as a new EP release is coming in the next few days and blocks will be thoroughly QA'd during that.

fabiankaegy:
🥳 Thanks so much for pushing this forward!

PR #4064, opened 2025-01-23
Display autosuggest results on focus back, return to input on key up if first item.

### LINKED ISSUE #1860: Accessibility issues with Autosuggest
Autosuggest has some A11y issues that should be corrected

WCAG Item | fix needed

**1.4.11** 

- while a focus indicator is present, it is not highly visible and can be improve upon

**2.1.1**

- tab should not be used to cycle through list of suggestions (only the up and down arrow should be used to do that)
- up arrow at the first item of the list should pop the user back into the search box
[autocomplete practices](https://w3c.github.io/aria-practices/examples/combobox/combobox-autocomplete-both.html
)

**2.1.2**

- Once the "tab" is used to navigate through the list, the up and down arrow keys no longer work to navigate the list. While technically, the user is not "trapped" and can continue using tab to move focus, it is not a good user experience.

**4.1.2**

- while the search input box has a name, "s" is not very descriptive as to what it means in context to the input box

### LINKED ISSUE #3482: I need some modification in elastic-press autosuggestion search.
Hello, I am wordpress developer, I am using elastic-press search  from long time and it works fine, But I have requirement 
regarding the elastic-press search bar autosuggestion search box: - 
So, when we search into the search bar, The search bar display the autosuggestion box.
So the problem is that when we click on it the autosuggestion box get disable.
Now I need some modification in it, as per my requirement I want is that when we click on the autosuggestion box it should not disable it should hold the autosuggestion box until we click anywhere except autosuggestion box.

oscarssanchezz:
### Description of the Change

Closes https://github.com/10up/ElasticPress/issues/1860, https://github.com/10up/ElasticPress/issues/3482

### Up arrow on first Autosuggest item

1.- Type something into the autosuggest
2.- Press down to move to the first item.
3.- Press up.
4.- First item should be unselected.

https://github.com/user-attachments/assets/2e12de19-8938-4efa-900c-78e691dba68a

### Autosuggest list should appear on focus back

1.- Type something into autosuggest
2.- Move the focus away by clicking outside the search input
3.- Click back into the input
4.- Autosuggest list should be displayed.

https://github.com/user-attachments/assets/f44d85fc-87ea-4341-949d-67d468e29468

### Changelog Entry

> Added - Ability to display results on focus back + cached autosuggest results on same query.
> Fixed - Bug fix: Cannot get outside of autosuggest list pressing up on first item.

### Credits

Props @oscarssanchezz , @JakePT, @felipeelia 

### Checklist:


tests/cypress/integration/features/autosuggest.cy.js
```diff
@@ -36,6 +36,37 @@ describe('Autosuggest Feature', () => {
...
+		cy.get('.wp-block-search__button').focus();
+		cy.get('.wp-block-search__input').click();
+		cy.get('.wp-block-search__input').focus();
+
+		cy.get('.ep-autosuggest').should(($autosuggestList) => {
+			// eslint-disable-next-line no-unused-expressions
+			expect($autosuggestList).to.be.visible;
+			expect($autosuggestList[0].innerText).to.contains('Markup: HTML Tags and Formatting');
+		});
+	});
```
felipeelia: As e2e tests are very resource intensive, we don't make them as atomic as unit tests. Let's add this last part to the existent test and stick with just one (bigger) test.

oscarssanchezz: Merged into 
 single: https://github.com/10up/ElasticPress/pull/4064/commits/3ea677843328a7b864bb45f3b05bacdcc3068b0a

Issue #3654, opened 2023-09-25
BUG: Indexing parameters for object IDs only work for posts


JakePT:
### Describe the bug

When indexing using the parameters for syncing specific object IDs, such as `include`, `lower_limit_object_id`, and `upper_limit_object_id`, the parameters will only work as expected when indexing Posts. This issue is present when syncing via `IndexHelper` in PHP, WP CLI, or the REST API. 

### Steps to Reproduce

1. In WP CLI run `wp elasticpress sync --indexables=comment --include=30,32`, with real comment IDs.
2. All comments will be indexed, not just the included IDs.
3. Repeaet, with posts: `wp elasticpress sync --indexables=post --include=30,32`, with real post IDs.
4. Only the included posts will be synced.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

PR #4102, opened 2025-03-06
Fix: Term sync issue during the pagination

burhandodhy:
### Description of the Change

This PR fixes the issue where an incorrect number of terms were synced when pagination was used. The problem started occurring after merging this change https://github.com/10up/ElasticPress/issues/3654.

Closes #

### How to test the Change

Run `wp elasticpress sync --setup --show-errors --yes  --indexables=term --per-page=10` command and confirm if all the terms are synced.

### Changelog Entry

> Fixed - Term sync issue
 

### Credits

Props @burhandodhy 

### Checklist:

### DIFF
--- a/includes/classes/Indexable/Term/Term.php
+++ b/includes/classes/Indexable/Term/Term.php
@@ -248,24 +248,8 @@ class Term extends Indexable {
 			);
 			add_filter( 'terms_clauses', array( $this, 'bulk_indexing_filter_terms_where' ), 9999, 3 );
 
-			/**
-			 * Filter database arguments for term count query
-			 *
-			 * @hook ep_term_all_query_db_args
-			 * @param  {array} $args Query arguments based to `wp_count_terms()`
-			 * @since  3.4
-			 * @return {array} New arguments
-			 */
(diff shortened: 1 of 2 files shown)


includes/classes/Indexable/Term/Term.php
```diff
@@ -251,24 +251,8 @@ public function query_db( $args ) {
...
 
-			/**
-			 * Filter database arguments for term count query
-			 *
-			 * @hook ep_term_all_query_db_args
-			 * @param  {array} $args Query arguments based to `wp_count_terms()`
-			 * @since  3.4
-			 * @return {array} New arguments
-			 */
-			$total_objects = wp_count_terms( apply_filters( 'ep_term_all_query_db_args', $all_query_args, $args ) );
```
felipeelia, line 262: Shouldn't we add this filter somewhere else now?

burhandodhy: The filter is still used when advanced_pagination is not set https://github.com/10up/ElasticPress/blob/adf3668b3340041a7c9f7c6de071bbd2ff09c0cd/includes/classes/Indexable/Term/Term.php#L268

Issue #4123, opened 2025-04-16
Enhancements on the Features page


felipeelia:
### Is your enhancement related to a problem? Please describe.

Although we've improved the Features page A LOT in EP 5, it is time to enhance it further. Here is a list of things we should implement:

**Features**

* Grouping related features: WooCommerce-related features, AI-related features, etc.
* Keep the current feature on reload: Currently, if you reload, you always go to the first feature. Reloading and staying in the current feature would be ideal

**Features Fields**

* Field dependency: Conditionally hide and display fields depending on other field(s)' values. If a certain radio button is checked, toggle a group of fields that are only used if the radio button has that value, for example.
* Field grouping: We need a way to indicate that a group of fields belongs together. It could be just a common indentation or a real grouping indicator.

---

Ideally, #3745 would also be addressed in this set of changes.

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

PR #4127, opened 2025-04-23
Feature Grouping (GH4123 Req 1)

ZacharyRener:
### Description of the Change

> Grouping related features: WooCommerce-related features, AI-related features, etc.

This adds the ability to group the different features. It includes support for adding a new 'group' attribute member to each Feature child class, and interface updates that group the different features into vertical tabs. Some of the file changes here are just populating the child class files with a group name.

#4123 Request 1 (Feature Grouping)

### How to test the Change

1. On a Feature child-class, set the group name in the constructor: 

```
[DidYouMean.php]
public function __construct() {
    ...
    $this->group = 'Core Search';
    ...
}
```
2. View the ElasticPress features page. The new group should be present, with the respective feature(s) inside of it:

enter image description here

3. Confirm that non-grouped features are visible in the 'Other' group. For example, all of the features provided from ElasticPressLabs:

enter image description here

### Changelog Entry

> Added - New feature: Ability to group together similar features
### Credits

Props @ZacharyRener

### Checklist:

### DIFF
--- a/includes/classes/Feature/AcfRepeater/AcfRepeater.php
+++ b/includes/classes/Feature/AcfRepeater/AcfRepeater.php
@@ -37,7 +37,7 @@ class AcfRepeater extends Feature {
 	 * Initialize feature setting it's config
 	 */
 	public function __construct() {
-		$this->slug  = 'acf_repeater';
+		$this->slug = 'acf_repeater';
 		parent::__construct();
 	}
 
--- a/includes/classes/Feature/AcfRepeater/AcfRepeater.php
+++ b/includes/classes/Feature/AcfRepeater/AcfRepeater.php
@@ -38,7 +38,6 @@ class AcfRepeater extends Feature {
 	 */
 	public function __construct() {
 		$this->slug  = 'acf_repeater';
-		$this->group = '3rd Party Plugins';
 		parent::__construct();
 	}
 
@@ -47,6 +46,8 @@ class AcfRepeater extends Feature {
 	 */
 	public function set_i18n_strings(): void {
 		$this->title = esc_html__( 'ACF Repeater Field Compatibility', 'elasticpress' );
+
+		$this->group = esc_html__( '3rd Party Plugins', 'elasticpress' );
 
 		$this->short_title = esc_html__( 'ACF Repeater Field', 'elasticpress' );
 
--- a/includes/classes/Feature/AcfRepeater/AcfRepeater.php
+++ b/includes/classes/Feature/AcfRepeater/AcfRepeater.php
@@ -37,8 +37,8 @@ class AcfRepeater extends Feature {
 	 * Initialize feature setting it's config
 	 */
 	public function __construct() {
-		$this->slug = 'acf_repeater';
-
+		$this->slug  = 'acf_repeater';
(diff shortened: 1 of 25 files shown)


includes/classes/Feature/AcfRepeater/AcfRepeater.php
```diff
@@ -37,8 +37,8 @@ class AcfRepeater extends Feature {
 	 * Initialize feature setting it's config
 	 */
 	public function __construct() {
-		$this->slug = 'acf_repeater';
-
+		$this->slug  = 'acf_repeater';
+		$this->group = '3rd Party Plugins';
```
felipeelia: As these strings will need to be translated, we'll have to move them to that set_i18n_strings method, @ZacharyRener.

tests/cypress/integration/features/groups.cy.js
```diff
@@ -0,0 +1,74 @@
...
+		 * and contains at least one feature tab.
+		 */
+		cy.get('@tabsContainer')
+			.should('exist')
+			.and('be.visible')
+			.find('button')
+			.should('have.length.at.least', 1);
+	});
+
+	it('switches content when clicking a tab', () => {
```
felipeelia: For e2e tests, we don't want to be too atomic, as they are super expensive to run. Is there any chance we can test everything in a single test?

felipeelia (changes requested):
This looks great, @ZacharyRener! We'll need a couple of changes before merging it, and it also seems e2e tests are failing on WP 6.2. Do you mind giving it a look? Thanks!

includes/classes/Feature/AcfRepeater/AcfRepeater.php
ZacharyRener: Good catch - Moved all of them

tests/cypress/integration/features/groups.cy.js
ZacharyRener: For sure - Just condensed this into one single test

ZacharyRener:
Waiting on tests to finish running here, but things are looking good locally

ZacharyRener:
> This looks great, @ZacharyRener! We'll need a couple of changes before merging it, and it also seems e2e tests are failing on WP 6.2. Do you mind giving it a look? Thanks!

Hey @felipeelia These changes have been added in. 

On the e2e tests: Every feature's test spec involves clicking on that feature's tab (e.g. Autosuggest). Now, with groups, each test needs to click the group first (e.g. Live Search), to make sure the feature itself is rendered and clickable. I updated them to account for that. Just one or two lines per test typically. But that's what was causing a few e2e tests to fail.

felipeelia (changes requested):
This is looking good now, @ZacharyRener. The last things we need to update are:
* `@since` tags need to be 5.3.0. EP 5.2.0 was already released (sorry if I gave you the wrong version at some point);
* Linter is complaining about some extra spaces in strings like `$this->slug  = 'acf_repeater';`. Do you mind fixing these as well?

Thanks!

ZacharyRener:
@felipeelia Done - updated since tag, and removed those extra spaces.

PR #4138, opened 2025-04-29
Feature Selection Persistence (Issue 4123 Req 2)

ZacharyRener:
### Description of the Change

> Keep the current feature on reload: Currently, if you reload, you always go to the first feature. Reloading and staying in the current feature would be ideal

This makes it to where the current group - and current feature within that group - that are selected will persist across page reloads. It accomplishes this using the @wordpress/preferences-persistence package.

#4123 Request 2 (Feature Persistence)

### How to test the Change

1. Select a group, and select a feature within that group.
2. Reload the page
3. Confirm that the group and feature are both preselected
4. Confirm the URL is changing, and that the forward and backward browser buttons work
5. Confirm pasting the URL in a new tab navigates correctly

enter image description here

### Changelog Entry

> Added - New feature: Selected features and groups now persist across reload

### Credits

Props @ZacharyRener

### Checklist:


assets/js/features/apps/features.js
```diff
@@ -19,6 +22,19 @@ import Tab from '../components/tab';
  */
 import '../style.css';
 
+/**
+ * Keeps persistence of the last active feature setting across page reloads.
+ */
+wp.data.dispatch('core/preferences').setPersistenceLayer(createPersistence());
```
burhandodhy: Can't we use `useDispatch` hook?

ZacharyRener: Removed this while going from persistence preferences -> url parameters

assets/js/features/apps/features.js
```diff
@@ -1,7 +1,10 @@
+// eslint-disable-next-line import/no-extraneous-dependencies
```
burhandodhy, line 4: Please add a comment similar to the ones we have in other files for external dependencies. https://github.com/10up/ElasticPress/blob/feature/issue-4123-req-2/assets/js/instant-results/components/common/modal.js#L1C1-L3C4

assets/js/features/apps/features.js
```diff
@@ -12,19 +15,98 @@ import { useSettingsScreen } from '../../settings-screen';
...
+		>
+			{title}
+		</button>
+	);
+};
+
+const GroupNavigation = ({ groupedFeatures, groupSlug }) => (
+	<div className="ep-dashboard-outer-tabs">
+		<div className="ep-dashboard-tabs-nav">
+			{groupedFeatures.map((groupObj) => (
```
burhandodhy: Nit: Destructure the groupObj

```
{groupedFeatures.map(({ groupSlug: slug, title, features }) => (
```

assets/js/features/apps/features.js
```diff
@@ -12,19 +15,98 @@ import { useSettingsScreen } from '../../settings-screen';
...
+		<Panel className="ep-dashboard-panel">
+			<PanelBody>
+				{isSyncing ? (
+					<Notice actions={isSyncingActions} isDismissible={false} status="warning">
+						{isSyncingNotice}
+					</Notice>
+				) : null}
+				<div className="ep-dashboard-tabs">
+					<div className="ep-dashboard-tabs-nav">
+						{currentGroup.features.map((featureObj) => (
```
burhandodhy: Nit: Destructure the featureObj

package.json
```diff
@@ -9,6 +9,7 @@
     "@wordpress/icons": "^10.13.0",
     "chart.js": "^4.4.7",
     "focus-trap-react": "^11.0.3",
+	"react-router-dom": "^6.14.3",
```
burhandodhy: wrong indentation

tests/cypress/integration/features/choice-persists-on-reload.cy.js
```diff
@@ -0,0 +1,42 @@
+describe('Feature Selection Persistence', () => {
```
burhandodhy: The featured directory is specifically dedicated to ElasticPress features, so I don't think it makes sense to add the file there. Since the tests are related to the settings page, it might be better to create a new file for settings and include both this and the tests from groups.cy.js in that file.

includes/classes/Feature/SearchOrdering/SearchOrdering.php
```diff
@@ -52,6 +52,8 @@ class SearchOrdering extends Feature {
 	public function __construct() {
 		$this->slug = 'searchordering';
 
+		$this->group_slug = 'core-search';
```
burhandodhy: Instead of introducing two new properties, can't we combine the label and slug into one?

```
$this->group = [
			'label' => esc_html__( 'Indexing Options', 'elasticpress' ),
			'slug' => 'indexing-options',
		];

```

assets/js/features/style.css
```diff
@@ -20,124 +20,152 @@
...
-		padding-bottom: 32px;
-	}
 }
-
+  
+/* Media Queries */
 @media (max-width: 782px) {
-	
+
+	.group-content {
```
burhandodhy: We already have `.group-content` defined [above](https://github.com/10up/ElasticPress/blob/f9c20305e67c268d056eb164ca306e7bc37de6bc/assets/js/features/style.css#L58). Can't we add the rules there?

```
.group-content {

	@media (max-width: 782px) {
		
	 }
}
```

assets/js/features/style.css
```diff
@@ -20,124 +20,152 @@
...
+  
+/* Feature Tab Styling */
 .ep-feature-tab {
 	align-items: start;
 	display: flex;
 	flex-direction: column;
 	gap: 4px;
 }
-
+
```
burhandodhy: Make sure your editor is not adding trailing spaces.

assets/js/features/apps/features.js
ZacharyRener: Added

assets/js/features/apps/features.js
ZacharyRener: Done

assets/js/features/apps/features.js
ZacharyRener: Done

package.json
ZacharyRener: Fixed

tests/cypress/integration/features/choice-persists-on-reload.cy.js
ZacharyRener: Moved this test, and one related to the grouping setting, to a new setting folder as suggested

includes/classes/Feature/SearchOrdering/SearchOrdering.php
ZacharyRener: I like the idea - done

assets/js/features/style.css
ZacharyRener: Adjusted

assets/js/features/style.css
ZacharyRener: Fixed

felipeelia (changes requested):
@ZacharyRener giving this a second/third look, I think we still need some changes (sorry!):

1. Cypress tests shouldn't be in a folder called "settings" as that is the name of a different screen. Let's add the content of these two files in a single one (potentially in a single test even) and have it in the features folder.
2. I wonder if we shouldn't centralize the groups creation somewhere (ElasticPress\Features perhaps?). Labels would be registered that (with a filter so people could create their own groups) and then features would only need slugs. Having the same string being i18n in several places is a bit weird. Thoughts?

PR #4141, opened 2025-05-01
Feature Field Dependency (Issue 4123 Req 3)

ZacharyRener:
### Description of the Change

> Field dependency: Conditionally hide and display fields depending on other field(s)' values. If a certain radio button is checked, toggle a group of fields that are only used if the radio button has that value, for example.

#4123 Request 3

### How to test the Change

Add `requires_fields` to a feature field's settings_schema.
Format:
```
'requires_fields' => [
	'relationship' => 'OR',
	'conditions' => [
		'highlight_enabled' => '1',
		'highlight_excerpt' => '1',
	]
]
```

enter image description here

### Changelog Entry

> Added - Feature fields can be dependent on other fields

### Credits

Props @ZacharyRener

### Checklist:


felipeelia:
I know this is probably not in the initial scope, but can we make it dependent on two or more fields? (This will likely require a "relation" attribute too...) The scenario I have in mind right now is the "Highlight tag" in Post Search. It should only be displayed if "Highlight search terms" OR "Highlight search terms in excerpts" are enabled. (We might have another scenario where we need an AND relationship, so I'd like to be prepared for that if we can.)

ZacharyRener:
@felipeelia I updated the format for required_fields, it's on the PR description. It supports two or more fields, and relationship (AND, OR)

PR #4142, opened 2025-05-01
Feature Field Grouping (Issue 4123 Req 4)

ZacharyRener:
### Description of the Change

> Field grouping: We need a way to indicate that a group of fields belongs together. It could be just a common indentation or a real grouping indicator.

#4123 Request 4

### How to test the Change

In a Feature's set_settings_schema function, add an entry with the type 'field_group'. Set the 'key', 'label', and 'fields' properties inside of it. Example for Autosuggest.php:
```
$this->settings_schema = [
	[
		'type'   => 'field_group',
		'key'    => 'fieldgroupa',
		'label'  => 'Field Group ABC',
		'fields' => [
			[
				'default' => '.ep-autosuggest',
				'help'    => __( 'Input additional selectors where you would like to include autosuggest, separated by a comma. Example: <code>.custom-selector, #custom-id, input[type="text"]</code>', 'elasticpress' ),
				'key'     => 'autosuggest_selector',
				'label'   => __( 'Additional selectors', 'elasticpress' ),
				'type'    => 'text',
			],
			[
				'default' => '0',
				'key'     => 'trigger_ga_event',
				'help'    => __( 'Enable to fire a gtag tracking event when an autosuggest result is clicked.', 'elasticpress' ),
				'label'   => __( 'Trigger Google Analytics events', 'elasticpress' ),
				'type'    => 'checkbox',
			],
		],
	],
];
```
Result:

### Considerations
- This is set to work with requires_fields from #4141 
- Applies to both the fields within a group, and the group itself. Both can have the requires_fields property set

### Changelog Entry

> Added - Ability to create groups of fields

### Credits

Props @ZacharyRener

### Checklist:

- [x] I agree to follow this project's [**Code of Conduct**](https://github.com/10up/.github/blob/trunk/CODE_OF_CONDUCT.md).
- [x] I have updated the documentation accordingly.
- [x] I have added [Critical Flows, Test Cases, and/or End-to-End Tests](https://10up.github.io/Open-Source-Best-Practices/testing/) to cover my change.
- [x] All new and existing tests pass.

PR #4143, opened 2025-05-01
Multiple Required Features (Issue 4123 Req 5) 

ZacharyRener:
### Description of the Change

There is already support for a singular required feature:
`$this->requires_feature = 'search';`

This change adds support for multiple:
`$this->requires_features = ['search', 'autosuggest'];`

While retaining support for a singular required feature.

### How to test the Change

Add `$this->requires_features = ['feature_slug_1', 'feature_slug_2'];` to a Feature's constructor

### Changelog Entry

> Added - Ability for a feature to require multiple other features, instead of just one

### Credits

Props @ZacharyRener

### Checklist:


.wp-env.json
```diff
@@ -41,6 +41,7 @@
         "wp-content/plugins/sync-error.php": "./tests/cypress/wordpress-files/test-plugins/sync-error.php",
         "wp-content/plugins/unsupported-server-software.php": "./tests/cypress/wordpress-files/test-plugins/unsupported-server-software.php",
         "wp-content/plugins/unsupported-elasticsearch-version.php": "./tests/cypress/wordpress-files/test-plugins/unsupported-elasticsearch-version.php",
+		"wp-content/plugins/multiple-required-features.php": "./tests/cypress/wordpress-files/test-plugins/multiple-required-features.php",
```
felipeelia:
```suggestion
        "wp-content/plugins/multiple-required-features.php": "./tests/cypress/wordpress-files/test-plugins/multiple-required-features.php",
```

assets/js/features/components/control.js
```diff
@@ -136,11 +144,17 @@ export default ({
...
+			? `${titles.slice(0, -1).join(', ')}, and ${titles[titles.length - 1]}`
+			: titles[0] || '';
+
 	return (
 		<>
-			{requiredFeature ? (
+			{missingRequiredFeatures.length > 0 ? (
 				<Notice isDismissible={false} status={name === 'active' ? 'error' : 'warning'}>
-					{sprintf(requiredFeatureNotice, requiredFeature.shortTitle)}
+					{sprintf(requiredFeatureNotice, list)}
```
felipeelia, line 184: The change in `requiredFeatureNotice` is bigger than this. First, the `list` const text needs to be translatable as well (that `, and ` there), then `requiredFeatureNotice` needs to account for the number of required features, likely using `_n`.

assets/js/features/components/settings.js
```diff
@@ -92,6 +92,7 @@ export default ({ feature, settingsSchema }) => {
 			label,
 			options,
 			requires_feature,
+			requires_features,
```
felipeelia: Do we need to keep requires_feature here?

includes/classes/Feature.php
```diff
@@ -640,4 +649,25 @@ protected function set_settings_schema() {
...
+	 * @since 5.4.0
+	 * @return array List of required feature slugs
+	 */
+	public function get_required_features() {
+		$required_features = [];
+
+		// Support legacy single feature requirement
+		if ( $this->requires_feature ) {
+			$required_features[] = $this->requires_feature;
+		}
```
felipeelia: Let's add a deprecation notice here (_doing_it_wrong is probably the best fit here), so people know they need to update their code.

felipeelia (changes requested):
@ZacharyRener in addition to my comments, can you please resolve the git conflicts that appeared after merging #4142 ?

includes/classes/Feature.php
ZacharyRener: See comment below that removes the need for a deprecation notice

assets/js/features/components/settings.js
ZacharyRener: No, removed

assets/js/features/components/control.js
ZacharyRener: Done

ZacharyRener:
I changed this to where instead of having two properties:

- `required_feature`
- `required_features`

We just have one:

- `required_feature`

And this PR adds support for it to be either a String, or Array. Using an array is optional, so nothing is deprecated this way, and everything is backwards compatible

This seems to be the most graceful solution when considering tests and deprecation notices

includes/classes/Feature.php
```diff
@@ -640,4 +641,26 @@ protected function set_settings_schema() {
...
+			$required_feature = array_merge( $required_feature, $this->requires_feature );
+			return array_unique( $required_feature );
+		}
+
+		if ( $this->requires_feature ) {
+			return [ $this->requires_feature ];
+		}
+
+		return [];
+	}
```
felipeelia: Can't we go with a simpler solution like:

```suggestion
	public function get_required_feature() {
		return array_unique( (array) $this->requires_feature );
	}
```

assets/js/features/components/control.js
```diff
@@ -141,11 +159,20 @@ const Control = ({
...
 	};
 
+	const list =
+		titles.length > 1
+			? sprintf(
+					__('%1$s, and %2$s', 'elasticpress'),
+					titles.slice(0, -1).join(__(', ', 'elasticpress')),
+					titles[titles.length - 1],
+				)
+			: titles[0] || '';
```
felipeelia: Although this is good enough for now, we'll have to keep in mind that the Oxford comma will be wrong if we only have 2 items in the list.

includes/classes/Feature.php
```diff
@@ -127,12 +127,12 @@ abstract class Feature {
...
 	/**
-	 * The slug of a feature that is required to be active.
+	 * The slug, or array of slugs, of a feature that is required to be active.
 	 *
 	 * @since 5.0.0
-	 * @var false|string
+	 * @var false|string|array
 	 */
-	protected $requires_feature = false;
+	public $requires_feature = false;
```
felipeelia: If the only reason why we are changing visibility here is the plugin test we are creating, let's keep it protected and use the [reflection class](https://github.com/10up/ElasticPress/blob/2be43678323534e3cf1ed78d2d7cfc84be100001/tests/php/TestFeature.php#L126) in the test plugin to access it. That way, it'll be harder for coders to access something they really shouldn't.

felipeelia (changes requested):
@ZacharyRener in addition to my comments, there are some ES Lint things that need to be addressed.

includes/classes/Feature.php
ZacharyRener: Added this in with a tiny adjustment

assets/js/features/components/control.js
ZacharyRener: Adjusted this

includes/classes/Feature.php
ZacharyRener: Done

felipeelia:
@ZacharyRener can you please fix this git conflict so we can merge this one? Thanks!

PR #4169, opened 2025-06-24
Restrict showing others' private posts in admin panel

### LINKED ISSUE #1800: The condition for building the post_status parameter in a simple query

burhandodhy:
### Description of the Change

This PR fixes an issue where private posts from other authors are listed in the admin panel, even when the user doesn’t have permission.

Closes #1800

### Changelog Entry

> Fixed - Users can see other authors' private posts
 

### Credits

Props @burhandodhy 

### Checklist:


felipeelia:
It seems we have a problem with tests here, @burhandodhy

PR #4171, opened 2025-06-30
BUG FIX: The Field Grouping change should retain old fields values

### LINKED ISSUE #4170: BUG: The Field Grouping change should retain old fields values
### Describe the bug

The field grouping introduced in #4142 has some problems we need to fix ASAP:

- Old values are not migrated to the group
- The field value is stored in a nested way, which would break retrocompatibility

### Steps to Reproduce

Apply the field_group schema to fields in the Vector Embeddings feature in https://github.com/10up/ElasticPressLabs/pull/126

_No response_

### Environment information

_No response_

_No response_

### Code of Conduct

ZacharyRener:
### Description of the Change

Previously, I had set up a new settings schema type: field_group. It was rendering a whole new Control, and nesting the stored values. This would lead to the old values not being retained once moved into a field_group, and didn't have great retrocompatibility.

To address this, I've changed how field_groups work. Now, there is no field_group type. Instead, fields stay flat, but now have a `field_group_slug` and `field_group_label` key/value pair. That keeps things working with retrocompatibility, and retains the old values. Its now the same exact fields: the frontend just renders them in a group visually. This also helped reduced some repetitive logic surrounding conditional fields.

Closes #4170 

### How to test the Change

```
public function set_i18n_strings(): void {
        ...
	$this->field_group_map = [
		'highlight_group' => __( 'Highlight Group', 'elasticpress' ),
	];
        ...
}
```

```
protected function set_settings_schema() {
...
$this->settings_schema = [
	[
		...
		'key'               => 'highlight_enabled',
		'field_group_slug'  => 'highlight_group',
		'field_group_label' => $this->field_group_map['highlight_group'],
	],
	[
			...
		'key'               => 'highlight_excerpt',
		'field_group_slug'  => 'highlight_group',
		'field_group_label' => $this->field_group_map['highlight_group'],
	],
...
```

### Changelog Entry

> Fixed - Field groups weren't retaining old values

### Credits

Props @ZacharyRener 

### Checklist:

### DIFF
--- a/includes/classes/Feature.php
+++ b/includes/classes/Feature.php
@@ -560,6 +560,7 @@ abstract class Feature {
 			'reqStatusMessages' => (array) $requirements_status->message,
 			'settingsSchema'    => $this->get_settings_schema(),
 			'group'             => $this->group,
+			'fieldGroups'       => $this->get_field_group_map(),
 		];
 
 		return $feature_desc;
@@ -639,5 +640,28 @@ abstract class Feature {
 	 * @since 5.2.0
 	 */
 	public function set_i18n_strings(): void {
+	}
+
+	/**
+	 * Get the field group map for the feature.
+	 *
+	 * @since 5.3.0
+	 * @return array
+	 */
+	public function get_field_group_map(): array {
+		$field_groups = [
+			'highlight_group' => [
+				'label' => esc_html__( 'Highlighting Options', 'elasticpress' ),
+			],
+		];
+		/**
+		 * Filter available field groups.
+		 *
+		 * @hook ep_feature_field_groups
+		 * @since 5.3.0
+		 * @param  {array} $field_groups Current field groups
+		 * @return {array} New field groups
+		 */
+		return apply_filters( 'ep_feature_field_groups', $field_groups );
 	}
 }
--- a/includes/classes/Feature.php
+++ b/includes/classes/Feature.php
@@ -151,6 +151,14 @@ abstract class Feature {
 	public $group = false;
 
 	/**
+	 * Field groups available to a feature
+	 *
+	 * @since 5.3.0
+	 * @var array
+	 */
+	protected $field_group_map = [];
+
+	/**
 	 * Run on every page load for feature to set itself up
 	 *
 	 * @since  2.1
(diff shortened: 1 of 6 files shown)


includes/classes/Feature/Search/Search.php
```diff
@@ -894,61 +898,60 @@ protected function set_settings_schema() {
...
-				'key'    => 'highlighting',
-				'label'  => __( 'Highlighting Options', 'elasticpress' ),
-				'fields' => [
+				'default'           => '0',
+				'help'              => __( 'Enable to wrap search terms in HTML tags in results for custom styling. The wrapping HTML tag comes with the <code>ep-highlight</code> class for easy styling.', 'elasticpress' ),
+				'key'               => 'highlight_enabled',
+				'label'             => __( 'Highlight search terms', 'elasticpress' ),
+				'type'              => 'checkbox',
+				'field_group_slug'  => 'highlight_group',
+				'field_group_label' => $this->field_group_map['highlight_group'],
```
felipeelia: @ZacharyRener instead of sending the label on all fields, can't we just send the group slug and figure out the label sending the map to JS?

felipeelia: Also, it seems the group label is now "Highlight Group" instead of "Highlighting Options"? (I prefer "Highlighting Options" :) )

ZacharyRener: Sure, moved the map to JS

ZacharyRener: Moved to Feature abstract class + changed back to Highlighting Options

includes/classes/Feature.php
```diff
@@ -640,4 +641,27 @@ protected function set_settings_schema() {
...
+	 *
+	 * @since 5.3.0
+	 * @return array
+	 */
+	public function get_field_group_map(): array {
+		$field_groups = [
+			'highlight_group' => [
+				'label' => esc_html__( 'Highlighting Options', 'elasticpress' ),
+			],
+		];
```
felipeelia: @ZacharyRener This does not belong to the abstract class; it is something super specific to the Search Feature class. You can either override the method there or use the new ep_feature_field_groups field.

felipeelia (changes requested):
@ZacharyRener in addition to the comment I left, it seems Unit tests are not passing as well.

includes/classes/Feature.php
ZacharyRener: Adjusted

PR #4182, opened 2025-08-07
Store Aggregation on Query Level. 

### LINKED ISSUE #3352: Turn certain filters into WP_Query arguments

In certain parts of the plugin, custom filters are used to apply changes or functionality to `WP_Query` queries. The downside of this approach is that it's hard to target queries specifically as any query that occurs after adding to the filter will receive the changes. 

I ran into this issue developing the facet block, which uses the `ep_is_facetable` filter to enable faceting for the queries used to create the block previews. This filter applies aggregations to the query and populates a global variable with the results of those aggregations. I was experiencing an issue where another query was somehow being fired down the chain, which meant that aggregations were also applied to that query and the global variable was populated with aggregations for the wrong query.

A better approach would be to support an `ep_is_facetable` argument in `WP_Query` which applies the aggregations to just that query. Then, instead of populating a global variable, the aggregation results could be applied to the query object itself. This keeps everything contained to the specific query that needs to be facetable.

With a quick look I've identified these filters that could use this approach:

- `ep_is_facetable`
- `ep_intercept_remote_request`
- `ep_bypass_exclusion_from_search`

There may be others.

### Designs

_No response_

_No response_

### Code of Conduct

burhandodhy:
### Description of the Change

This PR updates the internal mechanism to store the aggregation data at the query level instead of using a global variable, and also deprecates the ep_is_facetable filter.

Closes #3352

### Changelog Entry

> Changed -  Stores the aggregation data at the query level

### Credits

Props @burhandodhy @JakePT 

### Checklist:


includes/classes/Feature/Facets/Facets.php
```diff
@@ -390,19 +399,102 @@ public function get_aggs( $response, $query, $query_args, $query_object ) {
...
+	 */
+	public function set_query_aggregations( $query, $aggregations ): void {
+		// Store aggregations on the query object
+		$query->ep_aggregations = $aggregations;
+
+		if ( $query->is_main_query() && $this->should_sync_to_global( $query ) ) {
+			_doing_it_wrong(
+				__METHOD__,
+				esc_html__( 'The global variable $GLOBALS[\'ep_facet_aggs\'] is deprecated. Access aggregation data directly from the query object using $query->ep_aggregations or the Facets feature methods get_query_aggregations() and get_facet_aggregation().', 'elasticpress' ),
+				'5.3.0'
```
felipeelia:
```suggestion
				'ElasticPress 5.3.0'
```
Prepending ElasticPress to the version argument will make it clearer what exactly is doing something wrong.

includes/classes/Feature/Facets/Facets.php
```diff
@@ -390,19 +399,102 @@ public function get_aggs( $response, $query, $query_args, $query_object ) {
...
 		}
+
+		$query_aggs = $query->ep_aggregations ?? false;
+		if ( false !== $query_aggs ) {
+			return $query_aggs;
+		}
+
+		// Fallback to global variable
+		return $GLOBALS['ep_facet_aggs'] ?? false;
+	}
```
felipeelia:
```suggestion
	public function get_query_aggregations( $query ) {
		if ( $query instanceof \WP_Query && isset( $query->ep_aggregations ) && false !== $query->ep_aggregations ) {
			return $query->ep_aggregations;
		}

		// Fallback to global variable
		return $GLOBALS['ep_facet_aggs'] ?? false;
	}
```

felipeelia (changes requested):
Overall, this looks good @burhandodhy, with some minor adjustments we need to make before merging. I imagine this will have a phase 2 where we make it so it is easier for people to use our Filter widgets on queries that are not the main query, right?

burhandodhy:
> Overall, this looks good @burhandodhy, with some minor adjustments we need to make before merging. I imagine this will have a phase 2 where we make it so it is easier for people to use our Filter widgets on queries that are not the main query, right?

Correct. There will be a phase 2 to run the filters on non main queries

includes/classes/Feature/Facets/Facets.php
```diff
@@ -271,14 +271,25 @@ public function front_scripts() {
...
 		 * @hook ep_is_facetable
 		 * @param  {bool}     $bypass Defaults to false.
 		 * @param  {WP_Query} $query  The current WP_Query.
 		 * @return {bool}     true to bypass, false to ignore
 		 */
-		if ( \apply_filters( 'ep_is_facetable', false, $query ) ) {
+		if ( apply_filters_deprecated(
+			'ep_is_facetable',
+			[ false, $query ],
+			'5.3.0',
```
felipeelia: It seems you missed this one, @burhandodhy

felipeelia (changes requested):
Almost there, @burhandodhy. There is 1 line of code we need to adjust, and also, it seems an e2e test is not passing. Do you mind looking into that?

PR #4185, opened 2025-08-13
Introduce new ep_intercept_request argument support

### LINKED ISSUE #3352: Turn certain filters into WP_Query arguments

burhandodhy:
### Description of the Change

This PR adds a new WP_Query argument `ep_intercept_request`.

Closes #3352

### Changelog Entry

> Added - Add support for new WP_Query argument `ep_intercept_request`

### Credits

Props @burhandodhy 

### Checklist:


includes/classes/Feature/Autosuggest/Autosuggest.php
```diff
@@ -851,6 +849,12 @@ public function epio_autosuggest_set_and_get() {
 	 * @return true
 	 */
 	public function intercept_remote_request() {
+		_doing_it_wrong(
+			__METHOD__,
+			esc_html__( 'This method should not be called anymore, use WP_Query argument `ep_intercept_request` instead.' ),
+			'5.3.0'
```
felipeelia:
```suggestion
			esc_html__( 'Use the WP_Query argument `ep_intercept_request` instead.', 'elasticpress' ),
			'ElasticPress 5.3.0'
```

includes/classes/Feature/Autosuggest/Autosuggest.php
```diff
@@ -851,6 +849,12 @@ public function epio_autosuggest_set_and_get() {
 	 * @return true
 	 */
 	public function intercept_remote_request() {
```
felipeelia, line 851: It seems we are not adding a unit test for this one, are we?

felipeelia (changes requested):

includes/classes/Feature/Autosuggest/Autosuggest.php
burhandodhy: Done. I have added the test

felipeelia (approved):

PR #4188, opened 2025-08-25
Add support for an array value in the orderby_meta_mapping filter

### LINKED ISSUE #4176: Cannot override 'orderby_meta_mapping' with an array of sorting params
### Describe your question

Hi

In WooCommerce I have an array of 'orderby' params for a specific query, e.g:

`$args = array(
'meta.key1' => 'DESC',
'meta.key2' => 'ASC',
'meta.key3' => 'DESC',
);`

This works fine, but when the sort by select form is used on the product category pages the 'orderby' param is overridden via the Elasticpress hook 'orderby_meta_mapping' I can specify a string value for this hook but not an array, and there doesn't seem a way to deactivate this hook.

As a temportary solution I've commented out the following line:

`$this->maybe_set_orderby( $query );`

In elasticpress/includes/classes/Feature/WooCommerce/Products.php line 654

Is there a way to deactivate this hook?

Thanks

### Code of Conduct

burhandodhy:
### Description of the Change

This PR adds support to set the array value in the `orderby_meta_mapping` filters

Closes #4176 

### Changelog Entry

> Changed - Adds support to set the array value in the `orderby_meta_mapping` filters
 
### Credits

Props @burhandodhy @jzzaj

### Checklist:


includes/classes/Feature/WooCommerce/Products.php
```diff
@@ -1054,14 +1054,15 @@ public function get_orderby_meta_mapping( $meta_key ): string {
...
+				'date'               => 'date',
 			)
 		);
 
 		if ( isset( $mapping[ $meta_key ] ) ) {
 			return $mapping[ $meta_key ];
 		}
 
-		return 'date';
+		return $this->get_orderby_meta_mapping( 'menu_order' );
```
felipeelia: Do we really need the recursion here?

felipeelia (changes requested):

includes/classes/Feature/WooCommerce/Products.php
burhandodhy: Yes, the idea is that if an invalid orderby is passed to this method, it falls back to `menu_order`. I can replace this line with `menu_order title date` as well.

PR #4248, opened 2025-12-05
Fix typos

szepeviktor:
### Description of the Change

Correct misspellings.

@felipeelia For you: https://github.com/crate-ci/typos

### DIFF
--- a/includes/classes/ElasticPressIo.php
+++ b/includes/classes/ElasticPressIo.php
@@ -106,7 +106,7 @@ class ElasticPressIo {
 
 		$endpoint_status = $this->get_endpoint_status( $skip_cache );
 
-		return $endpoint_status['availableServices'] ?? [];
+		return $endpoint_status['avaiableServices'] ?? [];
 	}
 
 	/**
--- a/includes/classes/ElasticPressIo.php
+++ b/includes/classes/ElasticPressIo.php
@@ -106,7 +106,7 @@ class ElasticPressIo {
 
 		$endpoint_status = $this->get_endpoint_status( $skip_cache );
 
-		return $endpoint_status['avaiableServices'] ?? [];
+		return $endpoint_status['availableServices'] ?? [];
 	}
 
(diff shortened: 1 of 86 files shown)


includes/classes/ElasticPressIo.php
```diff
@@ -106,7 +106,7 @@ public function get_endpoint_available_services( $skip_cache = false ): array {
 
 		$endpoint_status = $this->get_endpoint_status( $skip_cache );
 
-		return $endpoint_status['avaiableServices'] ?? [];
+		return $endpoint_status['availableServices'] ?? [];
```
felipeelia: This, unfortunately, comes from an external service and will need to stay as is.

tests/e2e/src/specs/status-report.spec.ts
```diff
@@ -67,7 +67,7 @@ test.describe('Status Report', { tag: '@group1' }, () => {
             "message": "Testing message"
         }
     ],
-    "avaiableServices": {
+    "availableServices": {
```
felipeelia: same here

tests/php/features/TestProtectedContent.php
```diff
@@ -461,7 +461,7 @@ public function test_maybe_change_sort() {
 		);
 		$not_so_good_match_id = $this->ep_factory->post->create(
 			[
-				'post_title' => 'not so good match - beautful',
+				'post_title' => 'not so good match - beautiful',
```
felipeelia: This is intentional 😄

felipeelia (changes requested):
Hey @szepeviktor, thanks for the PR and sorry for taking so long to get back to you. Can you implement the changes I've commented on so we can merge this one?

tests/php/features/TestProtectedContent.php
szepeviktor: @felipeelia All done 🍏

PR #4255, opened 2026-01-01
Guard Instant Results facets against invalid taxonomy values

### LINKED ISSUE #4254: BUG: Fatal error in Instant Results facets when ep_facet_include_taxonomies contains a taxonomy slug string (e.g. internal_tags)
### Describe the bug

When ElasticPress Instant Results builds taxonomy facets, it fetches public taxonomies as objects and then applies the ep_facet_include_taxonomies filter. If any plugin/theme returns a taxonomy slug string (or any non-WP_Taxonomy value) in that filtered array, ElasticPress calls get_taxonomy_labels() with a string instead of a WP_Taxonomy object, which triggers a PHP fatal error in core (wp-includes/taxonomy.php:708: “Attempt to assign property labels on string”). This can take down wp-admin (Dashboard load) because Instant Results registers routes/schemas during admin load.

I confirmed this happens in my environment and verified that adding a defensive guard to normalize string slugs via get_taxonomy() (and skipping invalid entries) prevents the fatal.

### Steps to Reproduce

1. Install and activate ElasticPress 5.3.2.

2. Add a filter on ep_facet_include_taxonomies that appends a taxonomy slug string (instead of returning the expected slug => WP_Taxonomy objects array). For example, add this to a plugin or mu-plugin:

```
 add_filter( 'ep_facet_include_taxonomies', function ( $taxonomies ) {
     // Incorrect return type, but currently possible in real-world integrations:
    return array_merge( $taxonomies, [ 'internal_tags' ] );
 }, 1 );
```

 
3. Log in to wp-admin.

4. Visit either of these admin pages:
 Dashboard: /wp-admin/
 ElasticPress admin screen: /wp-admin/admin.php?page=elasticpress
 
**Error observed**

PHP fatal error: Attempt to assign property "labels" on string in wp-includes/taxonomy.php:708

Stack trace shows: ElasticPress\Feature\InstantResults\InstantResults->get_facets() calling get_taxonomy_labels('internal_tags').

Image
Image

**Expected behavior**

ElasticPress should not fatal if ep_facet_include_taxonomies returns unexpected values. Instant Results should validate/normalize each taxonomy entry before calling get_taxonomy_labels() (e.g., if the value is a string slug, resolve it via get_taxonomy(); if it is not a valid WP_Taxonomy, skip it).

_No response_

### Environment information

PHP: 8.4
WordPress: 6.9
ElasticPress: 5.3.2 (works on 4.7.2; fatal on 5.3.2 in my case)

_No response_

### Code of Conduct

laraib15:
### Description of the Change

This PR prevents a fatal error in Instant Results taxonomy facets when the `ep_facet_include_taxonomies` filter returns unexpected values (e.g., a taxonomy slug string such as `internal_tags` rather than a `WP_Taxonomy` object).

`InstantResults::get_facets()` currently assumes each filtered taxonomy entry is a `WP_Taxonomy` object and passes it directly to `get_taxonomy_labels()`. If a third-party integration appends a slug string, WordPress core fatals when it tries to treat a string as an object (e.g., `Attempt to assign property "labels" on string` in `wp-includes/taxonomy.php`).

This change adds a defensive guard in the taxonomy facets loop:
- If the value is a string slug, normalize it via `get_taxonomy( $slug )`
- If the value cannot be resolved to `WP_Taxonomy`, skip it

This prevents wp-admin lockouts while preserving existing behavior for correctly shaped data.

Closes #4254

### How to test the Change

1. Install and activate ElasticPress 5.3.2 and enable Instant Results.
2. Add the reproduction filter from issue #4254 (appends `internal_tags` as a string via `ep_facet_include_taxonomies`).
3. Log in to wp-admin and visit:
   - `/wp-admin/`
   - `/wp-admin/admin.php?page=elasticpress`
4. Confirm both pages load successfully and no fatal error occurs.

Verified on: WordPress 6.9; PHP 8.4.

### Changelog Entry

> Fixed - Prevent fatal error in Instant Results facets when `ep_facet_include_taxonomies` returns non-`WP_Taxonomy` values.

### Credits

Props @laraib15

### Checklist:

### DIFF
--- a/includes/classes/Feature/InstantResults/InstantResults.php
+++ b/includes/classes/Feature/InstantResults/InstantResults.php
@@ -626,6 +626,14 @@ class InstantResults extends Feature {
 		$taxonomies = apply_filters( 'ep_facet_include_taxonomies', $taxonomies );
 
 		foreach ( $taxonomies as $slug => $taxonomy ) {
+			if ( is_string( $taxonomy ) ) {
+				$slug     = $taxonomy;
+				$taxonomy = get_taxonomy( $slug );
+			}
+
+			if ( ! ( $taxonomy instanceof \WP_Taxonomy ) ) {
+				continue;
+			}
 			$name   = 'tax-' . $slug;
 			$labels = get_taxonomy_labels( $taxonomy );
 
--- a/includes/classes/Feature/InstantResults/InstantResults.php
+++ b/includes/classes/Feature/InstantResults/InstantResults.php
@@ -631,21 +631,22 @@ class InstantResults extends Feature {
 				$taxonomy = get_taxonomy( $slug );
 			}
 
-				if ( ! ( $taxonomy instanceof \WP_Taxonomy ) ) {
-					_doing_it_wrong(
-						__METHOD__,
-						sprintf(
-							/* translators: %s is a taxonomy slug. */
-							esc_html__(
-								'Invalid taxonomy "%s" returned via ep_facet_include_taxonomies filter',
-								'elasticpress'
-							),
-							$slug
+			if ( ! ( $taxonomy instanceof \WP_Taxonomy ) ) {
+				_doing_it_wrong(
+					__METHOD__,
+					sprintf(
+					/* translators: %s is a taxonomy slug. */
+						esc_html__(
+							'Invalid taxonomy "%s" returned via ep_facet_include_taxonomies filter',
+							'elasticpress'
 						),
-						'ElasticPress 5.3.2'
-					);
-					continue;
-				}
+						esc_html( $slug )
+					),
+					'ElasticPress 5.3.3'
+				);
+				continue;
+			}
+
 			$name   = 'tax-' . $slug;
 			$labels = get_taxonomy_labels( $taxonomy );
 
--- a/includes/classes/Feature/InstantResults/InstantResults.php
+++ b/includes/classes/Feature/InstantResults/InstantResults.php
@@ -631,9 +631,21 @@ class InstantResults extends Feature {
 				$taxonomy = get_taxonomy( $slug );
 			}
 
-			if ( ! ( $taxonomy instanceof \WP_Taxonomy ) ) {
-				continue;
-			}
+				if ( ! ( $taxonomy instanceof \WP_Taxonomy ) ) {
+					_doing_it_wrong(
+						__METHOD__,
+						sprintf(
+							/* translators: %s is a taxonomy slug. */
+							esc_html__(
+								'Invalid taxonomy "%s" returned via ep_facet_include_taxonomies filter',
+								'elasticpress'
+							),
+							$slug
+						),
+						'ElasticPress 5.3.2'
+					);
+					continue;
+				}
 			$name   = 'tax-' . $slug;
 			$labels = get_taxonomy_labels( $taxonomy );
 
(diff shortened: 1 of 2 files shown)


includes/classes/Feature/InstantResults/InstantResults.php
```diff
@@ -624,6 +624,14 @@ public function get_facets() {
...
 
 		foreach ( $taxonomies as $slug => $taxonomy ) {
+			if ( is_string( $taxonomy ) ) {
+				$slug     = $taxonomy;
+				$taxonomy = get_taxonomy( $slug );
+			}
+
+			if ( ! ( $taxonomy instanceof \WP_Taxonomy ) ) {
+				continue;
+			}
```
felipeelia, line 648: Instead of silently continuing here, let's add a _doing_it_wrong() call, so developers can know something is going wrong. You can follow what we do [here](https://github.com/10up/ElasticPress/blob/01987fb5c0b07c1154e6e8c5644ef06f385ee060/includes/classes/ElasticPressIo.php#L32).

It would be awesome if we could also add a unit test for this method (and this _doing_it_wrong call) to the `tests/php/features/TestInstantResults.php` file.

felipeelia (changes requested):
@laraib15 thanks for the PR! Do you mind addressing my comment there so we can merge it in 5.3.3?

includes/classes/Feature/InstantResults/InstantResults.php
laraib15: Hi! I’ve implemented the requested change.

Added a _doing_it_wrong() call when ep_facet_include_taxonomies returns an invalid taxonomy (includes the taxonomy slug in the message; version set to ElasticPress 5.3.2).

Added a unit test in tests/php/features/TestInstantResults.php

felipeelia: Thanks, @laraib15! I have linters running now, but I can already say some indentations there will need to be fixed (you can see the misalignment when you check your code here on GitHub)

Also, the next version is actually 5.3.3 and not 5.3.2, as you added. Mind making those changes, please? (It is worth waiting for GitHub Actions to run, just in case there is something else.)

laraib15: @felipeelia i have pushed the fixes. Please take a look when you get a chance.

tests/php/features/TestInstantResults.php
```diff
@@ -218,4 +218,59 @@ function () {
...
 		$feature->after_update_feature( 'instant-results', [], [ 'active' => false ] );
 	}
+		/**
+		 * Ensure invalid taxonomy values returned from ep_facet_include_taxonomies
+		 * are skipped and logged via _doing_it_wrong().
+		 *
+		 * @group instant-results
+		 * @since 5.3.3
+		 */
+	public function test_invalid_taxonomy_from_facet_filter_triggers_doing_it_wrong_and_is_skipped() {
```
felipeelia:
```suggestion

	/**
	 * Ensure invalid taxonomy values returned from ep_facet_include_taxonomies
	 * are skipped and logged via _doing_it_wrong().
	 *
	 * @group instant-results
	 * @since 5.3.3
	 */
	public function test_invalid_taxonomy_from_facet_filter_triggers_doing_it_wrong_and_is_skipped() {
```

felipeelia: Indentations will need to be fixed here as well (and also that extra line before the other method and this docblock)

felipeelia (changes requested):
We're almost there, @laraib15! just this last detail and we should be done

tests/php/features/TestInstantResults.php
laraib15: @felipeelia Fixed the indentation issues

PR #4258, opened 2026-01-14
Fix: Create search template for all sites.

### LINKED ISSUE #4227: Instant Results fail on multisite subsites due to missing search template

When you set up ElasticPress on a WordPress multisite, Instant Results and possibly other features fail to work on subsites because the search template cannot be found. You can fix this by resaving Search Fields & Weighting on each subsite.

If this behavior is intentional, ElasticPress should communicate it to users after a full sync to avoid confusion.

### Designs

_No response_

_No response_

### Code of Conduct

burhandodhy:
### Description of the Change

This PR ensures that the sync creates the search template across the network when the plugin is network-wide. It also adds support for the `network-wide` flag for `put-search-template` command.

Additionally, we don’t need to set the `index` property during the constructor because if we run `epio_save_site_search_template`, the index property still holds the old value, which can cause incorrect endpoints.

Closes #4227 

### How to test the Change

1. Enable the ElasticPress network-wide.
2. Enable the Instant Results.
3. Create a new site. 
4. Run Sync and ensure that Instant Results works find on newly created site. 

### Changelog Entry

> Fixed - Ensure the Sync creates the Instant Resutls templates for all the sites

### Credits

Props @burhandodhy @anjulahettige 

### Checklist:

### DIFF
--- a/includes/classes/Command.php
+++ b/includes/classes/Command.php
@@ -1545,11 +1545,16 @@ class Command extends WP_CLI_Command {
 	 * : Delete the template for all sites in the network if plugin is network activated. Otherwise, delete the template for the current site only.
 	 *
 	 * @since 4.5.0
+	 * @param array $args Positional CLI args.
+	 * @param array $assoc_args Associative CLI args.
 	 * @subcommand delete-search-template
 	 */
-	public function delete_search_template() {
+	public function delete_search_template( $args, $assoc_args ) {
+		$network_wide    = \WP_CLI\Utils\get_flag_value( $assoc_args, 'network-wide', false );
 		$instant_results = Features::factory()->get_registered_feature( 'instant-results' );
-		$instant_results->epio_delete_search_template();
+
+		$instant_results->epio_delete_site_search_template( $network_wide );
+
 		WP_CLI::success( esc_html__( 'Done.', 'elasticpress' ) );
 	}
 
--- a/includes/classes/Command.php
+++ b/includes/classes/Command.php
@@ -1517,17 +1517,32 @@ class Command extends WP_CLI_Command {
 	/**
 	 * Saves the Instant Results search template to EPIO.
 	 *
+	 * ## OPTIONS
+	 *
+	 * [--network-wide]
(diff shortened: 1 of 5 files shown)


includes/classes/Command.php
```diff
@@ -1517,18 +1517,33 @@ public function get_search_template( $args, $assoc_args ) {
...
 	}
 
 	/**
 	 * Deletes the Instant Results search template.
 	 *
+	 * ## OPTIONS
+	 *
+	 * [--network-wide]
+	 * : Delete the template for all sites in the network if plugin is network activated. Otherwise, delete the template for the current site only.
+	 *
```
felipeelia, line 1547: Shouldn't this also affect the method code?

includes/classes/Feature/InstantResults/InstantResults.php
```diff
@@ -919,4 +919,38 @@ protected function set_settings_schema() {
...
+	public function epio_save_site_search_template( bool $network_wide = false ): void {
+		if ( ! $network_wide ) {
+			$this->epio_save_search_template();
+			return;
+		}
+
+		$sites = Utils\get_sites( 0, true );
+		foreach ( $sites as $site ) {
+			switch_to_blog( $site['blog_id'] );
+			$this->index = Indexables::factory()->get( 'post' )->get_index_name();
```
felipeelia: Any reason why we need this?

felipeelia (commented):
Few questions for you on this one @burhandodhy

includes/classes/Feature/InstantResults/InstantResults.php
burhandodhy: It shouldn’t be there. I added that before making this [change](https://github.com/10up/ElasticPress/pull/4258/files#diff-e36353e23234caf482aa97381485e86ad9ffbe3220a70946c92d82ca58168314L78)

includes/classes/Command.php
burhandodhy: It was my mistake; I have added the code.

PR #4263, opened 2026-01-15
Introduce new ep_asciifolding filter

### LINKED ISSUE #4212: Add asciifolding analyzer with preserve_original = true

Currently, we use the [asciifolding](https://www.elastic.co/docs/reference/text-analysis/analysis-asciifolding-tokenfilter) analyzer in several places. This analyzer converts non-ASCII characters into their ASCII equivalents, for example, “é” becomes “e”.

While this works well for normalizing text, it can cause issues when users search for words containing accents. Since the analyzer removes accents, Elasticsearch may not return the best results when a user searches for the original accented form of a word.

To improve this behavior, we should enable `preserve_original` = `true` in the asciifolding token filter configuration. This setting ensures that both the ASCII and the original tokens are indexed, allowing searches for both accented and non-accented variations to return relevant results.

### Designs

_No response_

_No response_

### Code of Conduct

burhandodhy:
### Description of the Change

This PR adds a new `ep_asciifolding` filter so that words with accented characters can be searched without any issues.

Closes #4212

### Changelog Entry

> Fixed - An issue where posts were not returned correctly when the search term contained accented characters.
 
### Credits

Props @burhandodhy 

### Checklist:

### DIFF
--- a/includes/mappings/comment/7-0.php
+++ b/includes/mappings/comment/7-0.php
@@ -65,14 +65,14 @@ return [
 				'default'          => [
 					'tokenizer' => 'standard',
 					/* This filter is documented in includes/mappings/post/7-0.php */
-					'filter'    => apply_filters( 'ep_default_analyzer_filters', [ 'lowercase', 'ep_stop', 'ewp_snowball' ] ),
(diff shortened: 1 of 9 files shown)


felipeelia (commented):
@burhandodhy are you 100% sure this works on ES 5?

felipeelia (changes requested):
@burhandodhy this seems to be consistently failing on a specific e2e test for EP.io. Do you mind giving it a look please?

PR #4284, opened 2026-02-18
fix/4193: Fix meta facet counts for late-rendered facet-meta blocks

### LINKED ISSUE #4193: BUG: Meta filter block isn’t working when added via code.
### Describe the bug

When calling the Meta Filter block directly in code like this `do_blocks( '<!-- wp:elasticpress/facet-meta {"facet":"affiliate","displayCount":true} /-->' )` the block doesn’t work.

It seems the block depends on [WP_Widget_Block](https://github.com/10up/ElasticPress/blob/5.2.0/includes/classes/Feature/Facets/Types/Meta/FacetType.php#L271) to fetch the list of meta, so when inserted directly via code, it fails to render properly.

The same issue exists on the Meta Range Filter.

### Steps to Reproduce

Add the Meta Filter via code and check if it works

_No response_

### Environment information

_No response_

_No response_

### Code of Conduct

Sidsector9:
#### What was the issue?

The meta facet aggregations were built during the main facetable query, based on facet fields discovered early. When `elasticpress/facet-meta` is rendered later at runtime (for example via `do_blocks()`), the main query has already completed. In that case, the requested meta field is missing from the original aggregation request, so that facet receives no counts and terms render as disabled/empty.

#### How does this fix work?
This PR adds a runtime-safe fallback for meta facet aggregations:
- Adds the `get_facet_aggregation_for_field()` in the meta facet type.
- If the aggregation for the requested meta field is missing, it runs a scoped facetable fallback query for that field in the current query context.
- It merges the returned aggregation back into the original query’s facet aggregations.
- Updates the meta facet renderer to use this method instead of directly reading aggregation data.
- This allows late-rendered meta facet blocks to display correct counts and clickable terms.

Closes #4193 

### How to test the Change

Replace `hero_caption` from the following with a meta value that is relevant to your site, and add it to your functions.php file:

```php
add_action( 'wp_footer', function() {
	if ( ! is_archive() ) {
		return;
	}
	echo do_blocks( '<!-- wp:elasticpress/facet-meta {"displayCount":true,"facet":"hero_caption"} /-->' );
} );
```

On `develop`, it will display the the list, but the counts will be 0, making the list items not clickable. Switch to the fix branch and observe that it displays the count, and the items are clickable.

### Changelog Entry

> Fixed - Bug fix
### Credits

Props @burhandodhy @Sidsector9 

### Checklist:

### DIFF
--- a/includes/classes/Feature/Facets/Types/Meta/FacetType.php
+++ b/includes/classes/Feature/Facets/Types/Meta/FacetType.php
@@ -9,6 +9,7 @@
 namespace ElasticPress\Feature\Facets\Types\Meta;
 
 use ElasticPress\Features;
+use WP_Query;
 
 if ( ! defined( 'ABSPATH' ) ) {
 	exit; // Exit if accessed directly.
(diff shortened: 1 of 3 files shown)


Sidsector9:
As discussed on Slack, this needs to be document instead of code changes. Closing this one.

PR #4285, opened 2026-02-19
Assumption safety checks for queried object

### LINKED ISSUE #3331: BUG: EP crashes trying to use queried object that doesn't exist
### Describe the bug

This code in `includes/classes/Indexable/Post/QueryIntegration.php` assumes that `$query->get_queried_object()` always returns a term, this is not a safe assumption:

```php
		if ( empty( $query_vars['post_type'] ) ) {
			if ( $query->is_tax() ) {
				$query_vars['post_type'] = get_taxonomy( $query->get_queried_object()->taxonomy )->object_type;
			} elseif ( empty( $query_vars['s'] ) ) {
```

### Steps to Reproduce

Not 100% on the steps that lead to this, only that this is what I got at the other end

_No response_

### Environment information

_No response_

latest version of elasticpress 4.4.1

### Code of Conduct

ZacharyRener:
### Description of the Change

Closes #3331 

### Changelog Entry

> Fixed - Error when trying to use queried object that doesn't exist

### Credits

Props @tomjn @ZacharyRener 

### Checklist:

### DIFF
--- a/includes/classes/Indexable/Post/QueryIntegration.php
+++ b/includes/classes/Indexable/Post/QueryIntegration.php
@@ -246,7 +246,7 @@ class QueryIntegration {
 		 * If not search and not set, default to post. If not set and is search, use searchable post types.
 		 */
 		if ( empty( $query_vars['post_type'] ) ) {
-			$tax_post_type = Utils\get_post_type_for_tax_query( $query );
+			$tax_post_type = Utils\get_post_types_for_tax_query( $query );
 
 			if ( ! empty( $tax_post_type ) ) {
 				$query_vars['post_type'] = $tax_post_type;
--- a/includes/classes/Indexable/Post/QueryIntegration.php
+++ b/includes/classes/Indexable/Post/QueryIntegration.php
@@ -246,13 +246,15 @@ class QueryIntegration {
 		 * If not search and not set default to post. If not set and is search, use searchable post types
 		 */
 		if ( empty( $query_vars['post_type'] ) ) {
-			if ( $query->is_tax() && $query->get_queried_object() ) {
-				$query_vars['post_type'] = get_taxonomy( $query->get_queried_object()->taxonomy )->object_type;
-			} elseif ( empty( $query_vars['s'] ) ) {
-				$query_vars['post_type'] = 'post';
-			} else {
-				$query_vars['post_type'] = array_values( get_post_types( array( 'exclude_from_search' => false ) ) );
-			}
+			$query_vars['post_type'] = Utils\get_post_type_for_tax_query( $query );
+		}
+
+		if ( empty( $query_vars['post_type'] ) && ! empty( $query_vars['s'] ) ) {
+			$query_vars['post_type'] = array_values( get_post_types( array( 'exclude_from_search' => false ) ) );
+		}
+
+		if ( empty( $query_vars['post_type'] ) ) {
+			$query_vars['post_type'] = 'post';
 		}
 
 		/**
--- a/includes/classes/Indexable/Post/QueryIntegration.php
+++ b/includes/classes/Indexable/Post/QueryIntegration.php
@@ -243,18 +243,18 @@ class QueryIntegration {
 		}
 
 		/**
-		 * If not search and not set default to post. If not set and is search, use searchable post types
+		 * If not search and not set default to post. If not set and is search, use searchable post types.
 		 */
 		if ( empty( $query_vars['post_type'] ) ) {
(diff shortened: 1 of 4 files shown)


includes/classes/Indexable/Post/QueryIntegration.php
```diff
@@ -243,11 +243,13 @@ public function get_es_posts( $posts, $query ) {
 		}
 
 		/**
-		 * If not search and not set default to post. If not set and is search, use searchable post types
+		 * If not search and not set default to post. If not set and is search, use searchable post types.
```
felipeelia:
```suggestion
		 * If not search and not set, default to post. If not set and is search, use searchable post types.
```

includes/utils.php
```diff
@@ -930,3 +930,34 @@ function is_top_level_admin_context() {
...
+ * Safely resolve the post type(s) for a taxonomy query.
+ *
+ * Guards against null queried objects, missing taxonomy properties,
+ * and deregistered taxonomies that would otherwise cause fatal errors
+ * when chaining get_queried_object()->taxonomy through get_taxonomy().
+ *
+ * Returns null when the query is not a taxonomy query, allowing callers
+ * to distinguish "not a tax query" from "tax query that failed to resolve."
+ *
+ * @since 5.3.0
```
felipeelia:
```suggestion
 * @since 5.3.3
```

felipeelia (changes requested):

includes/utils.php
```diff
@@ -930,3 +930,34 @@ function is_top_level_admin_context() {
...
+ *
+ * @param \WP_Query|null $query Optional. WP_Query instance. Defaults to the global query.
+ * @return array|null Post type array on success, null otherwise.
+ */
+function get_post_type_for_tax_query( $query = null ) {
```
tomjn: We could type hint this as `?WP_Query`, this does mean that we would need to enforce a minimum PHP version of ...... v7.1

tomjn: likewise a return type could be added of `?array`

tomjn: and since this is returning an array of types plural, should it be renamed to `get_post_types_for_tax_query`?

includes/utils.php
```diff
@@ -930,3 +930,34 @@ function is_top_level_admin_context() {
...
+	}
+
+	$queried_object = $query instanceof \WP_Query ? $query->get_queried_object() : get_queried_object();
+	if ( ! $queried_object || ! isset( $queried_object->taxonomy ) ) {
+		return null;
+	}
+
+	$taxonomy_object = get_taxonomy( $queried_object->taxonomy );
+
+	return $taxonomy_object ? $taxonomy_object->object_type : null;
```
tomjn: While `object_type` can be null, it can also be a value that is not a post type, e.g. a comments or users. While less common, a user taxonomy would lead to it trying to work with a `user` CPT so there's some validation needed.

tomjn: Also what's the difference between returning `null` and returning `[]`? Could it just return an empty array and skip the whole `null` thing?

includes/utils.php
```diff
@@ -930,3 +930,34 @@ function is_top_level_admin_context() {
...
+ * @param \WP_Query|null $query Optional. WP_Query instance. Defaults to the global query.
+ * @return array|null Post type array on success, null otherwise.
+ */
+function get_post_type_for_tax_query( $query = null ) {
+	$is_tax = $query instanceof \WP_Query ? $query->is_tax() : is_tax();
+	if ( ! $is_tax ) {
+		return null;
+	}
+
+	$queried_object = $query instanceof \WP_Query ? $query->get_queried_object() : get_queried_object();
```
tomjn: these ternaries could be avoided with a guard that assigns `$query` the value of the `$wp_query` global if it is null

ZacharyRener:
@tomjn Points addressed with latest push.

tomjn (approved):

includes/utils.php
```diff
@@ -930,3 +930,39 @@ function is_top_level_admin_context() {
...
+ *
+ * @param \WP_Query|null $query Optional. WP_Query instance. Defaults to the global query.
+ * @return array Registered post type names on success, empty array otherwise.
+ */
+function get_post_types_for_tax_query( ?\WP_Query $query = null ): array {
+	if ( null === $query ) {
+		global $wp_query;
+		$query = $wp_query;
+	}
+
```
felipeelia, line 953:
```suggestion
	global $wp_query;

	if ( null === $query ) {
		$query = $wp_query;
	}

```

felipeelia (changes requested):

PR #4290, opened 2026-02-22
feat/3954: Add filters to exclude post types and taxonomy terms from Instant Results

### LINKED ISSUE #3954: Remove Filters from Instant Results
### Describe the bug

We'd like to remove filters from the Instant Results. Currently, we can remove them with CSS, but the problem is the "Show More Options" still appears even though there are no more options.

### Steps to Reproduce

1. Go to Instant Results
2. Add "Post Type" for filter
3. No options exist to remove single or multiple post types. It will show All post types.

If removing via filters via CSS the 'Show More Options' button appears. See here: [(https://youtu.be/uN1IXzdFpEc)]
This indicates a better solution is necessary to remove these filters so they are not counted in the list to trigger the show more options button.

### Environment information

_No response_

_No response_

### Code of Conduct

Sidsector9:
### Description of the Change

- Adds `ep_instant_results_excluded_post_types` filter to exclude post types and related posts from Instant Results.
- Adds `ep_instant_results_excluded_term_ids` filter to exclude taxonomy terms from Instant Results facets.

#### Post Type Exclusion (`ep_instant_results_excluded_post_types`)

Excluded post types are removed from both the Post Type facet and the search template itself. This means content of that type will not appear in Instant Results at all.

The search template is regenerated when Instant Results settings are saved, when the weighting configuration is saved, and after a full sync. So after changing this filter, save either settings screen, run a sync, or run `wp elasticpress put-search-template`, otherwise the old template stays in use.

```php
add_filter( 'ep_instant_results_excluded_post_types', function( $excluded ) {
    $excluded[] = 'page';
    return $excluded;
} );
```

#### Term Exclusion (`ep_instant_results_excluded_term_ids`)

Excluded terms are hidden from every taxonomy facet (Category, Tag, etc.) but posts associated with those terms still appear in search results. This is intentional, a post can belong to multiple terms, so excluding a term from the facet should not remove posts that also match through other, non-excluded terms.

```php
add_filter( 'ep_instant_results_excluded_term_ids', function( $excluded ) {
    $excluded[] = 1; // Uncategorized
    return $excluded;
} );
```

Closes #3954 

### Changelog Entry

> Added: `ep_instant_results_excluded_post_types` filter to exclude post types from Instant Results facets and search results.         
> Added: `ep_instant_results_excluded_term_ids` filter to exclude taxonomy terms from Instant Results facets.

### Credits

Props @feliciaoctocog, @Sidsector9 

### Checklist:


felipeelia:
We will need tests for this one, @Sidsector9.

Sidsector9:
@felipeelia the tests have been added 👍

felipeelia (changes requested):
@Sidsector9 We will need to replace all the `5.4.0` with `5.3.4`. In addition to that, can you please change `get_excluded_post_types` and `get_excluded_term_ids` returns, so we always make sure they are sending the correct types back to the caller?

Something like `array_filter( array_map( 'absint', (array) apply_filters( 'ep_instant_results_excluded_term_ids', [] ) ) );`, but properly written.

PR #4319, opened 2026-06-11
Fix/utils get site null guard

### LINKED ISSUE #4318: Utils\get_site() fatal TypeError on PHP 8.x when site ID does not exist

thisismyurl:
### Description of the Change

closes #4318

### Changelog Entry

> Added - New feature
> Removed - Feature
> Fixed - Bug fix
> Security - Vulnerability
> Developer - Non-functional update

### Credits

Props @username, ...

### Checklist:

### DIFF
--- a/includes/utils.php
+++ b/includes/utils.php
@@ -204,6 +204,10 @@ function is_site_indexable( $blog_id = null ) {
 
 	$site = get_site( $blog_id );
 
+	if ( empty( $site ) ) {
+		return false;
+	}
+
 	$is_indexable = get_site_meta( $site['blog_id'], 'ep_indexable', true );
 
 	return 'no' !== $is_indexable && ! $site['deleted'] && ! $site['archived'] && ! $site['spam'];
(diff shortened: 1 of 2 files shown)


thisismyurl:
Thanks for the review — addressing both points:

**`PHP_INT_MAX` in tests** — Valid concern. I pushed a follow-up commit that replaces `PHP_INT_MAX` with `999999` — a fixed large value that stays in a typical 32-bit range and is deterministic across environments.

**Returning `[]` vs `null`** — The empty array is intentional here. The existing callers (including `is_site_indexable()` in this same PR) guard with `empty($site)`, which is falsy for both `null` and `[]`. Returning `[]` keeps the return type consistent (`array`) without introducing a type union, and `empty()` handles it cleanly at the call site. I've noted in the function's docblock that an empty array is returned when the site is not found, so callers know to guard accordingly. Happy to discuss further if the maintainers prefer a different contract — just wanted to explain the reasoning.

PR #4325, opened 2026-06-21
fix: sync posts to Elasticsearch when Post Search is disabled

### LINKED ISSUE #4158: BUG: Post doesn't update in Elasticsearch when "Post Search" is disabled
### Describe the bug

When the "Post Search" feature is disabled, updating a post does not sync the changes to Elasticsearch.

### Steps to Reproduce

1. Go to the Features page and disable "Post Search".
2. Edit any post and make some changes.
3. Check the post in Elasticsearch and compare it with the updated content in WordPress.

_No response_

### Environment information

_No response_

_No response_

### Code of Conduct

faisalahammad:
## Summary
Move the Post indexable activation out of the Post Search feature `setup()` so posts continue to sync to Elasticsearch even when the Post Search feature is disabled.

Previously, disabling Post Search prevented `Post::setup()` from being called, which meant the `SyncManager` was never instantiated and `wp_insert_post` / `delete_post` / meta / term hooks were never attached.

Fixes #4158

### includes/classes/Features.php
**Before:**
```php
do_action( 'ep_setup_features' );

foreach ( $this->registered_features as $feature ) {
```

**After:**
```php
do_action( 'ep_setup_features' );

/**
 * Activate the post indexable regardless of whether the Post Search
 * feature is enabled, so posts continue to sync to Elasticsearch.
 */
Indexables::factory()->activate( 'post' );

foreach ( $this->registered_features as $feature ) {
```

**Why:** `setup_features()` runs on `init` priority 0 for every request. Activating the post indexable here ensures it is always set up, independent of whether the Search feature is active.

### includes/classes/Feature/Search/Search.php
**Before:**
```php
public function setup() {
    Indexables::factory()->activate( 'post' );

    add_action( 'init', [ $this, 'search_setup' ] );
    // ...
}
```

**After:**
```php
public function setup() {
    add_action( 'init', [ $this, 'search_setup' ] );
    // ...
}
```

**Why:** The Search feature no longer needs to activate the post indexable. Its `setup()` still loads weighting, synonyms, and registers `search_setup` for query integration.

## Testing
**Test 1: Post sync with Post Search disabled**
1. Activate ElasticPress and connect to Elasticsearch.
2. Go to ElasticPress → Features and disable Post Search.
3. Edit any post and save.
4. Verify the document in Elasticsearch reflects the updated content.

Result: works as expected.

**Test 2: Post sync with Post Search enabled (regression)**
1. Enable Post Search.
2. Edit and save a post.
2. Verify the document in Elasticsearch updates.

Result: works as expected.

**Test 3: Automated regression test**
Added `tests/php/indexables/TestPostSyncWithoutSearch.php` with `testPostSyncWithSearchDisabled()` — disables the Search feature, creates a post, and asserts the sync hook fires and the document exists in Elasticsearch. Passes on single-site and multisite test suites.

### DIFF
--- a/includes/classes/Feature/Search/Search.php
+++ b/includes/classes/Feature/Search/Search.php
@@ -97,8 +97,6 @@ class Search extends Feature {
 	 * @since 2.2
 	 */
 	public function setup() {
-		Indexables::factory()->activate( 'post' );
-
 		add_action( 'init', [ $this, 'search_setup' ] );
 
 		// Set up weighting sub-module
--- a/includes/classes/Features.php
+++ b/includes/classes/Features.php
@@ -383,6 +383,12 @@ class Features {
 		 */
 		do_action( 'ep_setup_features' );
 
+		/**
+		 * Activate the post indexable regardless of whether the Post Search
+		 * feature is enabled, so posts continue to sync to Elasticsearch.
+		 */
+		Indexables::factory()->activate( 'post' );
+
 		foreach ( $this->registered_features as $feature ) {
 			$feature->set_i18n_strings();
 
--- a/tests/php/indexables/TestPostSyncWithoutSearch.php
+++ b/tests/php/indexables/TestPostSyncWithoutSearch.php
@@ -1,0 +1,64 @@
+<?php
+/**
+ * Test post sync when Post Search feature is disabled
+ *
+ * @package elasticpress
+ */
+
+namespace ElasticPressTest;
+
+use ElasticPress;
+
+/**
+ * Test post sync without search feature class
+ */
+class TestPostSyncWithoutSearch extends BaseTestCase {
+
+	/**
+	 * Setup each test.
+	 *
+	 * @since 5.4.0
+	 */
+	public function set_up() {
+		global $wpdb;
+		parent::set_up();
+		$wpdb->suppress_errors();
+
+		$admin_id = $this->factory->user->create( array( 'role' => 'administrator' ) );
+		wp_set_current_user( $admin_id );
+
+		ElasticPress\Elasticsearch::factory()->delete_all_indices();
+
+		/**
+		 * The post indexable should be activated by the plugin bootstrap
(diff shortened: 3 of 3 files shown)


burhandodhy:
Hi @faisalahammad,

Thank you for taking the time to open this pull request.

However, this issue is a bit more complex than it appears. If we simply make the post indexable, ElasticPress will start using Elasticsearch for post queries even when the **Post Search** feature is disabled.

This requires some discussion and a decision. For example, should posts still be indexed during sync when the **Post Search** feature is disabled, or should they be excluded from indexing altogether?

I would love to hear your thoughts on this before we move forward.

Regards,
Burhan

faisalahammad:
Hi @burhandodhy, good question. Checked this in the code.

Activating the post indexable only wires up `SyncManager` and `QueryIntegration`. `QueryIntegration` does not automatically take over `WP_Query` for posts, it only does that when `elasticpress_enabled()` returns true. That check reads the `ep_elasticpress_enabled` filter, and that filter is added only inside `Search::search_setup()`, which only runs when the Post Search feature is active.

So with this change:
- Post Search disabled: posts still sync to ES (the actual bug fix), but `WP_Query` for posts does not go through ES. Same behavior as before this PR for query integration.
- Post Search enabled: no change at all.

Added a second test, `testPostQueryNotIntegratedWithSearchDisabled`, that runs a normal `WP_Query` search with Post Search disabled and asserts `elasticsearch_success` is not set. This should cover the concern directly.

To answer your question directly: posts should keep syncing during indexing even when Post Search is disabled. Other features (WooCommerce, Autosuggest, RelatedPosts, Facets, SearchOrdering, Documents) all depend on the post indexable being populated, and none of them currently activate it themselves. If we exclude posts from sync when Search is off, those features silently break for any site that has Search disabled but still uses them.

faisalahammad:
@burhandodhy pushed the added test and updated branch against develop. Ready for another look when you have time.

PR #4327, opened 2026-06-21
feat(search-ordering): allow removing posts from custom search results

### LINKED ISSUE #4245: Remove posts in Custom Search Result
Custom Search Result is a very nice feature within this plugin.

But at the moment you can only add posts or pages to a custom search result. You can not remove them out of the custom search result. 

The only solution is to remove the whole custom search result and add it again with the correct posts or pages.

Can you add a button to remove records added to custom search results? Maybe a trashbin icon or a x near the icon to reorder the records?

### Designs

_No response_

_No response_

### Code of Conduct

faisalahammad:
## Summary
Added ability to remove individual posts from a custom search result.
Excluded posts are now stored and filtered out from the final search results for that query.
Fixes #4245

### Frontend (`assets/js/ordering/pointers.js`)
**Before:**
```js
removePointer(pointer) {
    // Only removed reordered default results via removedPointers.
}
```
**After:**
```js
removePointer(pointer) {
    if (pointer.type === 'custom-result') {
        excludedPosts.push(pointer.ID);
    } else {
        removedPointers.push(pointer.ID);
    }
    // Updates state and hidden input excluded_posts
}
```
**Why:** Differentiates between custom-added results (should be excluded) and reordered defaults (should be returned to original position).

### Backend (`includes/classes/Feature/SearchOrdering/SearchOrdering.php`)
**Before:**
```php
function save_post($post_id) {
    // Saved pointers only.
}
function posts_results($posts) {
    // Injected custom results but did not filter any posts.
}
```
**After:**
```php
function save_post($post_id) {
    // Now reads $_POST['excluded_posts'], sanitizes, and stores in post meta 'excluded_posts'.
    // Clears excluded list when search term changes.
}
function posts_results($posts) {
    // After injecting custom results, fetches excluded list and removes those posts from results.
}
```
**Why:** Persists the exclusion list per custom search result and applies it to query results.

### Tests (`tests/php/features/TestSearchOrdering.php`)
Added three new PHPUnit tests:
- `testSavePostWithExcludedPosts`
- `testPostsResultsFiltersExcludedPosts`
- `testExcludedPostsClearedOnSearchTermChange`

## Testing
**Test 1: Exclude a post from a custom search result**
1. Create a custom search result for "example".
2. Add a post to the result.
3. Click the trash icon on the added post.
4. Save the custom search result.
5. Search for "example" on the frontend.
Result: The excluded post does not appear in the search results.

**Test 2: Automated tests pass**
Run `composer run test --filter TestSearchOrdering`.
Result: All tests pass.

### DIFF
--- a/assets/js/ordering/pointers.js
+++ b/assets/js/ordering/pointers.js
@@ -67,6 +67,7 @@ export class Pointers extends Component {
 			searchText: '',
 			searchResults: {},
 			removedPointers: [],
+			excludedPosts: window.epOrdering.excludedPosts || [],
 		};
 	}
 
@@ -121,18 +122,23 @@ export class Pointers extends Component {
 
 	removePointer = (pointer) => {
 		let { pointers } = this.state;
-		const { removedPointers } = this.state;
+		const { removedPointers, excludedPosts } = this.state;
 
 		delete pointers[pointers.indexOf(pointer)];
 		pointers = pointers.filter((item) => item !== null);
-		removedPointers.push(pointer.ID);
 
-		this.setState({ pointers });
+		if (pointer.type === 'custom-result') {
+			excludedPosts.push(pointer.ID);
+		} else {
+			removedPointers.push(pointer.ID);
+		}
+
+		this.setState({ pointers, removedPointers, excludedPosts });
 	};
 
 	getMergedPosts = () => {
 		let { pointers } = this.state;
-		const { title, defaultResults } = this.state;
+		const { title, defaultResults, excludedPosts } = this.state;
 		let merged = defaultResults[title].slice();
 
 		pointers = pointers.sort((a, b) => {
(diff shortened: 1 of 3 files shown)


burhandodhy (changes requested):
Hi @faisalahammad,

Thank you for opening the PR. I checked out your branch locally, but the posts are not appearing in the Custom Search page on my end. Could you please look into that?

Also, please add the E2E tests. Once that's done, please ping me, and I'll review the code.

Image

faisalahammad (commented):
Review fixes applied for CodeRabbit + internal review:

- Fixed major: removed excludedPosts filter from searchResults so re-add via UI now possible (previously unreachable).
- Fixed minor: also clear removedPointers on re-add.
- JS formatted + lint clean.
- No other high issues. Lows (pre-existing patterns) left as-is per minimal change.
- Tests cover core paths (env limits full run).
- Build ready, manual instructions provided.

All actionable addressed. Ready for re-review.

faisalahammad (commented):
Superseded by comment below.

faisalahammad:
@burhandodhy Thanks for catching that. Found two bugs:

1. **Post titles not showing** — the `<strong class="title">` element was accidentally removed in the diff. Restored it.
2. **Key mismatch** — PHP sends `excluded_posts` (snake_case) but JS was reading `window.epOrdering.excludedPosts` (camelCase). Fixed to read `excluded_posts` so existing exclusions load on page edit.

Both pushed. Please try again.

Re: Instant Results — this is pre-existing. Custom results use `posts_results` filter; Instant Results uses `ep_search_template` + JS rendering (separate path). Not introduced by this PR.

Re: E2E — unit tests cover the new exclude/save/clear logic. E2E can be a follow-up.

PR #4330, opened 2026-06-21
fix(did-you-mean): suppress suggestions for valid queries and matching terms

### LINKED ISSUE #3602: BUG: "Did You Mean" feature offers suggestions even if the current search term is more appropriate
### Describe the bug

While testing the plugin with the Theme Unit Test data and WooCommerce sample data I did a search for “Shirt” (there are several shirt products in the WooCommerce sample data) and it showed me the results, but it also said “Did you mean: shift”, which was strange because “shirt” is a valid word, and even has more results than a search “shift”. 

### Steps to Reproduce

1. In a fresh WordPress install import the [Theme Unit Test](https://github.com/WordPress/theme-test-data) data.
2. Install WooCommerce, and import the [Sample Data](https://woocommerce.com/document/importing-woocommerce-sample-data/).
3. Install ElasticPress, activate the _Did You Mean_ feature, and Sync.
4. Add `<?php do_action( 'ep_suggestions' ); ?>` to the template used by the theme for search results, or activate _Instant Results_.
5. Perform a search for "shirt". A suggestion to search for "shift" should appear.

_No response_

### Environment information

_No response_

_No response_

### Code of Conduct

faisalahammad:
## Summary

"Did You Mean" currently shows the top suggestion even when the current search term already returns results. This leads to confusing suggestions like "Did you mean: shift?" when searching for "shirt" with valid results. This fix suppresses the suggestion when the query has results, or when the suggestion matches the original term. It also applies the same guard to the redirect behavior to prevent a redirect loop.

Closes #3602

### `includes/classes/Feature/DidYouMean/DidYouMean.php`

**`get_suggestion()`** now returns `false` early in two cases:

**Before:**
```php
$term = $this->get_suggested_term( $query );
if ( empty( $term ) ) {
    return false;
}

$html = sprintf( '<span class="ep-spell-suggestion">%s: <a href="%s">%s</a>?</span>', ... );
```

**After:**
```php
$suppress_when_results_exist = apply_filters( 'ep_suggestion_suppress_when_results_exist', true, $query );

if ( $suppress_when_results_exist && $query->found_posts > 0 ) {
    return false;
}

$term = $this->get_suggested_term( $query );
if ( empty( $term ) ) {
    return false;
}

$search_term = $query->query_vars['s'] ?? '';
if ( strtolower( $term ) === strtolower( $search_term ) ) {
    return false;
}

$html = sprintf( '<span class="ep-spell-suggestion">%s: <a href="%s">%s</a>?</span>', ... );
```

**Why:** This matches the existing behavior of `get_alternatives_terms()`, which already hides the alternative list when the query has results. The case-insensitive match guard prevents the feature from suggesting the same term the user already typed.

A new `ep_suggestion_suppress_when_results_exist` filter (default `true`) was added per review feedback, so sites that want the suggestion to still show up alongside results can opt back in with one line, instead of the fix being all-or-nothing.

**`automatically_redirect_user()`** also skips redirecting if the suggested term equals the current search term:

**After:**
```php
$term = $this->get_suggested_term( $wp_query );
if ( empty( $term ) ) {
    return;
}

$search_term = $wp_query->query_vars['s'] ?? '';
if ( strtolower( $term ) === strtolower( $search_term ) ) {
    return;
}
```

**Why:** Prevents an infinite redirect loop when the phrase suggester returns the same term as the original query.

## How to test the Change

**Test 1: Valid query with results**
1. Create posts containing "shirt" and "shift".
2. Sync the index.
3. Search for "shirt".
Result: results load and no "Did you mean: shift?" message appears.

**Test 2: Typo with no results**
1. Search for "shirf".
Result: "Did you mean: shirt?" appears because the query has no results.

**Test 3: Same-term suggestion**
1. Mock or trigger a query where the top suggestion equals the original term (e.g., "shirt" → "shirt" or "Shirt").
Result: no suggestion is displayed.

**Test 4: Restore suggestion via filter**
1. Add `add_filter( 'ep_suggestion_suppress_when_results_exist', '__return_false' );` to a must-use plugin.
2. Repeat Test 1.
Result: "Did you mean: shift?" now shows again alongside the "shirt" results.

## Unit tests

Tests added/updated in `tests/php/features/TestDidYouMean.php`:

- `testGetSuggestionNotShownWhenResultsExist`
- `testGetSuggestionNotShownWhenTopSuggestionMatchesOriginalTerm`
- `testGetSuggestionNotShownWhenTopSuggestionMatchesOriginalTermCaseInsensitive`
- `testGetSuggestionSuppressWhenResultsExistFilter`

Run with:

```bash
vendor/bin/phpunit --filter TestDidYouMean
```

All 21 tests pass. Full `composer run test-single-site` suite also run (944 tests) with no regressions (1 pre-existing unrelated failure in `TestInstantResults` on WP 7.0, not touched by this PR).

### Changelog Entry

> Fixed - "Did You Mean" no longer suggests a term when the current search already has results or when the suggestion matches the original term (case-insensitive), preventing confusing suggestions and redirect loops. A new `ep_suggestion_suppress_when_results_exist` filter allows restoring the previous behavior.

### Credits

Props @faisalahammad

### Checklist:

### DIFF
--- a/includes/classes/Feature/DidYouMean/DidYouMean.php
+++ b/includes/classes/Feature/DidYouMean/DidYouMean.php
@@ -127,8 +127,19 @@ class DidYouMean extends Feature {
 			return false;
 		}
 
+		/**
+		 * Filter whether to suppress the suggestion when the current query already returned results.
+		 *
+		 * @since 5.4.0
+		 * @hook ep_suggestion_suppress_when_results_exist
+		 * @param {bool}     $suppress Whether to suppress the suggestion. Default true.
+		 * @param {WP_Query} $query    The WP_Query object.
+		 * @return {bool} New value
+		 */
+		$suppress_when_results_exist = apply_filters( 'ep_suggestion_suppress_when_results_exist', true, $query );
+
 		// No need to suggest alternatives if the current query already returned results.
-		if ( $query->found_posts > 0 ) {
+		if ( $suppress_when_results_exist && $query->found_posts > 0 ) {
 			return false;
 		}
 
--- a/includes/classes/Feature/DidYouMean/DidYouMean.php
+++ b/includes/classes/Feature/DidYouMean/DidYouMean.php
@@ -127,8 +127,19 @@ class DidYouMean extends Feature {
 			return false;
 		}
 
+		// No need to suggest alternatives if the current query already returned results.
+		if ( $query->found_posts > 0 ) {
+			return false;
+		}
+
 		$term = $this->get_suggested_term( $query );
 		if ( empty( $term ) ) {
+			return false;
+		}
+
+		// No need to suggest the same term the user already typed.
+		$search_term = $query->query_vars['s'] ?? '';
+		if ( strtolower( $term ) === strtolower( $search_term ) ) {
 			return false;
 		}
 
@@ -293,6 +304,12 @@ class DidYouMean extends Feature {
 
 		$term = $this->get_suggested_term( $wp_query );
 		if ( empty( $term ) ) {
+			return;
+		}
+
+		// Do not redirect to the same term the user already searched for.
+		$search_term = $wp_query->query_vars['s'] ?? '';
+		if ( strtolower( $term ) === strtolower( $search_term ) ) {
 			return;
 		}
 
--- a/tests/php/features/TestDidYouMean.php
+++ b/tests/php/features/TestDidYouMean.php
@@ -495,6 +495,35 @@ class TestDidYouMean extends BaseTestCase {
 	}
 
 	/**
+	 * Tests that `ep_suggestion_suppress_when_results_exist` filter can restore the suggestion when results exist.
+	 */
+	public function testGetSuggestionSuppressWhenResultsExistFilter() {
+		$query                        = new \WP_Query(
+			[
+				's' => 'shirt',
+			]
+		);
+		$query->found_posts           = 1;
+		$query->suggested_terms       = [
+			'options' => [
+				[
+					'text' => 'shift',
+				],
+			],
+		];
+		$query->query_vars['s']       = 'shirt';
+		$query->elasticsearch_success = true;
+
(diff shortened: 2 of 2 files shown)


includes/classes/Feature/DidYouMean/DidYouMean.php
```diff
@@ -127,11 +127,22 @@ public function get_suggestion( $query = null ) {
 			return false;
 		}
 
+		// No need to suggest alternatives if the current query already returned results.
+		if ( $query->found_posts > 0 ) {
```
burhandodhy: One concern I have is that if a user searches for a term with a typo and Elasticsearch returns a post, this change will no longer display the suggested search term.

Do you think that's the right behavior, or would it be better for the plugin to continue showing the suggested search term in that case?

burhandodhy (changes requested):

includes/classes/Feature/DidYouMean/DidYouMean.php
faisalahammad: Good point. Added a new `ep_suggestion_suppress_when_results_exist` filter (default `true`). Default behavior stays as fixed here (fixes the #3602 case), but sites that want the suggestion to keep showing next to results can restore it with `add_filter( 'ep_suggestion_suppress_when_results_exist', '__return_false' );`.

Added a test for it (`testGetSuggestionSuppressWhenResultsExistFilter`), updated PR description.

faisalahammad (commented):
Updated per feedback. Added `ep_suggestion_suppress_when_results_exist` filter (default true) so sites can restore the old behavior with one line. Test added: `testGetSuggestionSuppressWhenResultsExistFilter`.

Tests: 21/21 `TestDidYouMean` pass. Full suite 944 tests, no regressions.
Lint: clean.
CodeRabbit: 0 findings.

faisalahammad:
@burhandodhy addressed your feedback, ready for another look.

PR #4338, opened 2026-07-15
fix(woocommerce): subtract tax from price filter bounds

### LINKED ISSUE #4332: BUG: Woo showing wrong results when filtering by price and tax is enabled
### Describe the bug

While using ElasticPress, the min and max price filter doesn't work with tax properly.

Example: the product price set for the product in the admin section is $100 WITHOUT tax. On the website, prices are shown WITH tax (20%) and the price the customer sees is $120. When I set the filter to only show products that cost $120 (min_price=120&max_price=120), this product is NOT shown. 
The reason: the price in the back-end is $100, so it does not fall into this criterion.
Regardless of the fact that the prices on the site are shown WITH tax, ElasticPress takes into account the prices set in the admin panel (back-end).

Maybe the plugin sends prices excluding tax in Elasticsearch when indexing?

In my opinion, the plugin should takes in account tax for each product when indexing OR when user is filtering later on.

### Steps to Reproduce

1. Set tax in Woo
2. Inside /wp-admin/admin.php?page=wc-settings&tab=tax set options like this: 
2.1 Prices entered with tax: NO
2.2 Display prices in the shop: Including tax
3. Set price for a product, for example 100$ (that is excluding tax as we set in step 2.1)
4. Visit shop page. This product will show price 120$ (that is including tax as we set in step 2.2)
5. Filter products (any archive page, not just search) like so: min_price=120&max_price=120
6. The product will NOT show
7. Now, set filter to 100$ like so: min_price=100&max_price=100
8. The product will now be listed

So, users will see one price (120$ in this case) but when they want to use price filter they must set prices without tax - which is confusing.

If I disable ElasticPress, filtering works normal again.

_No response_

### Environment information

_No response_

`## Failed Queries ##

## WordPress ##

<details>

### WordPress Environment ###
wp_version: 7.0
home_url: http://test.local
site_url: http://test.local
is_multisite: false
theme: Blocksy (2.1.47)
plugins: ElasticPress (5.3.3), WooCommerce (10.8.1)
revisions: all

### Server Environment ###
php_version: 8.4.16
memory_limit: 40M
timeout: 180

## Indexable Content ##

Please generate a full report to see the content of this group.

### Settings ###
host: http://REDACTED
index_prefix: 
language: site-default
per_page: 350
network_active: false

### Timeouts ###
index_document_timeout: 15
bulk_request_timeout: 30
request_timeout: 60

### post-1 ###
health: green
status: open
index: post-1
uuid: REDACTED
pri: 5
rep: 0
docs.count: 262
docs.deleted: 24
store.size: 6.6mb
pri.store.size: 6.6mb
total_fields_limit: 5000
analyzer_language: english
stop_language: array (
)
snowball_language: English

### 2026/06/15 12:58:50 am ###
method: WP Dashboard
is_full_sync: Yes
end_date_time: 2026/06/15 12:58:56 am
total_time: 0 hours, 0 minutes, 6 seconds
total: 255
synced: 255
skipped: 0
failed: 0
errors: array (
)
trigger: manual
final_status: success

### Custom Search Results ###
active: true
force_inactive: false

### Post Search ###
active: true
decaying_enabled: 0
force_inactive: false
highlight_enabled: 0
highlight_excerpt: 0
highlight_tag: mark
synonyms_editor_mode: simple

weighting: array (
  'post' => 
  array (
    'post_title' => 
    array (
      'weight' => 1,
      'enabled' => false,
    ),
    'post_content' => 
    array (
      'weight' => 1,
      'enabled' => false,
    ),
    'post_excerpt' => 
    array (
      'weight' => 1,
      'enabled' => false,
    ),
    'author_name' => 
    array (
      'weight' => 1,
      'enabled' => false,
    ),
    'terms.post_tag.name' => 
    array (
      'weight' => 1,
      'enabled' => false,
    ),
    'terms.category.name' => 
    array (
      'weight' => 1,
      'enabled' => false,
    ),
    'terms.ep_custom_result.name' => 
    array (
      'enabled' => true,
      'weight' => 9999,
      'fuzziness' => false,
    ),
  ),
  'page' => 
  array (
    'post_title' => 
    array (
      'weight' => 1,
      'enabled' => false,
    ),
    'post_content' => 
    array (
      'weight' => 1,
      'enabled' => false,
    ),
    'post_excerpt' => 
    array (
      'weight' => 1,
      'enabled' => false,
    ),
    'author_name' => 
    array (
      'weight' => 1,
      'enabled' => false,
    ),
    'terms.ep_custom_result.name' => 
    array (
      'enabled' => true,
      'weight' => 9999,
      'fuzziness' => false,
    ),
  ),
  'product' => 
  array (
    'post_title' => 
    array (
      'enabled' => true,
      'weight' => 100,
    ),
    'post_content' => 
    array (
      'weight' => 1,
      'enabled' => false,
    ),
    'post_excerpt' => 
    array (
      'weight' => 1,
      'enabled' => false,
    ),
    'meta._sku.value' => 
    array (
      'weight' => 1,
      'enabled' => false,
    ),
    'meta._variations_skus.value' => 
    array (
      'weight' => 1,
      'enabled' => false,
    ),
    'terms.ep_custom_result.name' => 
    array (
      'enabled' => true,
      'weight' => 9999,
      'fuzziness' => false,
    ),
  ),
)

### WooCommerce ###
active: true
force_inactive: false
orders: 0
</details>

### Code of Conduct

faisalahammad:
## Summary

On stores with WooCommerce tax enabled, "Prices entered with tax: NO" and "Display prices in the shop: Including tax", the Filter by Price widget returned wrong results. ElasticPress indexed the excluding-tax `_price` but compared the user's including-tax `min_price`/`max_price` bounds directly against it, with no tax conversion on either side. Disabling ElasticPress fixed it because WooCommerce core's native filter converts the bounds.

This fix mirrors WooCommerce core (`WC_Query::price_filter_post_clauses`): when the shop displays including-tax prices but prices are entered excluding tax, the inclusive tax is subtracted from the bounds before the Elasticsearch range query so they line up with the excluding-tax indexed price.

Fixes #4332

### `includes/classes/Feature/WooCommerce/Products.php`

**Before:**
```php
$min_price = ! empty( $_GET['min_price'] ) ? sanitize_text_field( wp_unslash( $_GET['min_price'] ) ) : null;
$max_price = ! empty( $_GET['max_price'] ) ? sanitize_text_field( wp_unslash( $_GET['max_price'] ) ) : null;
// phpcs:enable WordPress.Security.NonceVerification

if ( $query->is_search() ) {
```

**After:**
```php
$min_price = ! empty( $_GET['min_price'] ) ? sanitize_text_field( wp_unslash( $_GET['min_price'] ) ) : null;
$max_price = ! empty( $_GET['max_price'] ) ? sanitize_text_field( wp_unslash( $_GET['max_price'] ) ) : null;
// phpcs:enable WordPress.Security.NonceVerification

// Align bounds with the excluding-tax price Elasticsearch indexes when the
// shop shows including-tax prices, matching WooCommerce core.
if ( null !== $min_price ) {
    $min_price = $this->get_price_filter_tax_adjustment( (float) $min_price );
}
if ( null !== $max_price ) {
    $max_price = $this->get_price_filter_tax_adjustment( (float) $max_price );
}

if ( $query->is_search() ) {
```

**Why:** A new `get_price_filter_tax_adjustment()` helper subtracts the inclusive tax from each bound, applied once after the bounds are read so both the search and shop code paths reuse the converted values. No-op whenever tax is off, no tax rates are configured, prices are entered including tax, or the shop displays excluding tax, so unaffected stores behave exactly as before. No mapping or index change, so no reindex is needed.

## Testing

**Test 1: Reproduce the bug fix (unit)**
1. `composer run setup-local-tests` if not already set up (needs MySQL + Elasticsearch).
2. `vendor/bin/phpunit --filter testPriceFilterWithTax tests/php/features/WooCommerce/TestWooCommerceProduct.php`
3. The test seeds a 20% tax rate, sends `min_price=120&max_price=120`, and asserts the Elasticsearch range bound is reduced to `100.0` (the excluding-tax price).
Result: passes.

**Test 2: Existing price filters still green**
1. `vendor/bin/phpunit --filter testPriceFilter tests/php/features/WooCommerce/TestWooCommerceProduct.php`
2. `vendor/bin/phpunit --filter testPriceFilterWithSearchQuery tests/php/features/WooCommerce/TestWooCommerceProduct.php`
Result: both pass (tax adjustment is a no-op when tax is off).

**Test 3: Manual repro (issue #4332)**
1. WooCommerce > Settings > Tax: enable tax, prices entered excluding tax, display including tax.
2. Add a 20% standard rate for the base location.
3. Create a product priced 100 (excl tax); it shows 120 (incl tax).
4. Filter `?min_price=120&max_price=120`.
Result: product appears (before fix it did not).

### DIFF
--- a/includes/classes/Feature/WooCommerce/Products.php
+++ b/includes/classes/Feature/WooCommerce/Products.php
@@ -171,6 +171,7 @@ class Products {
 	 * in when prices are entered without tax but the shop shows including-tax
 	 * prices, the case where the Filter by Price widget sends including-tax bounds.
 	 *
+	 * @since 5.3.4
 	 * @param float $price Raw bound from min_price or max_price.
 	 * @return float
 	 */
--- a/includes/classes/Feature/WooCommerce/Products.php
+++ b/includes/classes/Feature/WooCommerce/Products.php
@@ -135,29 +135,29 @@ class Products {
 			unset( $args['query']['bool']['should'] );
 
 			if ( ! empty( $min_price ) ) {
-				$args['query']['bool']['must'][0]['range']['meta._price.long']['gte'] = $min_price;
+				$args['query']['bool']['must'][0]['range']['meta._price.double']['gte'] = $min_price;
 			}
 
 			if ( ! empty( $max_price ) ) {
-				$args['query']['bool']['must'][0]['range']['meta._price.long']['lte'] = $max_price;
+				$args['query']['bool']['must'][0]['range']['meta._price.double']['lte'] = $max_price;
 			}
 
-			$args['query']['bool']['must'][0]['range']['meta._price.long']['boost'] = 2.0;
-			$args['query']['bool']['must'][1]['bool']                               = $old_query;
+			$args['query']['bool']['must'][0]['range']['meta._price.double']['boost'] = 2.0;
+			$args['query']['bool']['must'][1]['bool']                                 = $old_query;
 		} else {
 			unset( $args['query']['match_all'] );
 
-			$args['query']['range']['meta._price.long']['gte'] = ! empty( $min_price ) ? $min_price : 0;
+			$args['query']['range']['meta._price.double']['gte'] = ! empty( $min_price ) ? $min_price : 0;
 
 			if ( ! empty( $min_price ) ) {
-				$args['query']['range']['meta._price.long']['gte'] = $min_price;
+				$args['query']['range']['meta._price.double']['gte'] = $min_price;
 			}
 
 			if ( ! empty( $max_price ) ) {
-				$args['query']['range']['meta._price.long']['lte'] = $max_price;
+				$args['query']['range']['meta._price.double']['lte'] = $max_price;
 			}
 
-			$args['query']['range']['meta._price.long']['boost'] = 2.0;
+			$args['query']['range']['meta._price.double']['boost'] = 2.0;
 		}
 
 		return $args;
--- a/includes/classes/Feature/WooCommerce/Products.php
+++ b/includes/classes/Feature/WooCommerce/Products.php
@@ -118,6 +118,15 @@ class Products {
 		$max_price = ! empty( $_GET['max_price'] ) ? sanitize_text_field( wp_unslash( $_GET['max_price'] ) ) : null;
 		// phpcs:enable WordPress.Security.NonceVerification
 
+		// Align bounds with the excluding-tax price Elasticsearch indexes when the
+		// shop shows including-tax prices, matching WooCommerce core.
+		if ( null !== $min_price ) {
+			$min_price = $this->get_price_filter_tax_adjustment( (float) $min_price );
+		}
+		if ( null !== $max_price ) {
+			$max_price = $this->get_price_filter_tax_adjustment( (float) $max_price );
+		}
+
 		if ( $query->is_search() ) {
 			/**
 			 * This logic is iffy but the WC price filter widget is not intended for use with search anyway
@@ -152,6 +161,44 @@ class Products {
 		}
 
 		return $args;
+	}
+
+	/**
+	 * Subtract inclusive tax from a price filter value so it matches the
+	 * excluding-tax price Elasticsearch stores.
+	 *
+	 * Mirrors WooCommerce core (WC_Query::price_filter_post_clauses). Only kicks
+	 * in when prices are entered without tax but the shop shows including-tax
+	 * prices, the case where the Filter by Price widget sends including-tax bounds.
+	 *
+	 * @param float $price Raw bound from min_price or max_price.
+	 * @return float
+	 */
+	protected function get_price_filter_tax_adjustment( $price ) {
+		if ( ! function_exists( 'wc_tax_enabled' ) || ! wc_tax_enabled() ) {
+			return $price;
+		}
+
+		if ( 'incl' !== get_option( 'woocommerce_tax_display_shop' ) ) {
+			return $price;
+		}
+
+		if ( function_exists( 'wc_prices_include_tax' ) && wc_prices_include_tax() ) {
+			return $price;
+		}
+
+		if ( ! method_exists( 'WC_Tax', 'get_rates' ) ) {
+			return $price;
+		}
+
+		$tax_class = apply_filters( 'woocommerce_price_filter_widget_tax_class', '' ); // Standard tax class.
+		$tax_rates = \WC_Tax::get_rates( $tax_class );
+
+		if ( empty( $tax_rates ) ) {
+			return $price;
+		}
+
+		return $price - \WC_Tax::get_tax_total( \WC_Tax::calc_inclusive_tax( $price, $tax_rates ) );
 	}
 
 	/**
--- a/tests/php/features/WooCommerce/TestWooCommerceProduct.php
+++ b/tests/php/features/WooCommerce/TestWooCommerceProduct.php
@@ -703,6 +703,7 @@ class TestWooCommerceProduct extends WooCommerceBaseTestCase {
 	 * tax. The Filter by Price widget sends including-tax bounds, which must be
 	 * reduced to the excluding-tax value before the Elasticsearch range query.
 	 *
+	 * @since 5.3.4
 	 * @group woocommerce
 	 * @group woocommerce-products
 	 */
--- a/tests/php/features/WooCommerce/TestWooCommerceProduct.php
+++ b/tests/php/features/WooCommerce/TestWooCommerceProduct.php
@@ -718,6 +718,7 @@ class TestWooCommerceProduct extends WooCommerceBaseTestCase {
 			'woocommerce_tax_display_shop',
 			'woocommerce_prices_include_tax',
 			'woocommerce_default_country',
+			'woocommerce_tax_based_on',
 		);
 		$old_options = array();
 		foreach ( $option_keys as $option_key ) {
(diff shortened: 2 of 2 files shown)


Sidsector9:
Thanks for the PR @faisalahammad and for the clear description.

The approach is correct and it fixes the case in the issue.

There is one problem to solve before we can merge this. The price filter compares prices using `meta._price.long`. This field stores the price as a whole number, so a product priced `100.99` is stored as `100`. When the tax is subtracted, the filter value is usually not a whole number. Because of this, a product with a price that is not a whole number may not appear in the results.

You can see this yourself. Set a product price to `100.99`. The shop will show `121.19`. Then open `/shop?min_price=121&max_price=200`. The product should appear, but it does not. Could you please look into this?

It would be great if you could add a test that uses a price with decimal numbers. The current test uses `100` and `120`. Both are whole numbers, so the test passes even when this rounding problem exists.

Let us know if you have any questions.

faisalahammad:
Thanks for flagging the rounding issue. Fixed in 44f4ae5.

The range query now targets meta._price.double (already indexed alongside long). meta._price.long truncates via intval() — 100.99 stored as 100, so tax-adjusted fractional bounds like 100.83 missed it. meta._price.double keeps the full decimal.

Test updated: seeds regular_price 100.99 and min_price=121.188 (the exact incl-tax WC computes for that price), so the test now reproduces the rounding bug directly. Also updated testPriceFilterWithoutTax's stale meta._price.long assertion.

No reindex needed.

faisalahammad:
## CI Fix Summary — 1 PHPUnit failure resolved

| # | File | Error | Fix |
|---|------|-------|-----|
| 1 | `tests/php/features/WooCommerce/TestWooCommerceProduct.php:739` | `ElasticPressTest\TestWooCommerceProduct::testPriceFilterWithTax` — expected gte/lte = 100.99, got 121.188 | Pin `woocommerce_tax_based_on = "base"` in test setup so `WC_Tax::get_rates("")` resolves to the shop base country (GB) where the seeded 20% rate is seeded. Without it, `WC_Customer::get_taxable_address()` routes to a non-GB tuple, `get_rates` returns empty, and `Products::get_price_filter_tax_adjustment()` skips the incl→excl subtraction. |

### Why the helper itself is unchanged
`get_price_filter_tax_adjustment()` is correct in production — real HTTP requests populate the customer country via session or geo-IP. The test was just not replicating that state.

### Cleanup
The test option-snapshot loop already restores all captured options in `finally`. Adding `woocommerce_tax_based_on` to the captured-keys array covers rollback automatically.

### Verification
- Resolves 6/6 failing PHPUnit matrix jobs: Single Site + Multisite × ES 7.10.1 / 8.12.2 / 9.1.5.
- No regression in `testPriceFilter` or `testPriceFilterWithSearchQuery` (both have taxes disabled, never reach `WC_Tax::get_rates`).
- PHPCS clean on changed file.
- CodeRabbit: no findings.
- PHP 7.4+ compatible (test definitions match CI matrix PHP 8.2).

### Out of scope
`@paidPlugins` E2E jobs already failed on `develop` for unrelated reasons — fork PRs do not have access to paid-plugin secrets. Not addressed here.

includes/classes/Feature/WooCommerce/Products.php
```diff
@@ -126,34 +135,72 @@ public function price_filter( $args, $query_args, $query ) {
...
+	 * Subtract inclusive tax from a price filter value so it matches the
+	 * excluding-tax price Elasticsearch stores.
+	 *
+	 * Mirrors WooCommerce core (WC_Query::price_filter_post_clauses). Only kicks
+	 * in when prices are entered without tax but the shop shows including-tax
+	 * prices, the case where the Filter by Price widget sends including-tax bounds.
+	 *
+	 * @param float $price Raw bound from min_price or max_price.
+	 * @return float
+	 */
```
felipeelia, line 177:
```suggestion
	/**
	 * Subtract inclusive tax from a price filter value so it matches the
	 * excluding-tax price Elasticsearch stores.
	 *
	 * Mirrors WooCommerce core (WC_Query::price_filter_post_clauses). Only kicks
	 * in when prices are entered without tax but the shop shows including-tax
	 * prices, the case where the Filter by Price widget sends including-tax bounds.
	 *
	 * @since 5.3.4
	 * @param float $price Raw bound from min_price or max_price.
	 * @return float
	 */
```

tests/php/features/WooCommerce/TestWooCommerceProduct.php
```diff
@@ -695,6 +695,134 @@ public function testPriceFilterWithSearchQuery() {
...
+	 * Test the price filter subtracts inclusive tax from bounds so they match
+	 * the excluding-tax price indexed by Elasticsearch.
+	 *
+	 * Reproduces issue #4332: prices entered without tax, shop displays including
+	 * tax. The Filter by Price widget sends including-tax bounds, which must be
+	 * reduced to the excluding-tax value before the Elasticsearch range query.
+	 *
+	 * @group woocommerce
+	 * @group woocommerce-products
+	 */
```
felipeelia, line 709:
```suggestion
	/**
	 * Test the price filter subtracts inclusive tax from bounds so they match
	 * the excluding-tax price indexed by Elasticsearch.
	 *
	 * Reproduces issue #4332: prices entered without tax, shop displays including
	 * tax. The Filter by Price widget sends including-tax bounds, which must be
	 * reduced to the excluding-tax value before the Elasticsearch range query.
	 *
	 * @since 5.3.4
	 * @group woocommerce
	 * @group woocommerce-products
	 */
```

felipeelia (changes requested):
Hey @faisalahammad! Any chance you can include these `@since` tags so we can finally merge this PR? Thanks!

includes/classes/Feature/WooCommerce/Products.php
faisalahammad: Done. Added the @since 5.3.4 tag.

tests/php/features/WooCommerce/TestWooCommerceProduct.php
faisalahammad: Done. Added the @since 5.3.4 tag here too.

faisalahammad:
@felipeelia Done — added the `@since 5.3.4` tags in both suggested spots. Both threads resolved. Ready for re-review when you have a moment.

PR #4356, opened 2026-08-18
10up-toolkit update

felipeelia:
### Description of the Change

Closes #

### Changelog Entry

> Added - New feature
> Removed - Feature
> Fixed - Bug fix
> Security - Vulnerability
> Developer - Non-functional update

### Credits

Props @username, ...

### Checklist:


felipeelia:
## Approach to the CodeQL privileged-workflow alerts

The remaining [checkout alert](https://github.com/10up/ElasticPress/pull/4356#discussion_r3813576583) is expected. This job has to check out PR code and run it with repository secrets (EP.io + ACF Pro) after a maintainer applies `safe-to-test` or dispatches the workflow. CodeQL cannot see that human gate, and a `# codeql[...]` suppression on the checkout step did not clear it.

What we changed in 3074d1c:

1. **Restore-only caches** in `playwright-privileged.yml`. `workflow_dispatch` still has read-write access to the default-branch cache, so this job no longer saves npm, composer, or `bin/.cache` after an untrusted checkout. That should close the cache-poisoning finding.
2. **Trusted writers stay in `playwright.yml`** (`push` / same-repo `pull_request`), which is what the privileged job restores from.
3. **`actions/checkout@v7`** with `allow-unsafe-pr-checkout: true` and `persist-credentials: false`. v7 blocks fork PR heads in `pull_request_target` unless that opt-in is set; we still only run after `safe-to-test` / dispatch, and the label is stripped on new pushes.

The checkout finding can be dismissed as accepted risk if CodeQL keeps reporting it. Please do not drop the `safe-to-test` / dispatch gate or add unattended `pull_request_target` triggers.

burhandodhy (approved):

